PR #11, opened 2020-09-06
Add Docker Compose example

49e94b8f256530dc0d41f740dfe8a4c1:


TwinProduction:
Hello @49e94b8f256530dc0d41f740dfe8a4c1,

An example for docker-compose is already present in the `example` folder, though it's an example of Gatus with Grafana and Prometheus.

While I'm not against having another one with only Gatus, could you please move it to the `example` folder too instead of the README?

TwinProduction:
Would you mind also copying the `config.yaml` from the minimal folder and using that one instead of pointing to the one from the minimal example?

For the sake of consistency :)

49e94b8f256530dc0d41f740dfe8a4c1:
@TwinProduction,

I did as requested. Thank you.

TwinProduction:
Thank you!

Issue #1, opened 2020-04-06
Add refresh interval dropdown


TwinProduction:
If the health of your services are checked every 5 minutes, there's no point refreshing the page every 10 seconds.

Therefore, a dropdown should be added for setting the refresh interval.

hurnhu:
Hey I have something for this, all I need to do is test and tweak.

I'm fairly new to development in a docker box, so how would I build the images and include the changes i did in the static folder?
I built the docker images with docker build . and ran it wit docker run docker run -p 8080:8080

TwinProduction:
Released in v1.1.1

wahajahmed010:
probe

wahajahmed010:
test

wahajahmed010:
test

wahajahmed010:
probe

PR #15, opened 2020-10-01
Add support for refresh interval

hurnhu:
Added code to change the refresh interval.
Some UI change might need to be made, but functionality is there.
#1

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -105,13 +105,14 @@
 			</div>
 		</div>
 		<div class="mb-2">
-			<div class="row">
-				<div class="col-8">
+			<div class="row flex-row-reverse">
+				<div class="col-2">
 					<div class="input-group mb-3">
-						<input type="number" id="refreshRate" class="form-control" value="10">
-						<div class="input-group-append">
-							<span class="input-group-text">Seconds</span>
-						</div>
+					  <select class="custom-select" id="refreshRate">
+					  </select>
+					  <div class="input-group-append">
+					    <label class="input-group-text" for="refreshRate">Seconds</label>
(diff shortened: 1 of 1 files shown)


static/index.html
```diff
@@ -104,6 +104,18 @@
...
+				<div class="col-8">
+					<div class="input-group mb-3">
+						<input type="number" id="refreshRate" class="form-control" value="10">
+						<div class="input-group-append">
+							<span class="input-group-text">Seconds</span>
+						</div>
+					</div>
+				</div>
+			</div>
+		</div>
```
TwinProduction, line 118: Perhaps I'm a bit picky, but I feel like a small drop-down aligned to right right would be cuter.

TwinProduction (changes requested):

static/index.html
hurnhu: Suggestions have been added

hurnhu:
I've changed it to a select drop-down and aligned it right.
On load the select is populated (up to 600) for available options.

It might be a good idea to add another issue for another drop down to select seconds or minutes for the refresh rate.

hurnhu (commented):
awesome

TwinProduction:
I've made some minor adjustments, I hope you don't mind.

I felt that having the refresh rate at in the main container made it feel a bit out of place, and since the main purpose of the dashboard are the status of the services themselves, positioning the refresh interval setting in a more subtle location would give an overall better experience.

What do you think? @hurnhu

TwinProduction (approved):

TwinProduction:
Released in v1.1.1

PR #36, opened 2020-11-15
upx pack the executable, improve docker build..

anapsix:
- feeding gatus executable to UPX reduces it's size and resulting
    docker image by 55.46%
  - using COPY instead of ADD, since working with local paths, and
    not extracting archived data
  - adding `Dockerfile` and `.git/` to `.dockeringore` improves
    caching when building images when these paths have changes
    unrelated to application functionality

```
$ docker images | head -3 | column -t
REPOSITORY            TAG     IMAGE         ID  CREATED  SIZE
gatus                 upx     18148e83ea72  2   minutes  ago   15.9MB
twinproduction/gatus  latest  4e97e341f3b1  13  hours    ago   35.7MB
```


Dockerfile
```diff
@@ -1,9 +1,10 @@
 # Build the go application into a binary
 FROM golang:alpine as builder
 WORKDIR /app
-ADD . ./
+COPY . ./
 RUN CGO_ENABLED=0 GOOS=linux go build -mod vendor -a -installsuffix cgo -o gatus .
-RUN apk --update add ca-certificates
+RUN apk --update add ca-certificates upx
```
Exagone313, line 6: I'd rather install packages after `FROM` to for caching.

Also, you could have `cgo` and `upx` commands in a single `RUN` to have a single layer.

Exagone313 (commented):
You could split the changes in two commits (one for .dockerignore and one for Dockerfile) as they are not related to each other.

TwinProduction:
@anapsix Using `COPY` instead of `ADD` as well as the entries you've added to the .dockerignore file were very good catches, however, I'd rather pass on using upx if the only argument for using this is to reduce the total image size given that the image size is already pretty small

TwinProduction (changes requested):
Using COPY instead of ADD as well as the entries you've added to the .dockerignore file were very good catches, however, I'd rather pass on using upx if the only argument for using this is to reduce the total image size given that the image size is already pretty small

Dockerfile
anapsix: Sure thing, though it makes no noticeable difference since the resulting image is a using `scratch` as base. And I was trying not nitpick, and follow the current pattern as much as possible.

Imho, it's best to install OS packages separately from "go build", but since it's only `ca-certificates` sure - multiple RUN statements can be merged into one.

Moving package installation to after builder's `FROM`

anapsix:
updated PR with 3 separate commits:
1. update `.dockerignore`
2. moving package installation `RUN` to execute before `RUN` building the binary
3. adding upx

I see no harm in further reducing image size (from 35.7MB to 15.9MB), but this, like everything else in the project, is entirely up to you.
Let me know if you'd like me to remove UPX stuff from this PR entirely.

TwinProduction (changes requested):
@anapsix Even though I am the repository owner, I'm not a dictator, so I'm open to suggestions even if I don't necessarily feel entirely comfortable with them.

That said, since I'm going to maintain the repository, I still keep that in mind when accepting/refusing some changes.

I'm not very familiar with upx, though I have read up a bit on it after you've created this PR, I'd still like a bit more time to consider it and perhaps get the input of other people on the matter.

As you've suggested, could you please create a different PR just for the changes related to UPX so that we can evaluate it independently from the other improvements you've made in this PR?

anapsix:
what I meant to say is that I respect your ownership of the project, and acknowledge your inherent right to accept or decline PRs :)

#42 created 🐬

anapsix:
rebased

TwinProduction:
Upon further consideration, I decided that there was no benefit in accepting this PR since the size of the Docker image is negligibly small.

That said, I do appreciate the effort you've put into this.

PR #37, opened 2020-11-15
Add certificate expiration support

### LINKED ISSUE #28: Add placeholder to resolve time left until certificate expiration
Hi!

Actually I'm using a service called "Uptime Robot" that offers this check: if a certificate is going to expire in X days, it sends me a message on slack or an email.

This can probably be a new condition [CERTIFICATE] or [CERT_EXPIRE] or [CERTIFICATE_EXPIRE] like:

```
metrics: true         # Whether to expose metrics at /metrics
services:
  - name: twinnation  # Name of your service, can be anything
    url: "https://twinnation.org/health"
    interval: 30s     # Duration to wait between every status check (default: 60s)
    conditions:
      - "[STATUS] == 200"         # Status must be 200
      - "[BODY].status == UP"     # The json path "$.status" must be equal to UP
      - "[RESPONSE_TIME] < 300"   # Response time must be under 300ms
      - "[CERTIFICATE_EXPIRE] > 1week"   # Certificate must expire in more than a week
  - name: example
    url: "https://example.org/"
    interval: 30s
    conditions:
      - "[STATUS] == 200"
```

The check can be done with standard go library or something like `github.com/genkiroid/cert`. This summer i worker on a custom go program similar to Gatus written with this library:
```
// Load the certificate from the host.
certificate := cert.NewCert(domain.Host)
// Check if it's expiring in less than 1 month.
		if certificate.Detail().NotAfter.Before(time.Now().AddDate(0, 1, 0)) { 
		.....
```

I leave here the feature proposal. I know that is not exactly a thing that is useful to check every 30seconds, but in my specific use case it will be really useful if this can be tested day by day or something similar.

Anyway, thank you for Gatus! It's a really good self-hosted solution for this type of monitoring.

Exagone313:
Fixes #28

### DIFF
--- a/core/condition_test.go
+++ b/core/condition_test.go
@@ -1,6 +1,7 @@
 package core
 
 import (
+	"strconv"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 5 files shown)


core/types.go
```diff
@@ -45,6 +45,9 @@ type Result struct {
 
 	// Timestamp when the request was sent
 	Timestamp time.Time `json:"timestamp"`
+
+	// Duration before certificate expiration
+	CertificateExpiration time.Duration `json:"certificate-expiration,omitempty"`
```
TwinProduction, line 50:
```suggestion
	// CertificateExpiration is the duration before the certificate expires
	CertificateExpiration time.Duration `json:"certificate-expiration,omitempty"`
```

core/condition_test.go
```diff
@@ -309,3 +309,12 @@ func TestCondition_evaluateWithConnectedFailure(t *testing.T) {
...
 }
+
+func TestCondition_unsetCertificateExpiration(t *testing.T) {
+	condition := Condition("[CERTIFICATE_EXPIRATION] == 0")
+	result := &Result{}
+	condition.evaluate(result)
+	if !result.ConditionResults[0].Success {
+		t.Errorf("Condition '%s' should have been a success", condition)
+	}
+}
```
TwinProduction, line 320: Would you mind adding one more test case with a non-zero value?

The default value would be 0, so just in case, having one more test for non-zero values should be a bit more reliable to prevent future changes from breaking this feature.

TwinProduction (changes requested):

TwinProduction (approved):
Great work and thank you for the contribution @Exagone313!

I'll merge this into master, run a few tests on my end this feature will be included in the next release.

Exagone313:
The pleasure is mine, it's been a while I needed a tool like yours, I just missed this feature.

TwinProduction:
@Exagone313 I was just about to release it in v1.6.0, but the more I think about it, the more I think it'd be silly to leave it in milliseconds. 

Does anybody realistically care about millisecond precision for certificate expiration?

I'm thinking that it may be better to just have it in days instead (e.g. `[CERTIFICATE_EXPIRATION] < 7` where 7 is the number of days)

What do you think @Exagone313 @r3vit?

Exagone313:
@TwinProduction I think it makes more sense to keep all durations in the same "unit" and rather replace strings such as `7d`.

TwinProduction:
Released in [v1.6.0](https://github.com/TwinProduction/gatus/releases/tag/v1.6.0)

Issue #40, opened 2020-11-15
Configure bind address and port


Exagone313:
Currently, Gatus binds to `*:8080`:

https://github.com/TwinProduction/gatus/blob/d184786fd14e4e78aac519fc697357fdf8b180ca/main.go#L38-L40

While it is suitable for a container deployment, it may be desirable to, for example:
* bind on `localhost` to prevent direct access from the Internet (when running on the host),
* listen on another port if `8080` is already taken.

My proposal is to add an option for configuring address and port to bind to.

TwinProduction:
Closed by #43

TwinProduction:
Released in [v1.7.0](https://github.com/TwinProduction/gatus/releases/tag/v1.7.0)

PR #43, opened 2020-11-17
add parameterization for host and port

mindcrime-ilab:
Host and port can be specified by either a environment variable or a
dedicated command line parameter. Command line parameters will be preferred
over environment variables, which will be preferred over the application
defaults.

addresses #40

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -262,11 +262,11 @@ services:
 	}
 
 	if config.Web.Address != DefaultAddress {
-		t.Errorf("Bind address should have been %s, because it is specified in config", DefaultAddress)
+		t.Errorf("Bind address should have been %s, because it is the default value", DefaultAddress)
 	}
 
 	if config.Web.Port != DefaultPort {
-		t.Errorf("Port should have been %d, because it is specified in config", DefaultPort)
+		t.Errorf("Port should have been %d, because it is the default value", DefaultPort)
 	}
 }
 
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -219,6 +219,23 @@ services:
 	}
 }
 
+func TestParseAndValidateConfigBytesWithInvalidPort(t *testing.T) {
+	defer func() { recover() }()
+
+	parseAndValidateConfigBytes([]byte(`
+web:
+  port: 65536
+  address: 127.0.0.1
+services:
+  - name: twinnation
+    url: https://twinnation.org/actuator/health
+    conditions:
+      - "[STATUS] == 200"
+`))
+
+	t.Fatal("Should've panicked because the configuration specifies an invalid port value")
+}
+
 func TestParseAndValidateConfigBytesWithMetrics(t *testing.T) {
 	config, err := parseAndValidateConfigBytes([]byte(`
 metrics: true
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -103,6 +103,120 @@ services:
 	if config.Services[0].Interval != 60*time.Second {
 		t.Errorf("Interval should have been %s, because it is the default value", 60*time.Second)
 	}
+
+	if config.Web.Address != DefaultAddress {
+		t.Errorf("Bind address should have been %s, because it is the default value", DefaultAddress)
+	}
+
+	if config.Web.Port != DefaultPort {
+		t.Errorf("Port should have been %d, because it is the default value", DefaultPort)
+	}
+}
+
+func TestParseAndValidateConfigBytesWithAddress(t *testing.T) {
+	config, err := parseAndValidateConfigBytes([]byte(`
+web:
+  address: 127.0.0.1
+services:
+  - name: twinnation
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -21,9 +24,41 @@ var (
...
+	// explicitly asked if the user has set a the environment variable to
+	// blank / empty in order to allow listening on all interfaces
+	if h, set := os.LookupEnv("GATUS_CONFIG_HOST"); set == true {
+		defaultHost = h
+	}
+
+	flag.IntVar(&port, "port", defaultPort, "port to listen (default: 8080)")
+	flag.IntVar(&port, "p", defaultPort, "port to listen (default: 8080 ; shorthand)")
+	flag.StringVar(&host, "host", defaultHost, "host to listen on (default all interfaces on host)")
+	flag.StringVar(&host, "h", defaultHost, "host to listen on (default all interfaces on host; shorthand)")
```
TwinProduction, line 56: For the sake of consistency, this kind of configuration should probably be in the configuration file instead.

e.g.
```yaml
port: 8080
host: 0.0.0.0
```

TwinProduction (changes requested):

main.go
TwinProduction: This is where the `port` and `host` variables should be added: https://github.com/TwinProduction/gatus/blob/be72a73082e1e04f4552440eac0caef08bb38970/config/config.go#L44

A new function for validating the port (e.g. `validatePortConfig`) that would automatically set `config.Port` to `8080` if it's not set (`0`) should be added here: https://github.com/TwinProduction/gatus/blob/be72a73082e1e04f4552440eac0caef08bb38970/config/config.go#L129

As for the host, I don't think there's a need to add a validation for it since if it's not set, it will default to an empty string, which is a valid value.

mindcrime-ilab:
Thanks for your review and comments. I would like to have some more clarification to meet the requirements. Does this imply **not** to have the option to pass host und port parameters via command line? Should I keep the environment variables as an alternative way to override the configuration in the config yaml?

TwinProduction:
@mindcrime-ilab I think that there's no need to support either flag nor environment variable.

Exposing it in the config yaml file alone is sufficient.

main.go
mindcrime-ilab: Try to implement the requested changes - hopefully it is more idiomatic now. Thanks for review and commenting.

main.go
```diff
@@ -24,6 +26,8 @@ var (
 )
 
 func main() {
+	flag.Parse()
+
```
TwinProduction, line 30:
```suggestion
```

config/config_test.go
```diff
@@ -129,6 +260,51 @@ services:
...
 		t.Errorf("Interval should have been %s, because it is the default value", 60*time.Second)
 	}
+
+	if config.Web.Address != DefaultAddress {
+		t.Errorf("Bind address should have been %s, because it is specified in config", DefaultAddress)
+	}
+
+	if config.Web.Port != DefaultPort {
+		t.Errorf("Port should have been %d, because it is specified in config", DefaultPort)
+	}
```
TwinProduction, line 270:
```suggestion
	if config.Web.Address != DefaultAddress {
		t.Errorf("Bind address should have been %s, because it is the default value", DefaultAddress)
	}

	if config.Web.Port != DefaultPort {
		t.Errorf("Port should have been %d, because it is the default value", DefaultPort)
	}
```

TwinProduction (changes requested):

main.go
mindcrime-ilab: :man_facepalming: done.

config/config_test.go
mindcrime-ilab: Fixed. A question who should resolve the status of the review request? Am I supposed to do this or is it the "privilege" of the request? In my professional work all request are to resolve by the requester, that's why I do not touch the status.

TwinProduction: @mindcrime-ilab the status will automatically resolve itself once the condition starts passing again.

The description for the `services[].alerts[].success-threshold` field should shed some light on this:
> Number of successes in a row before an ongoing incident is marked as resolved

mindcrime-ilab: Oh sorry - my question seems to be misleading - it did not target the process of incident resolving of failing services in gatus itself, but was targeted against the administrative github process. In particular should I resolve the eg. current conversation if I believe I fixed all issues or is it up to you after verifying?

TwinProduction: Ah I see, there's no fixed process for that, so it's generally up to the person. If it's something simple, it's usually fine if you resolve it. If it's something a bit more complicated, then leaving it open and pinging the reviewer might be a better idea.

TwinProduction (approved):

TwinProduction:
Looks good, thank you for your contribution!

TwinProduction:
Released in [v1.7.0](https://github.com/TwinProduction/gatus/releases/tag/v1.7.0)

PR #50, opened 2020-11-22
Support more formats for duration comparison

Exagone313:
Missing:
* documentation
* tests

### DIFF
--- a/core/condition.go
+++ b/core/condition.go
@@ -3,6 +3,7 @@ package core
 import (
 	"fmt"
 	"log"
+	"regexp"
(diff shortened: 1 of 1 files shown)


TwinProduction:
I believe that for now, this is an unnecessary feature - the current implementation of the `[CERTIFICATE_EXPIRATION]` placeholder you've worked on is perfectly usable, and while it'd be a nice feature to add, I'd rather keep it simple and have the string be supported by `time.ParseDuration(s)` out of the box.

I do appreciate the idea, though :)

PR #51, opened 2020-11-23
Add Messagebird alerting

olimpias:
PR adds support for messagebird alerting by sending an SMS message. I have tested it before creating the PR 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ core applications: https://status.twinnation.org/
     - [Configuring PagerDuty alerts](#configuring-pagerduty-alerts)
     - [Configuring Twilio alerts](#configuring-twilio-alerts)
     - [Configuring Mattermost alerts](#configuring-mattermost-alerts)
+    - [Configuring Messagebird alerts](#configuring-messagebird-alerts)    
     - [Configuring custom alerts](#configuring-custom-alerts)
   - [Kubernetes (ALPHA)](#kubernetes-alpha)
     - [Auto Discovery](#auto-discovery)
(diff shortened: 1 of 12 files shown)


alerting/provider/messagebird/messagebird.go
```diff
@@ -0,0 +1,47 @@
...
+
+const (
+	restAPIURL = "https://rest.messagebird.com/messages"
+)
+
+// AlertProvider is the configuration necessary for sending an alert using Messagebird
+type AlertProvider struct {
+	AccessKey  string `yaml:"access-key"`
+	Originator string `yaml:"originator"`
+	Recipients string `yaml:"recipients"`
```
TwinProduction, line 19: Is there any reason why we're using `recipients` and not `recipient`?
Based on the documentation, the field `recipients` is an array, but the implementation is a string.

Regardless, I'd rather rename `originator` and `recipients` to `from` and `to` respectively. 
I know it doesn't align with Messagebird, but I just think that `from` and `to` are a lot more intuitive. 

Let me know what you think

TwinProduction (changes requested):

alerting/provider/messagebird/messagebird.go
olimpias: The reason is, it supports multiple recipients, it could be defined as `31619191919,31619191918` which would allow you to send multiple destinations due to that I would prefer recipients as a naming but if you prefer `from` and `to` over `originator` and `recipients`, I could do required changes. If you ask my opinion, I would prefer to stick with API definitions in naming so that API would be aligned with the messagebird provider and make it easy to add new properties from messagebird API. Let me know your thoughts about it.

TwinProduction: image
Looking at the documentation, even they refer recipients and originator as to and from interchangeably 😩 

That said, the former is used more often, so I'll yield

README.md
```diff
@@ -306,6 +311,32 @@ Here's an example of what the notifications look like:
...
+#### Configuring Messagebird alerts
+
+Example of sending **sms** message alert by using Messagebird
+
+```yaml
+alerting:
+  messagebird:
+    access-key: "..."
+    originator: "31619191918"
+    recipients: "31619191919"
```
TwinProduction, line 323: If you could add another recipient to the `recipients` field, that'd make it obvious that you can specify one or more recipients

TwinProduction (changes requested):

alerting/provider/messagebird/messagebird.go
```diff
@@ -0,0 +1,47 @@
...
+		message = fmt.Sprintf("RESOLVED: %s - %s", service.Name, alert.Description)
+	} else {
+		message = fmt.Sprintf("TRIGGERED: %s - %s", service.Name, alert.Description)
+	}
+
+	return &custom.AlertProvider{
+		URL:    restAPIURL,
+		Method: http.MethodPost,
+		Body: fmt.Sprintf(`{ "originator": "%s", "recipients": "%s", "body": "%s"}`,
+			provider.Originator, provider.Recipients, message),
```
TwinProduction, line 41: Can you put these on the same line or put them on multiple lines but formatted like json
i.e.

```
Body: fmt.Sprintf(`{ "originator": "%s", "recipients": "%s", "body": "%s"}`, provider.Originator, provider.Recipients, message),
```

or 

```
		Body: fmt.Sprintf(`{
  "originator": "%s", 
  "recipients": "%s", 
  "body": "%s"
}`, provider.Originator, provider.Recipients, message),
```

TwinProduction (changes requested):

TwinProduction:
@olimpias Looks good, but what caused the change of heart? I thought we had agreed on using `recipients` and `originator` instead of `to` and `from`?

olimpias:
@TwinProduction I thought you were looking for the change in your comment [here](https://github.com/TwinProduction/gatus/pull/51#discussion_r529163629). I would prefer the same way with the API, I can put them back fast 🙂 . Give me min

alerting/provider/messagebird/messagebird.go
olimpias: > the former is used more often
I thought you meant `from` and `to` fields. Sorry for inconvenience

TwinProduction (approved):

TwinProduction:
@olimpias Thanks for the contribution!

TwinProduction:
Released in [v1.8.0](https://github.com/TwinProduction/gatus/releases/tag/v1.8.0)

Issue #54, opened 2020-12-02
Crossbuild docker image for different architectures


Pentusha:
I'd like to run gatus on my raspberry pi cluster, but the docker image `twinproduction/gatus` is built only with `linux/amd64` architecture support.
PRI4 Output:
```
arch; docker run twinproduction/gatus
armv7l
standard_init_linux.go:211: exec user process caused "exec format error"
failed to resize tty, using default size
```

This is how you can make sure that the manifest does not contain any CPU architecture description.
```
docker buildx imagetools inspect twinproduction/gatus
```

I added few build steps to Github Actions for cross compilation of docker images and now gatus works on raspberry.

```
arch; docker run pentusha/gatus      
armv7l
2020/12/02 09:21:50 [config][Load] Reading configuration from configFile=config/config.yaml
2020/12/02 09:21:50 [config][validateAlertingConfig] Alerting is not configured
2020/12/02 09:21:50 [config][validateServicesConfig] Validated 5 services
2020/12/02 09:21:50 [main][main] Listening on 0.0.0.0:8080/
2020/12/02 09:21:51 [watchdog][monitor] Monitored serviceName=frontend; success=true; errors=0; requestDuration=648ms; 
2020/12/02 09:21:51 [watchdog][monitor] Monitored serviceName=backend; success=true; errors=0; requestDuration=218ms; 
2020/12/02 09:21:52 [watchdog][monitor] Monitored serviceName=monitoring; success=true; errors=0; requestDuration=223ms; 
```
```
docker buildx imagetools inspect pentusha/gatus | grep Platform
  Platform:  linux/amd64
  Platform:  linux/arm/v6
  Platform:  linux/arm/v7
  Platform:  linux/arm64
```
I also tried to some other platforms, but build was [failed](https://github.com/Pentusha/gatus/runs/1485076752?check_suite_focus=true) so I keep only ARM.

TwinProduction:
Closed by #55

TwinProduction:
Released in [v1.9.1](https://github.com/TwinProduction/gatus/releases/tag/v1.9.1)

P.S. I've also removed `linux/arm/v6`, as `linux/arm/v7` and `linux/arm64` covers all requested use cases right now. 
This also lowers the time for the action to complete, but I might re-add it if it becomes necessary in the future.

PR #55, opened 2020-12-02
Crossbuild docker image for different architectures

Pentusha:
For issue #54 
In order to use this code, you need to setup secrets in the repository. Repo settings -> Secrets
`DOCKER_USERNAME`
`DOCKER_PASSWORD`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,3 +29,28 @@ jobs:
         uses: codecov/codecov-action@v1.0.14
         with:
           file: ./coverage.txt
+      - name: Get image repository
+        run: echo IMAGE_REPOSITORY=$(echo ${{ github.repository }} | tr '[:upper:]' '[:lower:]') >> $GITHUB_ENV
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+        with:
+          platforms: all
+        - name: Set up Docker Buildx
+          id: buildx
+          uses: docker/setup-buildx-action@v1
+          with:
+            version: latest
+        - name: Available platforms
+          run: echo ${{ steps.buildx.outputs.platforms }}
+        - name: Login to GitHub Packages Docker Registry
+          uses: docker/login-action@v1
+          with:
+            username: ${{ secrets.DOCKER_USERNAME  }}
+            password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Build and push docker image
+        uses: docker/build-push-action@v2
+        with:
+          platforms: linux/amd64,linux/arm/v7
+          pull: true
+          push: true
+          tags: ${{ env.IMAGE_REPOSITORY }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,7 @@ jobs:
       - name: Build and push docker image
         uses: docker/build-push-action@v2
         with:
-          platforms: linux/386,linux/amd64,linux/arm/v6,linux/arm/v7,linux/arm64,linux/ppc64le,linux/s390x
+          platforms: linux/amd64,linux/arm/v6,linux/arm/v7,linux/arm64
           pull: true
           push: true
           tags: ${{ env.IMAGE_REPOSITORY }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,7 +41,7 @@ jobs:
           version: latest
       - name: Available platforms
         run: echo ${{ steps.buildx.outputs.platforms }}
-      - name: Login to GitHub Packages Docker Registry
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -29,3 +28,28 @@ jobs:
         uses: codecov/codecov-action@v1.0.14
         with:
           file: ./coverage.txt
+      - name: Get image repository
+        run: echo IMAGE_REPOSITORY=$(echo ${{ github.repository }} | tr '[:upper:]' '[:lower:]') >> $GITHUB_ENV
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+        with:
+          platforms: all
```
TwinProduction, line 36: Are these steps really necessary?

.github/workflows/build.yml
```diff
@@ -29,3 +28,28 @@ jobs:
...
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+        with:
+          version: latest
+      - name: Available platforms
+        run: echo ${{ steps.buildx.outputs.platforms }}
```
TwinProduction, line 43: Same question as above

.github/workflows/build.yml
```diff
@@ -29,3 +28,28 @@ jobs:
...
+        with:
+          username: ${{ secrets.DOCKER_USERNAME  }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Build and push docker image
+        uses: docker/build-push-action@v2
+        with:
+          platforms: linux/amd64,linux/arm/v6,linux/arm/v7,linux/arm64
+          pull: true
+          push: true
+          tags: ${{ env.IMAGE_REPOSITORY }}
```
TwinProduction, line 55: Mind if we hardcode the repository? Though I appreciate the reusability, I feel like it might be a bit overkill

.github/workflows/build.yml
```diff
@@ -29,3 +28,28 @@ jobs:
...
+          platforms: all
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+        with:
+          version: latest
+      - name: Available platforms
+        run: echo ${{ steps.buildx.outputs.platforms }}
+      - name: Login to GitHub Packages Docker Registry
+        uses: docker/login-action@v1
```
TwinProduction, line 45:
```suggestion
      - name: Login to Docker Registry
        uses: docker/login-action@v1
```
Did you forget to delete `GitHub Packages` by any chance?

TwinProduction (changes requested):
Won't this just create multi-arch images for the latest tag?

I was initially planning on using a manifest instead of buildx and using Docker Hub's build rules to build the images rather than leveraging a GitHub action

.github/workflows/build.yml
Pentusha: QEMU is required. All platforms is not required and you may specify list of platforms, but this step is lasts only 5 seconds.

.github/workflows/build.yml
Pentusha: This step is optional for sure you may customize build step as you like with this list of platforms.

.github/workflows/build.yml
Pentusha: You may also get rid of `IMAGE_REPOSITORY` variable and replace it with `twinproduction/gatus`, but it useful in case someone will create the new fork of your repo.

Pentusha:
> Won't this just create multi-arch images for the latest tag?
It would create image for the latest tag. You may specify your custom tag or list of tags, even git commit.

TwinProduction:
Alright, I've had a closer look and could you make the following changes?:
1. Rename the existing `.github/workflows/build.yml` to `.github/workflows/test.yml`
2. Create a new workflow file (`.github/workflows/docker.yml`) specifically for building the docker images on releases:
```yaml
on:
  release:
    types: [published]
```
That way we should be able to use the `GITHUB_REF` environment variable to extract the tag.
The idea is that I want to only build images for other platforms when I release a new version rather than on the latest version.

Let me know if you need help.

Pentusha:
It seems to be working fine
image

TwinProduction:
Thank you for your contribution @Pentusha, and sorry for the delay. I've had a busy week 😅

TwinProduction (approved):

TwinProduction:
Released in [v1.9.1](https://github.com/TwinProduction/gatus/releases/tag/v1.9.1)

P.S. I've also removed `linux/arm/v6`, as `linux/arm/v7` and `linux/arm64` covers all requested use cases right now. 
This also lowers the time for the action to complete, but I might re-add it if it becomes necessary in the future.

PR #66, opened 2020-12-30
Introduce Postgres persistence layer

### LINKED ISSUE #32: Implement persistence through application restart
At the moment, it seems that the status results are stored in an in-memory map: https://github.com/TwinProduction/gatus/blob/3773f952a80058eb88f48fe9ae9ac51bf1c1efe7/watchdog/watchdog.go#L16

And also limited to only 20 results:
https://github.com/TwinProduction/gatus/blob/3773f952a80058eb88f48fe9ae9ac51bf1c1efe7/watchdog/watchdog.go#L59-L62

It'd be great if these were stored in a persistent data store somehow instead (e.g a database, or files on disk). Whilst Gatus currently only returns the last 20 results, it'd be nice to keep the history to review outages that might have occurred in the past. Storing the results in a database/persistence layer would enable this as the first step.

Of course, the option to retain results in memory should also be kept as it makes Gatus very easy to get up and running.

cjheppell:
Closes #32 

⚠️ This is a relatively large PR. I've tried my best to keep the Git history clean through aggressive rebasing, but ultimately failed to keep up with the upstream commits, so unfortunately some commits in this branch are just "fix rebase problems". Apologies!

# What is this?

This PR introduces a persistence layer for Gatus with a choice of two implementations:
- In memory (identical to the current implementation)
- PostgreSQL database

This is achieved by introducing a storage package to abstract the details of how the data is being stored and retrieved behind suitable interfaces. The storage package contains both the implementations for the in-memory storage and the postgresql storage.

Using Postgres as the persistence layer will ensure that even in the event of a gatus restart (e.g, due to an upgrade or crash) the history of observed status results will be kept.

The in-memory storage acts as before - restarts of gatus will cause the data to be lost.

## In-memory

I've moved the in-memory storage implementation out of the watchdog and into its own file in the storage package. It now also has its own set of tests

## Postgres

I decided to use [Gorm](https://gorm.io/) to act as the ORM for the interface between gatus and the database. The benefit of this is that we don't need to manage our own database migrations manually, as [Gorm offers auto-migration](https://gorm.io/docs/migration.html). This is invoked on application startup, and if we change the storage structs then it will also update the database accordingly.

NOTE: I'm not yet clear how things like field renames will affect the database. Other ORM's can understand this as a drop/create which is a destructive change, so it may be necessary to more carefully handcraft migrations via something like [Flyway](https://flywaydb.org/) if the data (and changes to it) becomes more complex. If it's decided that we wont use auto migrations and instead choose something like Flyway, Gorm can still be used as the package for CRUD operations.

## Extending to other database engines

Gorm is fairly indifferent to the underlying database engine. The only postgres-specific parts are in the `postgres.go` file which is responsible for constructing a Postgres connection. All of the CRUD operations work on Gorm's abstraction layer.

In theory, we should be able to trivially extend support to other relational database engines (e.g MySQL, Microsoft SQL Server, etc) if required.

## Backwards-compatibility

I've tried to ensure this change is backwards compatible. Configuring gatus to use storage is achieved via a new configuration section:

```yaml
# other gatus config here
storage:
  inMemory: false
  connectionString: "your-connection-string"
```

If the storage section isn't present, then gatus will default to the in-memory store. That means it _should_ be possible for all users to update to the latest version of gatus and preserve existing behaviour. Though it'll be possible to opt-in to postgresql storage by adding this new storage section to the config.


cjheppell:
I realise I've missed updating the documentation.

If this change looks sensible, I'll be very happy to add further commits to update the docs, but until then I'll hold off for feedback 😄

TwinProduction:
Hello @cjheppell and thank you for contributing.

As you said, it's quite a large PR, and I wish you'd have mentioned you were going to start working on this so we could have perhaps had a short discussion about the implementation prior to this. 

I'm quite strict with code - both mine and others - so I hope you won't hold it against me 😅

To start off, I see a couple of issues:


#### 1. `GetAll()` fetches everything
That's not scalable. 

For instance, if you have 10k results per ServiceStatus, every single time you call `GetAll`, you'd be retrieving a _lot_ of data, which . Don't get me wrong, 10k entries isn't much; in fact, we could store 100k entries in-memory with no problem; the issue is that for an application that monitors other applications, the action of retrieving 10k entries from an external source means hundreds of thousands of memory allocations, and the fact that these entries are short lived (the result of the GetAll doesn't get saved on the application itself) means a lot of garbage collection, which could eventually impact latency. 

Even if we ignore the performance impact, the number of entries in the database will eventually reach a point where retrieving all the data will cause the application to be killed because it ran out of memory. 

So what's the solution? Probably paging. Paging and the ability to get service statuses by service group + name would fix a couple of the issues.

To clarify, this is mainly because the non-in-memory implementation does not leverage `ServiceStatus.AddResult`'s function, which is what limits the result history to 20 per ServiceStatus.

Not to mention that the UI is currently built to only be able to show 20 status at a time, and the UI doesn't support paging either, so these extra 9980 results would either be useless, or they'd be squished on the UI (and would also probably be quite painful bandwidth-wise). 

Here's a test I just ran:
image
That service has an interval of 1s (obviously a bit too fast for what's reasonable, but this is mainly to simulate letting a service run for an extended period of time). After 5 minutes, the Postgres driver's already complaining about slow queries (>200ms).

This comment is getting quite long now, but long story short, I'm operating under the assumption that this PR's only purpose is to support persistence, and not also support storing more than 20 results per service, as these are two different tasks that have different implications. Both can also be tackled independently; even without persistence, we can store more than 20 results per service. I just haven't gotten around to supporting paging on the dashboard, which would allow me to simply increase the 20 results limit.


#### 2. `GetAll()` copies the entries
`copyResults` and the underlying functions are not really necessary - they'll just create unnecessary memory allocations.


#### 3. No validation
Say you add service `A`, `B`, and `C`, but a few days later, you decide to remove service `B`. When you fetch the data from the database, you'll log a `serviceStatuses` entry for a service that is no longer being monitored, nor updated, however, it will still display the data on the dashboard.



I also have a few other comments:
- Having `ResultGetter` and `ResultStorer` feels redundant. Every single storage medium (in-memory, sql, nosql, file-based, etc.) will have to provide both a way to store and a way to retrieve data, therefore, having a single interface `Storage` with both a function to retrieve and a way to store data should've been sufficient.
- In the `example` folder, you added `storage.inMemory: true` to almost all examples, but isn't that the default behavior when no storage is provided? Therefore, it's a bit redundant to add it in something like the `minimal` example, which is supposed to have the least amount of configuration possible.


I've placed a huge emphasis on performance for Gatus to allow it to be as accurate as possible, so I hope that you can understand that there are a few changes that I can't easily accept.



### TL;DR

There's a couple of issues in the suggested implementation, and I'm not sure it's the best approach. 
Even the usage of Gorm seems overkill here. Ultimately, a better and simpler implementation would be to store the serviceStatuses in a postgres database using the same pattern as the current in-memory implementation (key/value), but converting the value to glob when inserting/retrieving it.

If you're willing to give it another try, I'm open to discussing it with you before deciding on the solution. The most pertinent question I can think of to start this up are:
- Is Postgres a limitation for you? Or does any form of persistence work for you?
  - If it's not, then I will selfishly suggest [TwinProduction/gocache](https://github.com/TwinProduction/gocache), which supports persistence 
  - If it is, then I think vanilla SQL (without Gorm) with a key/value pattern will be easier, but we can discuss it further


P.S. For some reason, it seems like the last time you updated the dependencies in go.mod, the command you ran didn't respect the `replace` rules since `k8s.io/client-go` should've been pinned to `v0.18.14`.

cjheppell:
> As you said, it's quite a large PR, and I wish you'd have mentioned you were going to start working on this so we could have perhaps had a short discussion about the implementation prior to this.

Excellent point! I'll make sure to follow that going forwards. 

I found some spare time over the last few days being stuck in lockdown, so decided to keep myself busy playing around with this implementation. In hindsight, a discussion would've been handy. But I hope this serves as the frame for a discussion nonetheless 😃 

> 1. GetAll() fetches everything [...] That's not scalable.

Indeed. A quick solution would be to (as you say) return only the 20 most recent results in the db for now (and support paging later). That feels like a no-brainer for now and I can make that change very easily.

> copyResults and the underlying functions are not really necessary - they'll just create unnecessary memory allocations.

I think they _are_ necessary aren't they? The in memory store holds the map. If I just return the map, then you've effectively removed the safety provided by the mutexes in the in memory store since now the caller has access to the map without those locks, and that may well be modified by other goroutines.

Returning a _copy_ of the structure is considerably safer, as it means the caller gets the map exactly as it was when it requested it, and is free to make changes to its received result without impacting other threads.

> 3. No validation

Excellent point, and a glaring omission on reflection. It _does_ introduce the question of how to handle these sorts of things going forwards when we consider persistence. We could leave results the database and _only_ return the results for services currently configured to be monitored via the config. 

Alternatively we could delete the records for all no-longer-monitored services on startup which feels preferable.

> Having ResultGetter and ResultStorer feels redundant. [...] having a single interface Storage with both a function to retrieve and a way to store data should've been sufficient.

I'd disagree on this. [Effective Go](https://golang.org/doc/effective_go.html#interfaces_and_types) encourages small interfaces so I've tried to follow this. It also has the benefit of isolating services that only need to _read_ from those that need to _write_.

> serviceStatuses in a postgres database using the same pattern as the current in-memory implementation (key/value), but converting the value to glob when inserting/retrieving it.

It feels worthwhile to take advantage of the relational nature of the database and break down the results into the respective tables as provided by Gorm. For file storage, I'd agree - serialising it to the key/value would be absolutely sufficient.

> Is Postgres a limitation for you? Or does any form of persistence work for you?

Any form works just fine! Postgres is what I have most familiarity with, so I thought it'd be worth giving it a go. Though file storage would be absolutely perfect too - and remove the need for an ORM. Happy to implement it for file storage only first, and introduce relational databases later if needed.

> then I think vanilla SQL (without Gorm) with a key/value pattern will be easier

I would cautiously advise against this. Dumping key/values into a relational database with vanilla SQL sounds appealing, but it can quickly become difficult to manage as you end up having to do a lot of the heavy lifting that an ORM can do for you. It might seem heavy-handed at first, but often is worth it.



I really appreciate the review - thanks for you time 👍 

By the way, Gatus is great. I appreciate your concern for keeping things simple and performant - good to see those traits are being preserved!

cjheppell:
Based on the above, I think it's probably worth dropping the use of a relational database as a supported persistence layer, and implement file-based storage alongside the in-memory storage.

That should remove some of the complexity, and if we do a like-for-like implementation copy (e.g limited to 20 results per service) then this will take it some-way towards a persistence layer (albeit limited for now until something like paging is added).

For that, I'd suggest closing this PR for now and opening a follow-up with file-based storage in place of Postgres. I'll wait until we've reached consensus though 😺

TwinProduction:
I agree with everything you've said @cjheppell, though I could argue that while a small interfaces are more convenient, I usually prefer the smallest common denominator, and I can't see a use case where you'd want to implement `GetAll` without `Store`, and my issue with the `copy` is mostly due to the fact that once the struct is added into the map/persisted, it will no longer be modified, thus, it's pretty safe to assume that even without copying the struct, you'll get the same value -- except for the uptime, which is occasionally recomputed, but even then, the difference would be acceptable. All in all, the mutex is there to prevent concurrent map access, which could cause a panic, but I digress.

As for the implementation, since file-based persistence is sufficient, then I too think that we should shift away from relational databases, for now, at least.

Since it's acceptable to have a small gap in reporting, I don't think that immediate persistence is necessary (i.e. no need to persist every single new piece of data immediately), therefore, having a goroutine in the background that persists the data every X minutes (could also make this configurable) would be perfect, hence why I said gocache might be a good option here since it's already used in the project, but we can continue discussing this in #32 if you'd like.

Once again, thank you for being understanding 😅

cjheppell:
Thanks for the detailed response 👍 

Sounds like we're pretty much on the same page so I'm going to close this PR for now and switch back to the issue for further discussion.

Hopefully this can serve some use in the future if Gatus ends up persisting to a relational database. Though even if not, I'm happy this has started some useful discussion 🙂

Paxa:
Thanks @cjheppell I will try to use your fork

Without persistence we can't event run it in k8s properly

cjheppell:
I'd suggest taking a look at #69 as a branch to use if you're wanting to run in K8S.

This branch is a long way behind the master branch now, but #69 might be _good enough_ for you.

If Postgres persistence is something you need (as opposed to file-based) then I'd encourage you to explain your requirements in https://github.com/TwinProduction/gatus/issues/32 as that's a better place to capture the reasons for needing this.

TwinProduction:
FYI, the first mean of persistence will be merged in the coming days through #82.

It's not immediate persistence, but rather, in-memory with an auto save feature that dumps the necessary data to file every few minutes.

PR #69, opened 2021-01-08
Add file persistence

### LINKED ISSUE #32: Implement persistence through application restart

yarhamjohn:
Closes #32 

Adds an option to persist the cache to a file on disk. Uses https://github.com/TwinProduction/gocache to implement persistence. An example config.yaml file is provided - if no storage config is used the application defaults to existing in-memory storage behaviour. If a config is set, the specified interval determines when the cache is flushed to disk. If not interval is specified, the cache is flushed after every update.

I worked with @cjheppell on this following the discussions on https://github.com/TwinProduction/gatus/pull/66#issuecomment-752800783


storage/memory.go
```diff
@@ -1,62 +1,122 @@
...
+func (ims *Store) WithPersistence(filePath string, interval *time.Duration) *Store {
+	if _, err := os.Stat(filePath); os.IsNotExist(err) {
+		err := ioutil.WriteFile(filePath, []byte{}, 0644)
+		if err != nil {
+			panic(fmt.Sprintf("Could not write to file: %s", err))
+		}
+	} else {
+		numEvictions, err := ims.serviceStatuses.ReadFromFile(filePath)
+		if numEvictions != 0 {
+			panic(fmt.Sprintf("Unexpectedly dropped %d cache entries", numEvictions))
```
TwinProduction, line 47: Is it worth panicking here? This would prevent the application from ever recovering, because the next time the application starts, it'll read the file and probably panic again, looping until manual action is taken.

IMHO, since the usage of persistence on Gatus will likely be for convenience rather than a critical requirement.

I mean, technically, this shouldn't happen, but if it does, how bad can it be? 🤔

TwinProduction (changes requested):
Great work!
Could you add a few tests?

storage/memory.go
```diff
@@ -1,62 +1,122 @@
...
+	serviceStatuses *gocache.Cache
+	interval        *time.Duration
+	filePath        string
+	shouldPersist   bool
 }
 
-// NewInMemoryStore returns an in-memory store. Note that the store acts as a singleton, so although new-ing
+// NewStore returns an in-memory store. Note that the store acts as a singleton, so although new-ing
 // up in-memory stores will give you a unique reference to a struct each time, all structs returned
 // by this function will act on the same in-memory store.
```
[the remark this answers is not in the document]
yarhamjohn: No problem!

storage/memory.go
yarhamjohn: No that is a fair point - it shouldn't happen as the cache is not itself size limited, but panicking does seem a bit excessive

yarhamjohn:
> Could you add a few tests?

We were wondering this as well - it wasn't immediately obvious to us what useful (and practical) unit tests we could write around this that didn't simply test that the `gocache` library was working, since what we would ideally need to test is the behaviour of `gatus` on starting and restarting the application itself. However maybe there is scope for a relatively high-level test around creating a `NewStore().WithPersistence()` - happy to look at that

yarhamjohn:
Not quite sure what is going on with the new tests - they work on my machine but not here. 

image

The error is complaining that `t.TempDir()` requires go version 1.15, but that is the version being used according to the logs. Do you have any idea what might be happening here? It is a bit tricky for me to post a bug on the golang repo as I can't reproduce locally so just hoping there is some build setting causing an issue.

yarhamjohn:
> The error is complaining that `t.TempDir()` requires go version 1.15, but that is the version being used according to the logs. Do you have any idea what might be happening here? It is a bit tricky for me to post a bug on the golang repo as I can't reproduce locally so just hoping there is some build setting causing an issue.

Interestingly, when I avoid using t.TempDir(), I am getting a different error (one I also don't see locally), but the build log seems to be suggesting go 1.14 might actually be being used despite the GOROOT pointing at 1.15

image

TwinProduction:
@yarhamjohn FYI, I resolved the issue with `t.TempDir()`.

Because I was using `sudo` to run the tests (one of the tests requires it), it was using the superuser's `PATH`, which includes a `go` executable that isn't 1.15+.

yarhamjohn:
> @yarhamjohn FYI, I resolved the issue with `t.TempDir()`.
> 
> Because I was using `sudo` to run the tests (one of the tests requires it), it was using the superuser's `PATH`, which includes a `go` executable that isn't 1.15+.

Awesome thanks :)

yarhamjohn:
> @yarhamjohn FYI, I resolved the issue with `t.TempDir()`.
> 
> Because I was using `sudo` to run the tests (one of the tests requires it), it was using the superuser's `PATH`, which includes a `go` executable that isn't 1.15+.

Hmm unfortunately I seem to still be hitting the same issue (I have rebased onto master) with the go version

TwinProduction:
Hello @yarhamjohn, 

First, I'd like to apologize for the delay in answering.
For the past two weeks, I've been devoting my free time to migrating the front end to Vue & Tailwind (and learning them) as well as adding some features to said front end (see release notes for v2.0.0 for more details), and I've been purposely snoozing this notification.

I've been working on persistence, and I've tried and compared several different options:
- memory-only (current implementation)
- file-only (with bolt)
- gocache (memory + file, but persistence is achieved through autosave)

The file-only option was pretty terrible in term of performance, granted it may have been because the implementation itself was a little lazy.

memory-only was used more as a baseline to compare performance, but since the purpose was to achieve performance, it's a no-go.

Finally, gocache gave such good results. It's essentially on-par with memory while providing the persistence required, but well, that's why I suggested using it in the first place.

Anyways, I digress, the point is, I ended up implementing it on my side and making it a bit more convenient for the upcoming support for Postgres. That, coupled with the fact that I added a couple more tests makes me come to the conclusion that I may as well implement the version I worked on.

I did however take some inspiration from your implementation as well as the discussion you had with @cjheppell, so please don't take it to heart - I really didn't mean to make you work on a specific implementation only to end up going for a different implementation 😓

TwinProduction:
Superseded by #82

PR #81, opened 2021-02-01
fix: metrics were not working

menduz:
Yielding a 404 error, raised by the handler in the line above

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -60,10 +60,10 @@ func CreateRouter(cfg *config.Config) *mux.Router {
 	router.HandleFunc("/api/v1/statuses/{key}", secureIfNecessary(cfg, GzipHandlerFunc(serviceStatusHandler))).Methods("GET")
 	router.HandleFunc("/api/v1/badges/uptime/{duration}/{identifier}", badgeHandler).Methods("GET")
 	router.HandleFunc("/health", healthHandler).Methods("GET")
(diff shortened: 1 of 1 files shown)


controller/controller.go
```diff
@@ -60,10 +60,10 @@ func CreateRouter(cfg *config.Config) *mux.Router {
 	router.HandleFunc("/api/v1/statuses/{key}", secureIfNecessary(cfg, GzipHandlerFunc(serviceStatusHandler))).Methods("GET")
 	router.HandleFunc("/api/v1/badges/uptime/{duration}/{identifier}", badgeHandler).Methods("GET")
 	router.HandleFunc("/health", healthHandler).Methods("GET")
-	router.PathPrefix("/").Handler(GzipHandler(http.FileServer(http.Dir("./web/static"))))
```
TwinProduction, line 63: Good catch! 

Would you mind moving the 
```
	if cfg.Metrics {
		router.Handle("/metrics", promhttp.Handler()).Methods("GET")
	}
```
immediately after `router := mux.NewRouter()` instead?
I think the reason I initially moved it down was because it seemed out of place, so if it's at the very top, I doubt I'll make the same mistake again 😅

TwinProduction (changes requested):

controller/controller.go
TwinProduction: On second thoughts, I'm going to add some tests, so I'll take care of moving it.

TwinProduction:
Thanks for the contribution! @menduz

PR #82, opened 2021-02-02
Implement persistence

TwinProduction:


TwinProduction:
@yarhamjohn @cjheppell I'd love to have your feedback on this, if you can spare some time.
I've tested it locally, benchmarked it and added some tests, but an extra pair of eyes or two would be much appreciated.

TwinProduction:
Accidentally bricked the branch; recovered in -> #83

Issue #76, opened 2021-01-22
Make ALERT_TRIGGERED_OR_RESOLVED placeholder values configurable


GeorgFleig:
We're currently using Gatus for triggering health checks and having an internal dashboard. In order to unify with our public status reporting we'd like to use Gatus to set the status of components on our statuspage.io status page.

I've configured a custom alerting provider which works great when a service goes down. But I don't see a way to report it as up again at the moment. Statuspage.io requires different strings than `TRIGGERED` and `RESOLVED` to distinguish the statuses. Currently they are hard coded here `ALERT_TRIGGERED_OR_RESOLVED`: 
https://github.com/TwinProduction/gatus/blob/master/alerting/provider/custom/custom.go#L48

The configuration is as follows. I use the `ALERT_DESCRIPTION` to identify the service name at statuspage.io.
```
alerting:
  custom:
    url: "https://api.statuspage.io/v1/pages/XXX/components/[ALERT_DESCRIPTION]"
    method: "PATCH"
    headers:
      Authorization: "OAuth YYY"
    body: |
      { "component": { "status": "operational" } }

services:
  - name: foo
    url: "https://this.is.down"
    interval: 1m
    conditions:
      - "[STATUS] == 200"
    alerts:
      - type: custom
        enabled: true
        send-on-resolved: true
        description: a9gbjb11q8gy
```

Status of a component could be any of

```
"operational" "under_maintenance" "degraded_performance" "partial_outage" "major_outage" ""
```

(see https://developer.statuspage.io/#operation/patchPagesPageIdComponentsComponentId).

It would be great if the two strings were configurable on `alerting.custom` level, making the custom alert feature even more flexible. Maybe like this, defaulting to the current hardcoded-values:
```
alerting:
  custom:
    url: "https://api.statuspage.io/v1/pages/XXX/components/[ALERT_DESCRIPTION]"
    method: "PATCH"
    triggered_message: "partial_outage"
    resolved_message: "operational"
    headers:
      Authorization: "OAuth YYY"
    body: |
      { "component": { "status": "[ALERT_TRIGGERED_OR_RESOLVED]" } }
```

Gatus is great and keeps getting better. I wanted to say thanks for providing this neat tool and actively maintaining it!

TwinProduction:
Fascinating - so you're using Gatus to push data onto a different status page?
I was originally afraid people wouldn't get creative enough with the `custom` alerting provider, but it looks like my worries were unnecessary.

[...]

PR #85, opened 2021-02-17
Make ALERT_TRIGGERED_OR_RESOLVED placeholder values configurable

roberth1988:
I am a colleague of GeorgFleig which opened issue #76. This pull request  actually implements this feature in the form it was proposed in https://github.com/TwinProduction/gatus/issues/76#issuecomment-765863720

Would be awesome if it get merged. Feedback welcome.

### DIFF
--- a/alerting/provider/custom/custom.go
+++ b/alerting/provider/custom/custom.go
@@ -24,13 +24,6 @@ type AlertProvider struct {
 	Placeholders map[string]map[string]string `yaml:"placeholders,omitempty"`
 }
 
-var DefaultPlaceholderValues = map[string]map[string]string{
-	"ALERT_TRIGGERED_OR_RESOLVED": map[string]string{
-		"triggered": "TRIGGERED",
-		"resolved":  "RESOLVED",
-	},
-}
-
 // IsValid returns whether the provider's configuration is valid
 func (provider *AlertProvider) IsValid() bool {
 	return len(provider.URL) > 0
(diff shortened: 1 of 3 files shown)


alerting/provider/custom/custom.go
```diff
@@ -33,10 +41,28 @@ func (provider *AlertProvider) ToCustomAlertProvider(service *core.Service, aler
 	return provider
 }
 
+// GetPlaceholderValue returns the Placeholder value
+func (provider *AlertProvider) GetPlaceholderValue(name, status string) string {
```
TwinProduction, line 45: Since `ALERT_TRIGGERED_OR_RESOLVED` is the only placeholder which has one value or another, I believe that it's the only one which has a configuration that will be used like that.

As such, could you rename this function to something like `GetAlertStatePlaceholderValue(resolved bool)` which would use hardcoded values only inside that function instead of outside (e.g. `provider.Placeholders["ALERT_TRIGGERED_OR_RESOLVED"]["resolved"]`)?

That way, instead of calling `provider.GetPlaceholderValue("ALERT_TRIGGERED_OR_RESOLVED", "resolved")`, it'd be `provider.GetAlertStatePlaceholderValue(true)`.

TwinProduction (changes requested):

alerting/provider/custom/custom.go
roberth1988: Hi @TwinProduction  => I just did the changes for you :)

TwinProduction (approved):

TwinProduction:

TwinProduction:
FYI, I made some changes and the syntax is now:
```yaml
alerting:
  custom:
    placeholders:
      ALERT_TRIGGERED_OR_RESOLVED:
        TRIGGERED: "partial_outage"
        RESOLVED: "operational"
```

roberth1988:
Thank you :)

TwinProduction:
Released in [v2.2.0](https://github.com/TwinProduction/gatus/releases/tag/v2.2.0)

PR #100, opened 2021-03-13
Truncate long strings when using pattern function

### LINKED ISSUE #98: 'pat' keyword causes various issues
If the `pat` keyword is used to match against the whole body, this causes issues with:

* Persistent storage - The storage file balloons as gatus stores all the responses for a given number of attempts before truncating.
* Alerting providers - The whole contents of the body tries to get send to the alerting provider when the service goes down. Most of the time this causes the alerting provider to fail.

I'd propose to simply not include results of the pattern match, and instead show that the pattern isn't matched as a one-liner statement.

I'm aware that such aggressive use of the `pat` keyword is likely going to degrade performance. However there's little other way to monior off the shelf apps which are expected to return HTML with no dedicated health check endpoint.

dchidell:
Fixes #98 - Unsure if this will be a desirable solution or not, but figured I would at least try and contribute a fix!

### DIFF
--- a/core/condition.go
+++ b/core/condition.go
@@ -284,10 +284,10 @@ func prettify(parameters []string, resolvedParameters []string, operator string)
 	}
 
 	// If using pattern comparison, omit the contents
-	if strings.Contains(parameters[0], PatternFunctionPrefix) && len(resolvedParameters[1]) > 25 {
+	if strings.HasPrefix(parameters[0], PatternFunctionPrefix) && strings.HasSuffix(parameters[0], FunctionSuffix) && len(resolvedParameters[1]) > 25 {
 		resolvedParameters[1] = fmt.Sprintf("%.25s...(truncated)", resolvedParameters[1])
 	}
-	if strings.Contains(parameters[1], PatternFunctionPrefix) && len(resolvedParameters[0]) > 25 {
+	if strings.HasPrefix(parameters[1], PatternFunctionPrefix) && strings.HasSuffix(parameters[1], FunctionSuffix) && len(resolvedParameters[0]) > 25 {
 		resolvedParameters[0] = fmt.Sprintf("%.25s...(truncated)", resolvedParameters[0])
 	}
 
--- a/core/condition.go
+++ b/core/condition.go
@@ -282,6 +282,15 @@ func prettify(parameters []string, resolvedParameters []string, operator string)
 	if strings.HasSuffix(resolvedParameters[0], InvalidConditionElementSuffix) || strings.HasSuffix(resolvedParameters[1], InvalidConditionElementSuffix) {
 		return resolvedParameters[0] + " " + operator + " " + resolvedParameters[1]
 	}
+
+	// If using pattern comparison, omit the contents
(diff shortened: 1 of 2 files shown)


core/condition.go
```diff
@@ -282,6 +282,15 @@ func prettify(parameters []string, resolvedParameters []string, operator string)
 	if strings.HasSuffix(resolvedParameters[0], InvalidConditionElementSuffix) || strings.HasSuffix(resolvedParameters[1], InvalidConditionElementSuffix) {
 		return resolvedParameters[0] + " " + operator + " " + resolvedParameters[1]
 	}
+
+	// If using pattern comparison, omit the contents
+	if strings.Contains(parameters[0], PatternFunctionPrefix) && len(resolvedParameters[1]) > 25 {
+		resolvedParameters[1] = fmt.Sprintf("%.25s...(truncated)", resolvedParameters[1])
+	}
+	if strings.Contains(parameters[1], PatternFunctionPrefix) && len(resolvedParameters[0]) > 25 {
```
TwinProduction, line 290: Instead of using `strings.Contains` to detect `PatternFunctionPrefix`, can you use `strings.HasPrefix(parameters[0], PatternFunctionPrefix) && strings.HasSuffix(first, FunctionSuffix)`?

It's very explicit, but it's just to prevent surprises.

TwinProduction (changes requested):
Beautiful implementation, just a small change and it's perfect.

core/condition_test.go
```diff
@@ -274,6 +274,13 @@ func TestCondition_evaluate(t *testing.T) {
 			ExpectedSuccess: true,
 			ExpectedOutput:  "[BODY] == pat(*<div id=\"user\">john.doe</div>*)",
 		},
+		{
+			Name:            "body-pattern-html-failure",
+			Condition:       Condition("[BODY] == pat(*<div id=\"user\">john.doe</div>*)"),
+			Result:          &Result{body: []byte(`<!DOCTYPE html><html lang="en"><head><meta http-equiv="Content-Type" content="text/html; charset=UTF-8" /></head><body><div id="user">jane.doe</div></body></html>`)},
+			ExpectedSuccess: false,
+			ExpectedOutput:  "[BODY] (<!DOCTYPE html><html lang...(truncated)) == pat(*<div id=\"user\">john.doe</div>*)",
+		},
```
TwinProduction, line 283: Could you add another test case where `pat()` is on the left side of the operator instead of the right side?

e.g. 
```go
        Condition:       Condition("pat(*<div id=\"user\">john.doe</div>*) == [BODY]"),
```

TwinProduction (changes requested):

TwinProduction (approved):

TwinProduction:
Thank you for your contribution!

dchidell:
Will these changes also alleviate the size of the storage file?  I didn't dig deep enough to see if the full result.

TwinProduction:
@dchidell It should, but I've recently made some pretty decent improvements related to persistence; the result body was always persisted, but it's no longer the case.

TwinProduction:
Released in [v2.4.0](https://github.com/TwinProduction/gatus/releases/tag/v2.4.0)

PR #102, opened 2021-03-28
Add Telegram Alerting

jon4hz:
Hi,
I would like to use Gatus very much but I was missing the alarm provider Telegram. That's why I added it.


.gitignore
```diff
@@ -1,4 +1,5 @@
 .idea
 .vscode
 gatus
-db.db
\ No newline at end of file
+db.db
+config/config.yml
```
TwinProduction, line 5:
```suggestion
```

TwinProduction (changes requested):
Beautiful work @jon4hz, just one small change which I think may have been committed by mistake.

.gitignore
jon4hz: Actually this entry was intentional, because the file does not exist in the repository and I created it for testing. But of course I can delete the line again if you want to.

TwinProduction: Fair enough

TwinProduction:
Thank you for your contribution @jon4hz

TwinProduction:
Released in [v2.4.0](https://github.com/TwinProduction/gatus/releases/tag/v2.4.0)

PR #116, opened 2021-05-24
add basic authentication support to HTTP based services

mapl:
I have HTTP Service endpoints which require Basic Authentication when connecting. This PR adds a convenient way to add Basic Authentication to HTTP based services when required.

```
services:
  - name: front-end
    group: core
    url: "https://twinnation.org/health"
    interval: 1m
    conditions:
      - "[STATUS] == 200"
      - "[BODY].status == UP"
      - "[RESPONSE_TIME] < 150"
    security:
      basic:
        username: "gatus-monitoring-user"
        password: "someP@sSw0rd"
```

Not sure if it fits your style properly.

### DIFF
--- a/core/service.go
+++ b/core/service.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"github.com/TwinProduction/gatus/security"
 	"io/ioutil"
 	"net"
 	"net/http"
@@ -65,6 +66,9 @@ type Service struct {
 
 	// Headers of the request
 	Headers map[string]string `yaml:"headers,omitempty"`
+
+	// Basic HTTP Authentication
+	Security *security.Config `yaml:"security"`
 
 	// Interval is the duration to wait between every status check
 	Interval time.Duration `yaml:"interval,omitempty"`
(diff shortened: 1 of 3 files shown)


TwinProduction:
Since you can specify the headers for the request, is there really a need for this?

```yaml
services:
  - name: front-end
    group: core
    url: "https://twinnation.org/health"
    interval: 1m
    headers:
      Authorization: "Z2F0dXMtbW9uaXRvcmluZy11c2VyOnNvbWVQQHNTdzByZA=="
    conditions:
      - "[STATUS] == 200"
      - "[BODY].status == UP"
      - "[RESPONSE_TIME] < 150"
```

Note: `Z2F0dXMtbW9uaXRvcmluZy11c2VyOnNvbWVQQHNTdzByZA==` is the base64 encoded version of `gatus-monitoring-user:someP@sSw0rd`

mapl:
Previously I had used a custom basic authentication header, which worked. 
However, having a specific option to easily set up basic authentication without the need to encode the credential values to base64 value first was a major time saver in multiple situations.
I treat this as a commonly used method when connecting to an API endpoint where authentication is required.

TwinProduction:
Honestly, I think that's a bit too verbose. 

I'd rather keep things simple, and considering that all you need to do is encode `username:password` with base64 and then you can just use:
```
    headers:
      Authorization: "Z2F0dXMtbW9uaXRvcmluZy11c2VyOnNvbWVQQHNTdzByZA=="
```
I feel like this is way too verbose:
```
    security:
      basic:
        username: "gatus-monitoring-user"
        password: "someP@sSw0rd"
```

mapl:
You're welcome :)

Cartache:
> ```yaml
>     headers:
>       Authorization: "Z2F0dXMtbW9uaXRvcmluZy11c2VyOnNvbWVQQHNTdzByZA=="
> ```

Hello,

I tried your solution but had to add "Basic " to the headers in order to make it work.
So to summarize, my configuration to call a protected API Request with Base64 Encoded Basic Authentication should be according to my situation to:

`    headers:
      Authorization: "Basic XXXXXXXXXXXXXXXXXXXXXBase64Encoded=="   `

`    headers:
      Authorization: "Basic user:password" (Base64 Encoded)  `

PR #121, opened 2021-06-05
Add [MODIFIED] condition.

tjhowse:
This change adds a [MODIFIED] condition that can be used to check that the resource specified by the URL changed since last retrieval. It does this by comparing the ETag header in the HTTP response.

My use-case is a heartbeat status file that should be updated every minute if the service is functioning properly.

If this is a feature that the maintainers feel would be useful please let me know and I'll finish off this PR with readme updates and tests.

I am not particularly happy with the means I am using to pass the previous ETag value through to the conditional evaluation code, but I wasn't able to see a way forward that wouldn't require substantial re-architecturing.

### DIFF
--- a/core/condition.go
+++ b/core/condition.go
@@ -36,6 +36,11 @@ const (
 	// Values that could replace the placeholder: {}, {"data":{"name":"john"}}, ...
 	BodyPlaceholder = "[BODY]"
 
+	// ModifiedPlaceholder is a placeholder for the whether the ETag header changed since the previous request
+	//
+	// Values that could replace the placeholder: true, false
+	ModifiedPlaceholder = "[MODIFIED]"
+
 	// ConnectedPlaceholder is a placeholder for whether a connection was successfully established.
 	//
 	// Values that could replace the placeholder: true, false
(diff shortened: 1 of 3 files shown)


TwinProduction:
Sorry for taking so long to respond, it's been a busy couple of weeks and I've been unable to commit time to thinking of a solution for this.

To be honest, I share the same opinion as you in that I'm not fan of the implementation -- and I don't blame you, as every solutions I could think of also has its shortcomings.

I cannot accept the implementation as-is, because it's too targeted to a single use case (ETag) yet uses a very generic placeholder name (`[MODIFIED]`).

I think that the right approach would be to:
- Add support for header conditions (#4)
- Add support for a placeholder that would let us access the previous evaluation results

The two above would enable something like this:
```yaml
services:
  - name: example
    group: core
    url: "https://example.org"
    interval: 1m
    conditions:
      - "[STATUS] == 200"
      - "[HEADER].ETag != [LAST_RESULT][HEADER].ETag"
```

I completely understand if you don't want to undertake this task as its scope is significantly larger, but unfortunately, I cannot accept this PR as it is because the scope of the feature is far too small and I would rather trying to avoid breaking changes in the future since the two suggestions I made above was something I was planning on implementing at some point.

Regardless, thank you for the contribution! I appreciate it :)

PR #175, opened 2021-09-16
Add enabled parameter to service

1newsr:
In some cases, I want to adjust whether or not service is set by an environment variable.

e.g.

```
service:
  name: service-a
  enabled: ${SERVICE_A_ENABLED}
```

So, I added the `enabled` flag to the service config.
If you have any problems with this feature, please let me know.


watchdog/watchdog.go
```diff
@@ -26,9 +26,11 @@ var (
 func Monitor(cfg *config.Config) {
 	ctx, cancelFunc = context.WithCancel(context.Background())
 	for _, service := range cfg.Services {
-		// To prevent multiple requests from running at the same time, we'll wait for a little bit before each iteration
-		time.Sleep(1111 * time.Millisecond)
-		go monitor(service, cfg.Alerting, cfg.DisableMonitoringLock, cfg.Metrics, cfg.Debug, ctx)
+		if *service.Enabled {
```
TwinProduction, line 29: Could you add a method `IsEnabled` instead and use that?

For an example, see https://github.com/TwinProduction/gatus/blob/18420c2d60d1190510f7e02661102b4bc7e0c2ac/alerting/alert/alert.go#L56

Call me paranoid; I'm aware that `ValidateAndSetDefaults` will be called before this, but I'd rather do this to avoid a possible panic due to a code change in the future.

TwinProduction (changes requested):

watchdog/watchdog.go
1newsr: Thank you for your review. Fixed in 8d4b3d3.

TwinProduction (approved):
Looks good! Thank you for your contribution @1newsr!

TwinProduction:
Released in [v3.2.0](https://github.com/TwinProduction/gatus/releases/tag/v3.2.0)

PR #177, opened 2021-09-23
Health check for SSL/TLS services

Carlotronics:
- Add SSL service health check;
- Add timeout feature for StartTLS checks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ For more details, see [Usage](#usage)
   - [Monitoring a service using ICMP](#monitoring-a-service-using-icmp)
   - [Monitoring a service using DNS queries](#monitoring-a-service-using-dns-queries)
   - [Monitoring a service using STARTTLS](#monitoring-a-service-using-starttls)
+  - [Monitoring a service using TLS](#monitoring-a-service-using-tls)
   - [Basic authentication](#basic-authentication)
   - [disable-monitoring-lock](#disable-monitoring-lock)
   - [Reloading configuration on the fly](#reloading-configuration-on-the-fly)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -963,6 +964,24 @@ services:
...
+        timeout: 5s
+    conditions:
+      - "[CONNECTED] == true"
+      - "[CERTIFICATE_EXPIRATION] > 48h"
+```
+
+
+### Monitoring a service using SSL
+If you have an email server that you want to ensure there are no problems with, monitoring it through SSL
+will serve as a good initial indicator:
```
TwinProduction, line 977: Any chance you could update the description? It's pretty much the same as the description for `STARTTLS`.

Also, is it really accurate to make the protocol `ssl://` when it's in fact `tls`?

client/client.go
```diff
@@ -57,6 +61,19 @@ func CanPerformStartTLS(address string, config *Config) (connected bool, certifi
...
+// CanPerformSSL checks whether a connection can be established to an address using the SSL protocol
+func CanPerformSSL(address string, config *Config) (connected bool, certificate *x509.Certificate, err error) {
+	conn, err := tls.DialWithDialer(&net.Dialer{Timeout: config.Timeout}, "tcp", address, nil)
+	if err != nil {
+	    return
+	}
+	defer conn.Close()
+
+	chain := conn.ConnectionState().VerifiedChains[0] // VerifiedChains[0] == PeerCertificates[0]
+	certificate = chain[0]
```
TwinProduction, line 73: Would it be possible to check the length before calling these?

I'd rather avoid risking a `panic`.

TwinProduction (changes requested):

README.md
Carlotronics: > Also, is it really accurate to make the protocol ssl:// when it's in fact tls?

You're right, `tls://` would be more accurate. I've updated it.

client/client.go
Carlotronics: Sure. I've added both checks in c55d680.

Carlotronics:
@TwinProduction please tell me when you'll be ready to merge changes so I can squash last commits. I think it is clearer to keep them distinct for now but they won't be useful anymore in main tree.

TwinProduction:
@Carlotronics Sorry about that, looks like the test failed because of a feature I recently added.

I just fixed it in 428e415; could you pull the changes from master?

TwinProduction (approved):

TwinProduction:
Thank you for your contribution!

TwinProduction:
Released in [v3.2.0](https://github.com/TwinProduction/gatus/releases/tag/v3.2.0)

PR #178, opened 2021-09-23
WIP: add headers support

### LINKED ISSUE #4: Support header conditions
Gatus should provide a way to add conditions to validate the value of a header.

Here's what it could look like:
```yaml
  - name: example
    url: "https://example.org/"
    interval: 30s
    conditions:
      - "[HEADER][Content-Encoding] == gzip"
```

Note that thanks to RFC 2616, headers must be case insensitive.

Carlotronics:
Fixes #4 and #121.

WIP: Still needs to find a way to compare given value to each header values (currently only the first one is taken)

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,8 @@ Here are some examples of conditions you can use:
 | `[BODY].name == pat(john*)`  | String at JSONPath `$.name` matches pattern `john*`     | `{"name":"john.doe"}`      | `{"name":"bob"}` |
 | `[BODY].id == any(1, 2)`     | Value at JSONPath `$.id` is equal to `1` or `2`         | 1, 2                       | 3, 4, 5 |
 | `[CERTIFICATE_EXPIRATION] > 48h` | Certificate expiration is more than 48h away        | 49h, 50h, 123h             | 1h, 24h, ... |
+| `[HEADER].Content-Type == text/html` | HTTP response header `Content-Type` must be `text/html` | `text/html`        | `application/json`, ...  |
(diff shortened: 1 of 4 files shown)


TwinProduction:
Personally, I would prefer if you just retrieved each header using `response.Header.Get()`, put it in a map, and call it a day.
Out of all the time I've used Go, this has worked for all of my use cases, and surely it'll be enough to cover most use cases that everybody else may have.

Carlotronics:
The problem with `response.Header.Get()` is that you only get first header value. It will be enough in most cases, but not when working with response such as
```
HTTP/1.1 200 OK
[...]
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6)
User-Agent: AppleWebKit/537.36 (KHTML, like Gecko)
```

From [RFC7230](https://www.rfc-editor.org/rfc/rfc7230.html#section-3.2.2):
>    A recipient MAY combine multiple header fields with the same field
   name into one "field-name: field-value" pair, without changing the
   semantics of the message, by appending each subsequent field value to
   the combined field value in order, separated by a comma.

Means that joining `response.Header[headerKey]` using a comma should not break any value (except for `Set-Cookie` which I should treat as a special case).

So the way I will implement it depends on the question: is it better to handle all cases and keep full `response.Header[headerKey]`, or voluntarily exclude some of them for the sake of simplicity?

TwinProduction:
Personally, I think we're fine with just checking the first header and keeping it at that.

If somebody opens up an issue in the future, that'll be a problem for then and there :P

TwiN:
Since there has been no changes to this PR for over 6 months, I'm going to close it.
Thank you for the attempt, though, I appreciate it!

PR #186, opened 2021-10-15
WIP: Embed files into the binary

### LINKED ISSUE #47: feat: Bundle assets in binary
Actually, assets are loaded using a relative path in the repository:

https://github.com/TwinProduction/gatus/blob/76d45d7eb8bca0dcf01e503234832d45e8aee884/main.go#L34

My proposal is to bundle assets into the `gatus` binary, to ease distribution with a single binary.

This can probably be done with [go-bindata](https://github.com/shuLhan/go-bindata).

RasmusLindroth:
This would fix #47. But will need some more work and discussion.

### Work that needs to be done

* Update the Dockerfile. I'm no good with docker even though it should be an easy fix

### Things to consider

* Should you still be able to use the "old" way? One advantage is that it lets people change the looks of gatus without re-building the binary.

### Modifications besides loading files using embed instead of from a directory.

* In `spa.go` it will no longer write to the client on errors. This could be fixed, but I don't think it's necessary. If it still runs even when there's an error I think it only hides the error.

* I removed` favicon.go` and `router.HandleFunc("/favicon.ico", FavIcon(staticFolder)).Methods("GET")` because the catch all route can handle it.


TwiN:
There has been too many changes since this PR was created, and it's been 10 months since this PR has been set to draft, so I'm going to close this.

Regardless, I do appreciate the effort! Thanks!

PR #187, opened 2021-10-18
Introduce first step to have a longer retention

roberth1988:
We would like to be able to keep the Uptime for one year. So a first step would be making the uptimeRetention configureable.


roberth1988:
@TwiN merged in latest master plus adjusted things based on your feedback. Would be great to have this merged :)

TwiN:
To be honest, I think this adds more confusion than anything.

```yaml
storage:
  retention:
    days: 10
```
The above makes it sound like it's changing the retention for everything, but it only impacts the uptime retention. This one's on me since I asked for that change without looking into the PR too deeply -- sorry about that.

Disregarding the previous point, this change doesn't really do anything useful, since the only way that the uptime is currently exposed is through the uptime badges. Unfortunately, these uptime badges are hardcoded to only allow 1h, 24h and 7d, which means that even if the retention was set to 365, it wouldn't change anything.

I understand that the purpose is simply to introduce the first steps, but this is more likely to get in the way when I implement the full feature in case some other factors need to be taken into consideration 😅 I would rather have a full implementation, or nothing at all, otherwise maintaining the project might become too complex.

For that reason, I'm going to close the PR, I hope you understand.
Thank you for contributing, though! ❤️

Issue #188, opened 2021-10-18
`Failed to send an serviceAlert` errors due to unescaped quotes in the request body


tiwood:
I've noticed that some of our alerts where failing with the following error message:

```
Failed to send an serviceAlert for service=Very Cool Service "ServiceYXZ": call to provider alert returned status code 400: Bad payload received by generic incoming webhook.
```

It seems the issue is caused by unescaped quotes (`"`) in the HTTP POST - in this specific case in the `text` property:

```json
{
  "@type": "MessageCard",
  "@context": "http://schema.org/extensions",
  "themeColor": "#DD0000",
  "title": "&#x1F6A8; Gatus",
  "text": "An alert for *Very Cool Service "ServiceYXZ" * has been triggered due to having failed 3 time(s) in a row:\n> The health check failed",
  "sections": [
    {
      "activityTitle": "URL",
      "text": "yxz"
    },
    {
      "activityTitle": "Condition results",
      "text": "&#x2705; - `[STATUS] == 200`<br/>&#x274C; - `[BODY].value.availabilityState (Available) == Degraded`<br/>"
    }
  ]
}
```

PR #193, opened 2021-10-23
WIP: escape quotes in alert http request body

jrwren:


TwiN:
Sorry about the delay, my hands have been full with other projects for the past few weeks.

I'm not exactly sure what benefits this brings -- let me explain:
1. This only happens if a user is willingly using quotes in their alert description
2. If they are willingly using quotes, they most likely would not know about this new parameter `escape-body-quotes`.
3. If they do know about the `escape-body-quotes`, then they could just escape their alert descriptions or use single quotes (`'`) and achieve the same result.
4. To conclude, this feature doesn't protect users like #188 from running into this problem, unless they explicitly know that they have to escape double quotes.

Honestly, I would rather have something that prevents alert descriptions from having `"` when the alerting configuration is being validated on application start:
 https://github.com/TwiN/gatus/blob/9287e2f9e2bf51fb01f2a9acccbfe7685cb57544/config/config.go#L269

Same with the endpoint name and group (because honestly, having a `"` in your endpoint name/group is a recipe for disaster 🤣)

TL;DR: I think failing on validation will provide more value than adding a field that would only be used if the user knew about the escape requirement beforehand.

jrwren:
Makes sense to me.

Issue #199, opened 2021-11-07
Include group in alert message


LouisMT:
My Gatus instance is monitoring the same service on multiple servers. I've grouped everything per server which makes everything very clear on the dashboard. 

However I just received an alert on Telegram and it didn't include the group in the message. So I still had to open the dashboard to check which server the message is about.

It would be great if the group could be included in the message!

PR #220, opened 2021-12-10
Include group in alert message for slack provider

umputun:
This is similar to #199 but for slack. It prepends the group name to the message

### DIFF
--- a/alerting/provider/slack/slack.go
+++ b/alerting/provider/slack/slack.go
@@ -46,11 +46,15 @@ func (provider *AlertProvider) Send(endpoint *core.Endpoint, alert *alert.Alert,
 // buildRequestBody builds the request body for the provider
 func (provider *AlertProvider) buildRequestBody(endpoint *core.Endpoint, alert *alert.Alert, result *core.Result, resolved bool) string {
 	var message, color, results string
+	name := endpoint.Name
+	if endpoint.Group != "" {
+		name = endpoint.Group + ":" + name
+	}
 	if resolved {
-		message = fmt.Sprintf("An alert for *%s* has been resolved after passing successfully %d time(s) in a row", endpoint.Name, alert.SuccessThreshold)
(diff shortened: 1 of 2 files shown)


alerting/provider/slack/slack.go
```diff
@@ -46,11 +46,15 @@ func (provider *AlertProvider) Send(endpoint *core.Endpoint, alert *alert.Alert,
...
+	if endpoint.Group != "" {
+		name = endpoint.Group + ":" + name
+	}
 	if resolved {
-		message = fmt.Sprintf("An alert for *%s* has been resolved after passing successfully %d time(s) in a row", endpoint.Name, alert.SuccessThreshold)
+		message = fmt.Sprintf("An alert for *%s* has been resolved after passing successfully %d time(s) in a row", name, alert.SuccessThreshold)
 		color = "#36A64F"
 	} else {
-		message = fmt.Sprintf("An alert for *%s* has been triggered due to having failed %d time(s) in a row", endpoint.Name, alert.FailureThreshold)
+		message = fmt.Sprintf("An alert for *%s* has been triggered due to having failed %d time(s) in a row", name, alert.FailureThreshold)
```
TwiN, line 57: Personally, I'd prefer having something like 
```
message = fmt.Sprintf("An alert for *%s* in group *%s* has been resolved after passing successfully %d time(s) in a row", endpoint.Name, endpoint.Group, alert.SuccessThreshold)
```

I started working on adding this in the other providers on my local branch :P

TwiN (changes requested):

alerting/provider/slack/slack.go
TwiN: On second thoughts, I'm a little worried about what group names people are using.
If any of these group names are using `\` or `"`, it might cause some parsing issues.

I have to head out, but I'll give it some thought.

TwiN: Did some thinking and I decided to implement it in every alerting providers, as well as add validation to prevent the usage of `"` and `\` in endpoint names, groups and alert descriptions.

Sorry for stealing your PR! 😅

TwiN:

PR #223, opened 2021-12-14
Change metrics exposed

### LINKED ISSUE #209: How to consume metrics
I'm struggling with querying and visualizing the exposed metrics in grafana. Ideally I convert them to something useful like big green/red lights that indicate for every endpoint if it is up or not, and I can set up alerting based on the prometheus metrics as well, just to have everything in one place.

Do you have any examples or reference material I could look at to learn more?

In particular what I find challenging is that the `success` property is a label, instead of the value, of the time series. So there actually appear to be two series (success true, success false) for a single endpoint, which is really confusing to me. How do I properly query/consume this?

TwiN:
Before:
```
gatus_tasks{endpoint="blog-home",service="blog-home",status="200",success="true",url="http://blog.default:8080/"} 3108
```

After:
```
gatus_results_total{group="",key="_example",name="example",success="true"} 60
```

This is a breaking change for those using the metrics.

I've also updated the example with Prometheus & Grafana
image

Closes #209


TwiN:

Issue #222, opened 2021-12-13
[BUG] - Maintenance doesn't work


zeylos:
Hi there,

I'm trying to use the maintenance as stated in the doc but I can't make it work.
Here is my config : 
```
    - alerts:
        - type: teams
          failure-threshold: 1
          success-threshold: 1
      conditions:
        - '[STATUS] == 201'
      group: test
      interval: 10m
      maintenance:
        start: 12:00
        duration: 1h
      name: test-maintenance
      url: https://www.google.com
```
I'm in UTC+1 timezone so I should be in the maintenance range but I still receive the alerts as you can see below : 
![image](https://user-images.githubusercontent.com/24226163/145812338-d8d59f92-9ace-4f07-ba12-aaaa5e5270b3.png)

I also tried to start at 13:00 but it still triggers.
There is no error in the container logs, I only have the trigger and alert sending line.

I'm using gatus on kubernetes pulling `v3.4.0`.

Any idea what's going on or how I could debug that ?

Thanks !
Zeylos

TwiN:
Hello @zeylos, the maintenance configuration is at the root level, not at the endpoint level.

zeylos:
Hi @TwiN, I'm so sorry while reading the doc I assumed it was at endpoint level for some reason so I didn't test at root level.
Maybe we can put it in the main Configuration table in the README so idiots like me don't make issues :D ?

Thanks for the answer

TwiN:
In fact, you've got a really good point, there's no row for `Maintenance` in the configuration table like the one for `Alerting` 😔 
![image](https://user-images.githubusercontent.com/15699766/145917952-4f4e78c8-cafb-495d-ad09-84afb479194e.png)

Would you mind creating a PR? 😄

zeylos:
Sure, will do tomorrow :)
Btw sorry for the no reply under the smtp request, just saw your comment today :x

TwiN:
No worries 🤣

PR #226, opened 2021-12-14
Changed: add maintenance to README on root config table

zeylos:
Add maintenance to root config table on README to avoid confusion.

#222


TwiN:
Thank you!

PR #233, opened 2022-01-09
feat(security): Implement Bcrypt for basic auth and deprecate SHA512

TwiN:
I've re-written the code for this several times in the past before but always ended up not going all the way through with it because a hashed Bcrypt string has dollar signs in it, which caused issues with the config file due to environment variable support.

I finally decided to go through with it by forcing users to base64 encode the bcrypt hash.

--- 

The bottom line is, it was irresponsible of me to add support for SHA512 in the first place; it is a fast hash, and is not intended to be used as password.


TwiN:

PR #241, opened 2022-01-23
chore: Fix grammatical issues in README

ItsAzaria:
# Description of Problem
At several places throughout the readme document, there appears to be incorrect grammar. 

# This PR
This PR should fix the problem by correcting all instances of correct grammar in the readme.

# Changes made

`if there’s no traffic that makes it to your applications.` may be considered wordy. 
Replaced with `if no traffic makes it to your applications.`

`not create another incident, but mark the incident ` It appears you have an unnecessary comma in a compound predicate.
Replaced with `not create another incident but mark the incident`

`As a rule of the thumb` It seems there is an article usage problem here.
Replaced with `As a rule of thumb`

`Gatus can automatically generate a SVG badge for one of your monitored endpoints.` The article `a` may be incorrect. Changed to agree with the beginning sound of the word `SVG`
Replaced with `Gatus can automatically generate an SVG badge for one of your monitored endpoints.`

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ monitor these features and potentially alert you before any clients are impacted
 
 A sign you may want to look into Gatus is by simply asking yourself whether you'd receive an alert if your load balancer
 was to go down right now. Will any of your existing alerts be triggered? Your metrics won’t report an increase in errors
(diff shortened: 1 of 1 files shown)


TwiN:
@ItsAzaria Thank you for your contribution!

PR #259, opened 2022-03-07
feat(client): OAuth2 Client credential support

tiwood:
## Background

Some of our endpoints require authentication before the status can be discovered.

Here are two of our current use cases:
- We're using the Microsoft Azure Availability Status API to check if the underlying Azure resources are healthy. 
- We have pre-authentication enabled on some of our ingress controllers - you can't access without a specific cookie or auth header

Right now, we've implemented a simple API that acts as a proxy between our applications and Gatus, but we would like to remove this additional complexity.

## Proposed solution

This PR adds a new endpoint configuration option `oidc`, which tells Gatus to retrieve a `Bearer token` and present it via the `Authorization` header to the endpoint. (see https://auth0.com/docs/get-started/authentication-and-authorization-flow/client-credentials-flow)

This is an example that would work with Azure AD as identity provider:

```yaml
endpoints:
  - name: test
    url: https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000resourceGroups/RG-test/providers/Microsoft.Network/expressRouteCircuits/my-circuit/providers/Microsoft.ResourceHealth/availabilityStatuses/current?api-version=2018-07-01
    interval: 1m
    conditions:
      - "[BODY].properties.availabilityState == Available"
    oidc:
      issuer-url: https://login.microsoftonline.com/00000000-0000-0000-0000-000000000000/oauth2/v2.0/token
      client-id: 00000000-0000-0000-0000-000000000000
      client-secret: secret sauce
      scopes:
        - 'https://management.azure.com/.default'
```

@TwiN, I would love to hear your thoughts about this functionality and some the questions I had while implementing:

* Is this something you want to add to Gatus?
* Technically, `oidc` is not the best naming for feature, but I thought it may be a good idea to keep the same naming as the front end auth, WDYT?
* How should we implement testing? Do you have a test environment which can be used?


TwiN:
@tiwood Looks great to me!

I do think `oauth2` would make more sense than `oidc` here though (correct me if I'm wrong).

Also, do you think it would not make more sense to put it under `endpoints[].client`?

Maybe something like:
```yaml
endpoints:
  - name: test
    url: https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000resourceGroups/RG-test/providers/Microsoft.Network/expressRouteCircuits/my-circuit/providers/Microsoft.ResourceHealth/availabilityStatuses/current?api-version=2018-07-01
    interval: 1m
    client:
      oauth2:
        issuer-url: https://login.microsoftonline.com/00000000-0000-0000-0000-000000000000/oauth2/v2.0/token
        client-id: 00000000-0000-0000-0000-000000000000
        client-secret: secret sauce
        scopes:
          - 'https://management.azure.com/.default'
    conditions:
      - "[BODY].properties.availabilityState == Available"
```

This would have the nice side effect of also making your implementation available for the `custom` alerting provider, as it also uses the the `client.Config` struct.

https://github.com/TwiN/gatus/blob/fcf046cbe8b7dc36824d8655bf74a908be61accd/client/config.go#L29-L40

You'd probably need to add an extra step after this to retrieve the OAuth2 token if `endpoint.ClientConfig.OAuth2` is non-nil here: 
https://github.com/TwiN/gatus/blob/fcf046cbe8b7dc36824d8655bf74a908be61accd/core/endpoint.go#L258

It may also permit testing, as `client.GetHTTPClient` already supports injecting an HTTP client for testing purposes:
https://github.com/TwiN/gatus/blob/fcf046cbe8b7dc36824d8655bf74a908be61accd/client/client.go#L18
For examples on how to use it, see https://github.com/TwiN/gatus/blob/fcf046cbe8b7dc36824d8655bf74a908be61accd/alerting/provider/custom/custom_test.go#L42

tiwood:
Thanks for the great feedback @TwiN. 

The implementation now returns a modified HTTP client which will automatically obtain/refresh tokens as necessary

TwiN (approved):
Looks wonderful! Awesome work!

TwiN:
Thank you for your contribution @tiwood!

TwiN:
Released in [v3.7.0](https://github.com/TwiN/gatus/releases/tag/v3.7.0)

aliamin7:
Hello, thanks for your contribution :)

Do you confirm this only accepts the Client Credential grant type which is not supported by Google APIs today :(

TwiN:
@aliamin7 It's possible. I don't personally use that feature, so I can't confirm nor deny, but I welcome you to attempt implementing it and create a PR for the feature if you'd like to give it a shot!

If not, you can create an issue and perhaps somebody will attempt implementing it.

PR #267, opened 2022-03-21
feat(api): Configurable response time badge thresholds

Fry2412:
Thanks to the conditions on some routes, we can also carry out quasi-functional tests through realistic queries. However, these can take several hundred milliseconds, and we consider the response time badges to be allowed to be green. Therefore, we have tried to find a way to make the threshold values configurable.
We are not very familiar with go lang. Maybe there are some suggestions how to get the config in the badge generator.

# Solution

``` 
endpoints:
  - name: front-end
    group: core
    url: "https://twin.sh/health"
    interval: 5m
    conditions:
      - "[STATUS] == 200"
      - "[BODY].status == UP"
    ui:
      response-timer-badge-awesome: 550
      response-timer-badge-great: 850
      response-timer-badge-good: 1350
      response-timer-badge-passable: 1650
      response-timer-badge-bad: 1750
```

image


config.yaml
```diff
@@ -6,7 +6,12 @@ endpoints:
     conditions:
       - "[STATUS] == 200"
       - "[BODY].status == UP"
-      - "[RESPONSE_TIME] < 150"
+    ui:
+      response-timer-badge-awesome: 550
+      response-timer-badge-great: 850
+      response-timer-badge-good: 1350
+      response-timer-badge-passable: 1650
+      response-timer-badge-bad: 1750
```
TwiN, line 14: Does it really make sense for this to be under an endpoint configuration instead of the root `ui` configuration?

controller/controller_test.go
```diff
@@ -32,7 +32,7 @@ func TestHandle(t *testing.T) {
 	_ = os.Setenv("ROUTER_TEST", "true")
 	_ = os.Setenv("ENVIRONMENT", "dev")
 	defer os.Clearenv()
-	Handle(cfg.Security, cfg.Web, cfg.UI, cfg.Metrics)
+	Handle(cfg.Security, cfg.Web, cfg.UI, cfg, cfg.Metrics)
```
TwiN, line 35: Assuming the answer to my previous comment is `no`, this change would be unnecessary.
If not, then we'd just need to pass `cfg`, no need to pass `cfg.Security`,`cfg.Web`, `cfg.UI` and `cfg.Metrics` since they're all under `cfg` anyways.

TwiN (changes requested):

config.yaml
jesibu: Hi, we worked together on that change. 
The reason to configure this within the single endpoint was, that every endpoint could have other values. Some endpoints are fine with the default values and some have special needs in terms of an "awesome" response time. So we decided to det the values per endpoint.

controller/controller_test.go
jesibu: Yes this definitely needs to be refactored

Fry2412: Since we want to set the endpoints dedicated, we thought it would be better placed under the `endpoints.UIConfig`
I refactored the code a bit and reduced the config parameters to the minimum.

config.yaml
TwiN: Makes sense!

Regardless, I'd rather have a configuration that looks like this:
```yaml
    ui:
      badge:
        response-time:
          thresholds: [550, 850, 1350, 1650, 1750]
```

It's not as intuitive/readable as the solution you've suggested, however, it makes it more future proof by making it easier to add more parameters under a single type of badge, e.g.:
```yaml
    ui:
      badge:
        response-time:
          thresholds: [550, 850, 1350, 1650, 1750]
          theme: modern
```
and it also makes it take a little less space consuming if we were to add a similar feature for a different badge in the future, e.g.:
```yaml
    ui:
      badge:
        response-time:
          thresholds: [550, 850, 1350, 1650, 1750]
        uptime:
          thresholds: [99.9, 99, 95, 90, 75]
```

TwiN (changes requested):

controller/handler/chart_test.go
```diff
@@ -30,7 +30,7 @@ func TestResponseTimeChart(t *testing.T) {
 	}
 	watchdog.UpdateEndpointStatuses(cfg.Endpoints[0], &core.Result{Success: true, Duration: time.Millisecond, Timestamp: time.Now()})
 	watchdog.UpdateEndpointStatuses(cfg.Endpoints[1], &core.Result{Success: false, Duration: time.Second, Timestamp: time.Now()})
-	router := CreateRouter("../../web/static", cfg.Security, nil, cfg.Metrics)
+	router := CreateRouter("../../web/static", cfg, cfg.Metrics)
```
TwiN, line 33: The `cfg.Metrics` feels a bit redundant if we're already passing `cfg` 🤔

core/ui/ui.go
```diff
@@ -1,17 +1,58 @@
...
+				config.Badge.Responsetime.Thresholds[2] {
+				if config.Badge.Responsetime.Thresholds[2] >
+					config.Badge.Responsetime.Thresholds[1] {
+					if config.Badge.Responsetime.Thresholds[1] >
+						config.Badge.Responsetime.Thresholds[0] {
+						return nil
+					}
+				}
+			}
+		}
```
TwiN, line 41: Now that we have a slice instead of separate fields, we should be able to just do a for loop instead of this :P

core/ui/ui.go
```diff
@@ -1,17 +1,58 @@
...
 }
 
 // GetDefaultConfig retrieves the default UI configuration
 func GetDefaultConfig() *Config {
 	return &Config{
 		HideHostname:                false,
 		DontResolveFailedConditions: false,
+		Badge: &Badge{
+			Responsetime: &Thresholds{
+				Thresholds: []int{50, 200, 300, 500, 75},
```
TwiN, line 54:
```suggestion
				Thresholds: []int{50, 200, 300, 500, 750},
```

core/ui/ui.go
```diff
@@ -1,17 +1,58 @@
...
 	// HideHostname whether to hide the hostname in the Result
 	HideHostname bool `yaml:"hide-hostname"`
 	// DontResolveFailedConditions whether to resolve failed conditions in the Result for display in the UI
-	DontResolveFailedConditions bool `yaml:"dont-resolve-failed-conditions"`
+	DontResolveFailedConditions bool   `yaml:"dont-resolve-failed-conditions"`
+	Badge                       *Badge `yaml:"badge"`
+}
+
+type Badge struct {
+	Responsetime *Thresholds `yaml:"response-time"`
```
TwiN, line 15: `Responsetime` should be `ResponseTime` here

README.md
```diff
@@ -174,6 +174,7 @@ If you want to test it locally, see [Docker](#docker).
 | `endpoints[].ui`                                | UI configuration at the endpoint level.                                                                                                     | `{}`                       |
 | `endpoints[].ui.hide-hostname`                  | Whether to include the hostname in the result.                                                                                              | `false`                    |
 | `endpoints[].ui.dont-resolve-failed-conditions` | Whether to resolve failed conditions for the UI.                                                                                            | `false`                    |
+| `endpoints[].ui.badge.reponse-timer.thresholds` | Threshold array for response time badge color grading (values have to raise in an ascending order)                                          | `[50, 200, 300, 500, 750]` |
```
TwiN, line 177:
```suggestion
| `endpoints[].ui.badge.reponse-time.thresholds`  | List of response time thresholds. Each time a threshold is reached, the badge has a different color.                                        | `[50, 200, 300, 500, 750]` |
```

config.yaml
```diff
@@ -6,7 +6,10 @@ endpoints:
     conditions:
       - "[STATUS] == 200"
       - "[BODY].status == UP"
-      - "[RESPONSE_TIME] < 150"
+    ui:
+      badge:
+        response-time:
+          thresholds: [500, 850, 1350, 1650, 1750]
```
TwiN, line 12: As I consider this configuration a bit of an edge case, I'd rather not have it in the default `config.yaml`.

TwiN (changes requested):

config.yaml
```diff
@@ -6,10 +6,7 @@ endpoints:
     conditions:
       - "[STATUS] == 200"
       - "[BODY].status == UP"
-    ui:
-      badge:
-        response-time:
-          thresholds: [500, 850, 1350, 1650, 1750]
+
```
TwiN, line 9:
```suggestion
```

controller/handler/badge.go
```diff
@@ -216,17 +216,14 @@ func getBadgeColorFromResponseTime(responseTime int, key string, config *config.
 
 	group, name, _ := store.Get().GetGroupAndNameByKey(key)
 	ep := config.GetEndpointByGroupAndName(group, name)
+	colors := []string{badgeColorHexAwesome, badgeColorHexGreat, badgeColorHexGood, badgeColorHexPassable, badgeColorHexBad}
```
TwiN, line 219: This should probably be moved to a variable outside of this function, as it will be re-instantiated many times with the exact same values.

TwiN (changes requested):
Almost there! 🙏

config.yaml
```diff
@@ -6,7 +6,6 @@ endpoints:
     conditions:
       - "[STATUS] == 200"
       - "[BODY].status == UP"
-      - "[RESPONSE_TIME] < 150"
```
TwiN, line 9: Can you re-add this? 😅

README.md
```diff
@@ -189,6 +190,7 @@ If you want to test it locally, see [Docker](#docker).
 | `maintenance`                                   | [Maintenance configuration](#maintenance).                                                                                                  | `{}`                       |
 
 
+
```
TwiN, line 193:
```suggestion
```

TwiN (changes requested):

TwiN (approved):

storage/store/sql/sql.go
```diff
@@ -137,6 +137,18 @@ func (s *Store) GetEndpointStatusByKey(key string, params *paging.EndpointStatus
...
+func (s *Store) GetGroupAndNameByKey(key string) (string, string, error) {
+	tx, err := s.db.Begin()
+	if err != nil {
+		return "", "", err
+	}
+	_, group, name, err := s.getEndpointIDGroupAndNameByKey(tx, key)
+	if err == nil {
+		return group, name, nil
+	}
+	return "", "", err
```
TwiN, line 149:
```suggestion
	_, group, name, err := s.getEndpointIDGroupAndNameByKey(tx, key)
	if err != nil {
		return "", "", err
	}
	return group, name, nil
```

README.md
```diff
@@ -174,6 +174,7 @@ If you want to test it locally, see [Docker](#docker).
 | `endpoints[].ui`                                | UI configuration at the endpoint level.                                                                                                     | `{}`                       |
 | `endpoints[].ui.hide-hostname`                  | Whether to include the hostname in the result.                                                                                              | `false`                    |
 | `endpoints[].ui.dont-resolve-failed-conditions` | Whether to resolve failed conditions for the UI.                                                                                            | `false`                    |
+| `endpoints[].ui.badge.reponse-time.thresholds`  | List of response time thresholds. Each time a threshold is reached, the badge has a different color.                                        | `[50, 200, 300, 500, 750]` |
```
TwiN, line 177:
```suggestion
| `endpoints[].ui.badge.response-time.thresholds` | List of response time thresholds. Each time a threshold is reached, the badge has a different color.                                        | `[50, 200, 300, 500, 750]` |
```

controller/handler/badge.go
```diff
@@ -209,17 +216,17 @@ func generateResponseTimeBadgeSVG(duration string, averageResponseTime int) []by
...
+
+	group, name, _ := store.Get().GetGroupAndNameByKey(key)
+	ep := config.GetEndpointByGroupAndName(group, name)
+
+	// the threshold config requires 5 values, so we can be sure it's set here
+	for i := 0; i < 5; i++ {
+		if responseTime <= ep.UIConfig.Badge.ResponseTime.Thresholds[i] {
+			return badgeColors[i]
+		}
 	}
```
TwiN, line 229: I don't know how I feel about the fact that to retrieve the badge configuration, we need to make a call to the database.
Instead, `config` should have a function `GetEndpointByKey`. 

The key is generated using `ConvertGroupAndEndpointNameToKey(group, name)`, so it should be possible to just iterate through the endpoints and compare the given key with the each endpoints' key.

controller/handler/handler.go
```diff
@@ -3,50 +3,54 @@ package handler
...
-	if enabledMetrics {
-		router.Handle("/metrics", promhttp.Handler()).Methods("GET")
-	}
 	api := router.PathPrefix("/api").Subrouter()
 	protected := api.PathPrefix("/").Subrouter()
 	unprotected := api.PathPrefix("/").Subrouter()
-	if securityConfig != nil {
-		if err := securityConfig.RegisterHandlers(router); err != nil {
-			panic(err)
+	if config != nil {
```
TwiN, line 17: The config is validated before `CreateRouter` is called, so it shouldn't be necessary to check if the config is nil.

core/ui/ui.go
```diff
@@ -1,17 +1,51 @@
...
+	Badge                       *Badge `yaml:"badge"`
+}
+
+type Badge struct {
+	ResponseTime *Thresholds `yaml:"response-time"`
+}
+
+type Thresholds struct {
+	Thresholds []int `yaml:"thresholds"`
+}
```
TwiN, line 20: Shouldn't this struct be called `ResponseTime` instead?

core/ui/ui.go
```diff
@@ -1,17 +1,51 @@
...
+
+var (
+	ErrInvalidUiBadgeTimeConfig = errors.New("invalid endpoint ui configuration - invalid badgetime configuration: You need to set all responsetime settings and they have to be in an ascending value range")
```
TwiN, line 23: `ErrInvalidUiBadgeTimeConfig` should be `ErrInvalidBadgeResponseTimeConfig`, and I'd suggest being a bit more explicit in the error string:
```
invalid response time badge configuration: expected parameter 'thresholds' to have 5 ascending numerical values
```

core/ui/ui.go
```diff
@@ -1,17 +1,51 @@
...
+		return ErrInvalidUiBadgeTimeConfig
+	}
+	for i := 4; i > 0; i-- {
+		if config.Badge.ResponseTime.Thresholds[i] < config.Badge.ResponseTime.Thresholds[i-1] {
+			return ErrInvalidUiBadgeTimeConfig
+		}
+	}
+	return nil
+
 }
```
TwiN, line 38: We should add a nil check for `config.Badge` and `config.Badge.ResponseTime`; otherwise, this configuration would cause a panic:
```
// ...
  ui:
    badge:
      response-time: {}
```

storage/store/store.go
```diff
@@ -24,6 +24,8 @@ type Store interface {
 	// GetEndpointStatusByKey returns the endpoint status for a given key
 	GetEndpointStatusByKey(key string, params *paging.EndpointStatusParams) (*core.EndpointStatus, error)
 
+	GetGroupAndNameByKey(key string) (string, string, error)
```
TwiN, line 27: May not be necessary based on comment mentioning that making a database call to get the key is unnecessary.

TwiN (changes requested):

storage/store/sql/sql.go
TwiN: Actually, this may not be needed after all. See https://github.com/TwiN/gatus/pull/267#discussion_r847819551.

core/ui/ui.go
Fry2412: Since there is still an uptime badge that can be configured with thresholds in the same way, I thought that this way would already suffice and therefore prepared it. 
But you could also take out the threshold type directly and replace it with an int[] and thus reduce the config levels.

core/ui/ui.go
Fry2412: At this point I would also write a default value into the config if the nil check failed and rename the method `ValidateAndSetDefaults`.

README.md
```diff
@@ -174,6 +174,7 @@ If you want to test it locally, see [Docker](#docker).
 | `endpoints[].ui`                                | UI configuration at the endpoint level.                                                                                                     | `{}`                       |
 | `endpoints[].ui.hide-hostname`                  | Whether to include the hostname in the result.                                                                                              | `false`                    |
 | `endpoints[].ui.dont-resolve-failed-conditions` | Whether to resolve failed conditions for the UI.                                                                                            | `false`                    |
+| `endpoints[].ui.badge.reponse-time`  | List of response time thresholds. Each time a threshold is reached, the badge has a different color.                                        | `[50, 200, 300, 500, 750]` |
```
TwiN, line 177: Could you fix the alignment? 🙏

config/config.go
```diff
@@ -89,6 +90,17 @@ type Config struct {
 	lastFileModTime time.Time // last modification time
 }
 
+func (config *Config) GetEndpointByKey(key string) *core.Endpoint {
+
```
TwiN, line 94:
```suggestion
```

TwiN (changes requested):

controller/handler/endpoint_status_test.go
```diff
@@ -153,6 +153,7 @@ func TestEndpointStatus(t *testing.T) {
 }
 
 func TestEndpointStatuses(t *testing.T) {
+
```
TwiN, line 156:
```suggestion
```

core/ui/ui.go
```diff
@@ -1,17 +1,51 @@
...
+type Badge struct {
+	ResponseTime []int `yaml:"response-time"`
+}
+
+var (
+	ErrInvalidBadgeResponseTimeConfig = errors.New("invalid response time badge configuration: expected parameter 'response-time' to have 5 ascending numerical values")
+)
+
+func (config *Config) ValidateAndSetDefaults() error {
+
```
TwiN, line 23:
```suggestion
```

TwiN (changes requested):

core/ui/ui.go
```diff
@@ -1,17 +1,51 @@
...
+	if config.Badge != nil {
+		if config.Badge.ResponseTime != nil {
```
TwiN, line 25: This nil check (`config.Badge.ResponseTime != nil`) is not necessary. The if statement `len(config.Badge.ResponseTime) != 5` will take care of this.

TwiN (changes requested):

Fry2412:
🚀

README.md
```diff
@@ -187,9 +188,6 @@ If you want to test it locally, see [Docker](#docker).
...
 | `ui.logo`                                       | URL to the logo to display.                                                                                                                        | `""`                       |
 | `ui.link`                                       | Link to open when the logo is clicked.                                                                                                             | `""`                       |
-| `ui.buttons`                                    | List of buttons to display below the header.                                                                                                       | `[]`                       |
-| `ui.buttons[].name`                             | Text to display on the button.                                                                                                                     | Required `""`              |
-| `ui.buttons[].link`                             | Link to open when the button is clicked.                                                                                                           | Required `""`              |
```
TwiN, line 192: Why is this being removed? the root-level `ui` configuration is different from the endpoint-level `ui` configuration.

Also, I'm sorry for making you wait for so long - thank you for your patience :)

TwiN (changes requested):

README.md
Fry2412: idk, wasn't aware of this removal at all. We rushed it via the online diff editor. Maybe this is sourced in another branch/merge and my branch isn't on yet. This feature is completly new to me.
Will readd it asap

TwiN:
@Fry2412 Looks good! Could you add an example in the README? Just adding a section in the FAQ named `How to change the color thresholds of the response time badge` with an example of the minimal configuration will suffice.

Fry2412:
Two things
Firstly, I overlooked the fact that it was supposed to go in the FAQ section, so I moved it to where it made sense to me.

Second, since I only changed the README file, I don't know why there are merge conflicts again.

TwiN:
Could you resolve the conflicts? I just now noticed that you resolved my last concern through 384ca98 😅 sorry for the delay

Jesibu:
Hi TwiN,
Fry is not able to work on that pull request. He gave it to me and I updated the branch to the latest gatus version.

config/config.go
```diff
@@ -6,6 +6,7 @@ import (
 	"os"
 	"time"
 
+
```
TwiN, line 9:
```suggestion
```

controller/controller.go
```diff
@@ -8,10 +8,10 @@ import (
 	"os"
 	"time"
 
+
```
TwiN, line 11:
```suggestion
```

controller/handler/badge.go
```diff
@@ -21,13 +22,18 @@ const (
 	badgeColorHexBad      = "#cc8111"
 	badgeColorHexVeryBad  = "#c7130a"
 )
-
```
TwiN, line 24: Can you re-add the newline here?

controller/controller.go
```diff
@@ -21,19 +21,19 @@ var (
 )
 
 // Handle creates the router and starts the server
-func Handle(securityConfig *security.Config, webConfig *web.Config, uiConfig *ui.Config, enableMetrics bool) {
-	var router http.Handler = handler.CreateRouter(ui.StaticFolder, securityConfig, uiConfig, enableMetrics)
+func Handle(config *config.Config) {
```
TwiN, line 24: Could you pass the parameter `config` as `cfg` so that it doesn't collide with the `config` package?

controller/handler/badge.go
```diff
@@ -21,13 +22,18 @@ const (
...
 	HealthStatusDown    = "down"
 	HealthStatusUnknown = "?"
 )
 
+var (
+	badgeColors = []string{badgeColorHexAwesome, badgeColorHexGreat, badgeColorHexGood, badgeColorHexPassable, badgeColorHexBad}
+)
+
+
+
```
TwiN, line 36:
```suggestion

```

controller/handler/handler.go
```diff
@@ -3,43 +3,49 @@ package handler
...
-	"github.com/TwiN/gatus/v4/config/ui"
-	"github.com/TwiN/gatus/v4/security"
+    "github.com/TwiN/gatus/v4/config"
 	"github.com/TwiN/health"
 	"github.com/gorilla/mux"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
-func CreateRouter(staticFolder string, securityConfig *security.Config, uiConfig *ui.Config, enabledMetrics bool) *mux.Router {
+func CreateRouter(staticFolder string, config *config.Config) *mux.Router {
```
TwiN, line 12: Could you rename the parameter `config` to `cfg` so it doesn't collide with the `config` package?

controller/handler/handler.go
```diff
@@ -3,43 +3,49 @@ package handler
...
+			if err := config.Security.RegisterHandlers(router); err != nil {
+				panic(err)
+			}
+			if err := config.Security.ApplySecurityMiddleware(protected); err != nil {
+				panic(err)
+			}
 		}
+
+		// Endpoints
+
```
TwiN, line 31:
```suggestion
```

controller/handler/handler.go
```diff
@@ -3,43 +3,49 @@ package handler
...
+		unprotected.Handle("/v1/config", ConfigHandler{securityConfig: config.Security}).Methods("GET")
+		unprotected.HandleFunc("/v1/endpoints/{key}/response-times/{duration}/badge.svg", ResponseTimeBadge(config)).Methods("GET")
+		unprotected.HandleFunc("/v1/services/{key}/response-times/{duration}/badge.svg", ResponseTimeBadge(config)).Methods("GET")
+		// SPA
+		router.HandleFunc("/services/{name}", SinglePageApplication(staticFolder, config.UI)).Methods("GET") // XXX: Remove this in v4.0.0
```
TwiN, line 36: There's some strange conflict going on here.. This line should no longer exist.
See https://github.com/TwiN/gatus/blob/7daf2b5cac28201c33fc924a7533609f0eb6ee8e/controller/handler/handler.go#L41

core/ui/ui.go
```diff
@@ -1,13 +1,41 @@
...
+		for i := 4; i > 0; i-- {
+			if config.Badge.ResponseTime[i] < config.Badge.ResponseTime[i-1] {
+				return ErrInvalidBadgeResponseTimeConfig
+			}
+		}
+		config.Badge.ResponseTime = GetDefaultConfig().Badge.ResponseTime
+	}
+
+	config.Badge = GetDefaultConfig().Badge
+	return nil
```
TwiN, line 38: I don't think this is right.

So we're checking if the badge is set, if it's set, then we validate it. That's good so far, but at line 34, we overwrite it with the default configuration's `Badge.ResponseTime`..?

And _after_ that, we overwrite it a second time at line 37 with the default config's `Badge`?

Maybe it's just because I'm mentally exhausted right now, but I don't think it makes sense 🤔

TwiN (changes requested):
There's a couple of problems, and I think there's also a conflict with the history, because some of the code that was removed in the past is back.

TwiN:
I just merged a change in #307 that should make your life easier, if you manage to resolve the conflict. Basically, I took care of passing just the main config instead of multiple sub configs.

controller/handler/handler.go
TwiN: Resolved by #307

controller/controller.go
TwiN: Resolved by #307

controller/handler/handler.go
TwiN: Resolved by #307

core/ui/ui.go
Jesibu: I'm not into all details of the project right know. I will review this code.

TwiN:
I'm confused as to why GitHub thinks there's a conflict.
image

I'll try to manually compare each file to make sure there's nothing weird.

controller/handler/handler.go
```diff
@@ -11,34 +12,48 @@ import (
...
 	api := router.PathPrefix("/api").Subrouter()
 	protected := api.PathPrefix("/").Subrouter()
 	unprotected := api.PathPrefix("/").Subrouter()
-	if cfg.Security != nil {
-		if err := cfg.Security.RegisterHandlers(router); err != nil {
-			panic(err)
+	if cfg != nil {
+		if cfg.Metrics {
+			router.Handle("/metrics", promhttp.Handler()).Methods("GET")
 		}
```
TwiN, line 21: This shouldn't be here

controller/handler/handler.go
```diff
@@ -11,34 +12,48 @@ import (
...
+
+		// Endpoints
+		unprotected.Handle("/v1/config", ConfigHandler{securityConfig: cfg.Security}).Methods("GET")
+		unprotected.HandleFunc("/v1/endpoints/{key}/response-times/{duration}/badge.svg", ResponseTimeBadge(cfg)).Methods("GET")
+		unprotected.HandleFunc("/v1/services/{key}/response-times/{duration}/badge.svg", ResponseTimeBadge(cfg)).Methods("GET")
+		// SPA
+		router.HandleFunc("/services/{name}", SinglePageApplication(staticFolder, cfg.UI)).Methods("GET") // XXX: Remove this in v4.0.0
+		router.HandleFunc("/endpoints/{name}", SinglePageApplication(staticFolder, cfg.UI)).Methods("GET")
+		router.HandleFunc("/", SinglePageApplication(staticFolder, cfg.UI)).Methods("GET")
 	}
```
TwiN, line 39: this shouldn't be here either

TwiN (changes requested):
Okay, most of these reviews are about code that shouldn't exist. 

This, combined with the fact that GitHub is preventing me from rebasing the PR due to conflicts makes me believe that the commit history is borked, and I want to avoid causing issues with the commit history.

As such, I'm going to close this PR, and if you want the feature, you'll have to start a new branch/fork and copy the relevant changes you've made one by one.

On the bright side, there aren't many changes necessary if you remove the changes that should not be in this PR, so it should not require too much work.

I am very sorry that things turned out this way, and I do hope you'll create another PR for these changes.

Jesibu:
Hi Twin, I guess that is a good idea. I will fork the project again on my account and create a new pill request.

Issue #274, opened 2022-04-13
feat(metrics): Add more metrics (e.g. duration)


wei840222:
### Describe the feature request

In this project https://github.com/prometheus/blackbox_exporter, many metrics like probe_success, probe_duration etc... provided. And in dns probe https://github.com/prometheus/blackbox_exporter/blob/master/prober/dns.go#L129 also has metrics for dns.

### Why do you personally want this feature to be implemented?

Maybe we can add these metrics in gatus, and user can scrpe to promethus and make dashboard on Grafana.

### How long have you been using this project?

3 month

### Additional information

If the team agrees to add this feature, I can submit PR to contribute gatus

TwiN:
Sounds reasonable to me, feel free to give it a try!

One thing to keep in mind, however, is that I try to keep Gatus as lightweight as possible, and this extends to metrics too.
See https://status.twin.sh/metrics for an example of what metrics are currently exposed (I believe there's only `gatus_results_total` right now as far as custom metrics goes).

wei840222:
OK, I'll start developing

lucagouty:
@wei840222 thanks for having a look at that one, I'm very interested!

Duration is clearly a must have, but I think we should also add few more : 
- `url`, for now there's only name/key/group
- `cert_valid` / `cert_days_remaining`

I agree with @TwiN, it's great to keep it lightweight, but I'd say those are the bare minimum.

Some inspiration from a project quite similar : https://github.com/louislam/uptime-kuma/blob/master/server/prometheus.js

TwiN:
@lucagouty `cert_valid` and `cert_days_remaining` don't really sound that important to me. At most, you'd just have a graph that gradually goes down. 
Instead, you could just have an endpoint with a condition that uses the `[CERTIFICATE_EXPIRATION]` placeholder, and use the existing `gatus_results_total` metric to figure out when a certificate is about to expire or not.

I do agree on the `url` being a useful label that could be added.

lucagouty:
Hi @TwiN, thank you for taking the time to answer.

> At most, you'd just have a graph that gradually goes down.

Yes I agree, but it's actually quite nice to have it. I know Gatus alerts are very nice, but tbh I intend not to use it and do as much as possible using prometheus/alertmanager/grafana so I keep everything in the same place.

Here is the Grafana dashboard I've made for Uptime Kuma, I'd really love to achieve the same with Gatus :
[...]

PR #278, opened 2022-04-28
feat(metrics): Add more metrics

wei840222:
For issue [#274](https://github.com/TwiN/gatus/issues/274)


metric/metric.go
```diff
@@ -2,26 +2,161 @@ package metric
...
+
+	resultDNSReturnCode *prometheus.CounterVec
+
+	resultTCPConnected *prometheus.CounterVec
+
+	resultStartTLSConnected                          *prometheus.CounterVec
+	resultStartTLSSSLLastChainExpiryTimestampSeconds *prometheus.GaugeVec
+
+	resultTLSConnected                          *prometheus.CounterVec
+	resultTLSSSLLastChainExpiryTimestampSeconds *prometheus.GaugeVec
```
TwiN, line 29: Is there a reason why two of the variables mention only TLS, while the other two mention TLS and SSL?

metric/metric.go
```diff
@@ -2,26 +2,161 @@ package metric
...
+		if result.Connected {
+			resultHTTPConnected.WithLabelValues(endpoint.Key(), endpoint.Group, endpoint.Name, endpoint.URL).Inc()
+		}
+		if result.HTTPStatus != 0 {
+			resultHTTPStatusCode.WithLabelValues(endpoint.Key(), endpoint.Group, endpoint.Name, endpoint.URL, strconv.Itoa(result.HTTPStatus)).Inc()
+		}
+		if result.CertificateExpiration != 0 {
+			resultHTTPSSLLastChainExpiryTimestampSeconds.WithLabelValues(endpoint.Key(), endpoint.Group, endpoint.Name, endpoint.URL).Set(float64(result.CertificateExpiration / time.Second))
+		}
 	}
```
TwiN, line 161: A lot of these look very similar to the code that can be found at https://github.com/TwiN/gatus/blob/a9fc876173a9cb50507d12c1a45a610af99444ae/core/endpoint.go#L232

If possible, I'd like to avoid duplication. Perhaps you could add a method such as `Type` in the `Endpoint` struct and have that logic only there instead of at two different places?

metric/metric.go
```diff
@@ -2,26 +2,161 @@ package metric
...
+	resultTLSSSLLastChainExpiryTimestampSeconds *prometheus.GaugeVec
+
+	resultHTTPConnected                          *prometheus.CounterVec
+	resultHTTPStatusCode                         *prometheus.CounterVec
+	resultHTTPSSLLastChainExpiryTimestampSeconds *prometheus.GaugeVec
 )
 
+func ensurePrometheusMetrics() {
+	if namespace == "" {
+		namespace = "gatus"
```
TwiN, line 38: This is unnecessarily confusing. I'd rather have `namespace` as a const and have a bool like `initializedMetrics`

metric/metric.go
```diff
@@ -2,26 +2,161 @@ package metric
...
+			Namespace: namespace,
+			Name:      "results_http_status_code",
+			Help:      "Response HTTP status code",
+		}, []string{"key", "group", "name", "url", "code"})
+
+		resultHTTPSSLLastChainExpiryTimestampSeconds = promauto.NewGaugeVec(prometheus.GaugeOpts{
+			Namespace: namespace,
+			Name:      "results_http_ssl_last_chain_expiry_timestamp_seconds",
+			Help:      "Returns last SSL chain expiry in timestamp seconds",
+		}, []string{"key", "group", "name", "url"})
```
TwiN, line 110: I feel like a lot of these metrics are unnecessary.

For instance, why do we need `results_success` if `results_total` already has a label `success`?

`results_tcp_connected`, `results_starttls_connected`, `results_tls_connected`, `results_http_connected` and `results_http_status_code` should all be suffixed with `_total` for consistency.

But before even looking too deeply into that, does it really make sense to have all of these individual counters?

You can technically determine the type by simply looking at the url, granted your query might look weird, but even without resorting to this hack, we could just add the label `type` to `results_total` where the type is `http`, `starttls`, `dns`, `tcp` or `icmp`, and we would no longer need most of the metrics you added.

TwiN (changes requested):
Couple comments, but to summarize, I think a lot of the counters you added can be merged into the existing `results_total` metric if you added an extra `type` label.

While it may seem inconsequential to some, the reality is that some paid solutions for storing metrics have pricing plans based on the number of metrics, and adding multiple metrics instead of just increasing the cardinality of an existing metric a very tiny little bit is much more cost-efficient.

wei840222:
Hi @TwiN
Thx for your comments, I've apply some change for your advices.

metric/metric.go
wei840222: I follow this naming convention ```result{Type}SSSLLastChainExpiryTimestampSeconds``` for naming.
But after add a type label, I think the naming can be changed to resultSSLLastChainExpiryTimestampSeconds.

metric/metric.go
wei840222: add a method in core.Endpoint.Type() for get endpoint type.

metric/metric.go
wei840222: added initializedMetrics var for this usage

metric/metric.go
wei840222: I think you're right. After added the type label in each metrics. Many metrics can be merged.

core/endpoint.go
```diff
@@ -105,6 +115,24 @@ func (endpoint Endpoint) IsEnabled() bool {
...
+	case strings.HasPrefix(endpoint.URL, "icmp://"):
+		return EndpointTypeICMP
+	case strings.HasPrefix(endpoint.URL, "starttls://"):
+		return EndpointTypeSTARTTLS
+	case strings.HasPrefix(endpoint.URL, "tls://"):
+		return EndpointTypeTLS
+	default:
+		return EndpointTypeHTTP
+	}
+}
```
TwiN, line 134: Awesome work 👍

metric/metric.go
```diff
@@ -9,19 +9,85 @@ import (
...
+	resultDurationSeconds                    *prometheus.GaugeVec
+	resultConnectedTotal                     *prometheus.CounterVec
+	resultDNSReturnCodeTotal                 *prometheus.CounterVec
+	resultHTTPStatusCodeTotal                *prometheus.CounterVec
+	resultSSLLastChainExpiryTimestampSeconds *prometheus.GaugeVec
 )
 
+func ensurePrometheusMetrics() {
+	if !initializedMetrics {
+		initializedMetrics = true
```
TwiN, line 30: Could you rename `ensurePrometheusMetrics` to `initializePrometheusMetrics` and move the 
`if !initializedMetrics {` statement in the `PublishMetricsForEndpoint` function instead?

e.g.:
```go
func PublishMetricsForEndpoint(endpoint *core.Endpoint, result *core.Result) {
    if !initializedMetrics {
        initializePrometheusMetrics()
        initializedMetrics = true
    }
    // ...
```

metric/metric.go
```diff
@@ -9,19 +9,85 @@ import (
 )
 
 var (
+	// Check if the metric is initialized
+	initializedMetrics bool
+
+	// The prefix of the metrics
+	namespace string
```
TwiN, line 16: Can you make this a const instead?

TwiN (changes requested):

core/endpoint.go
wei840222: thx

wei840222:
Hi @TwiN
The request changes in comment have applied.
Plz check thx a lot :)

metric/metric.go
```diff
@@ -8,20 +8,66 @@ import (
...
+	resultDurationSeconds = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: namespace,
+		Name:      "results_duration_seconds",
+		Help:      "Duration of the request in seconds",
+	}, []string{"key", "group", "name", "type"})
+	resultConnectedTotal = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: namespace,
+		Name:      "results_connected_total",
+		Help:      "Total number of results in which a connection was successfully established",
+	}, []string{"key", "group", "name", "type"})
```
TwiN, line 38: I'm wondering how useful this is compared to just using `results_total`'s `success` field?

Let's say that you had this configuration:
```yaml
endpoints:
  - name: unreachable-ip
    url: "icmp://123.23.45.75"
    interval: 5m
    conditions:
      - "[CONNECTED] == false"
```
`success` would be `true` because the condition `[CONNECTED] == false` is true.

You could make a case that for `HTTP` endpoints, you'd have `success` as `false` even if you had the same condition:
```yaml
endpoints:
  - name: secret-host-that-should-not-be-reachable
    url: "http://top-secret-host-that-is-not-reachable/health"
    interval: 5m
    conditions:
      - "[CONNECTED] == false"
```
the reason being that there would be an error like `no such host`, which would make the result evaluate as a failure despite passing all conditions.

---

TL;DR: When would you use `results_connected_total` instead of `results_total`?

metric/metric.go
```diff
@@ -8,20 +8,66 @@ import (
...
+	resultCodeTotal = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: namespace,
+		Name:      "results_code_total",
+		Help:      "Total number of results by code",
+	}, []string{"key", "group", "name", "type", "code"})
```
TwiN, line 43: FYI I merged `results_dns_return_code_total` and `results_http_status_code_total` into `results_code_total` 3193b2c because they're effectively the same, and you can separate them by looking at their `type`.

metric/metric.go
wei840222: for this case
```
- name: website
  url: "https://twin.sh/healthssss"
  interval: 5s
  conditions:
  - "[CONNECTED] == true"
  - "[STATUS] == 200"
```

assume I have a http check, but the target path return 404 code not 200.
And I can check ```results_connected_total``` increase for connection is still working, but ```results_total``` is not success for code != 200.

metric/metric.go
wei840222: this change is good for reduce metrics nums.
good job :)

wei840222:
hi @TwiN
I added the ut for core.Endpoint.Type()

core/endpoint_test.go
```diff
@@ -23,6 +23,69 @@ func TestEndpoint_IsEnabled(t *testing.T) {
...
+	type fields struct {
+		URL string
+		DNS *DNS
+	}
+	tests := []struct {
+		name   string
+		fields fields
+		want   EndpointType
+	}{{
+		name: "",
```
TwiN, line 36: Any chance you can give these tests a proper name? (i.e. `dns`, `tcp`, `icmp`, etc.)

Actually, you could probably just remove the name altogether, and use `string(tt.want)`

TwiN (changes requested):

core/endpoint_test.go
wei840222: hi @TwiN 
change applied

TwiN (approved):
Looks good!

TwiN:
@wei840222 Awesome work, thank you for the contribution!

Issue #273, opened 2022-03-28
Using a DNS server other than the system one


danilvoe:
### Describe the feature request

The ability to use a dns server other than the system server for monitoring.
Implementation example  https://koraygocmen.medium.com/custom-dns-resolver-for-the-default-http-client-in-go-a1420db38a5d

### Why do you personally want this feature to be implemented?

At the moment, Gatus is running in the cloud (docker) inside which the container itself is with Gatus. This design makes it difficult to configure DNS at the OS level.

mapl:
> The ability to use a dns server other than the system server for monitoring.
You actually can specify the dns server to use
see https://github.com/TwiN/gatus#monitoring-an-endpoint-using-dns-queries

TwiN:
I believe @danilvoe is talking about the DNS server to use for non-DNS endpoints (e.g. HTTP) to, for instance, perform the hostname lookup.

danilvoe:
Works great! (just now had a chance to check in action)

PR #284, opened 2022-05-12
feat(client): Added client configuration option for using a custom DNS resolver

tiwood:
Added a new configuration option to the `client` to allow the user to specify a custom DNS resolver.

Example config:

```yaml
# use Cloudflare DNS to lookup the DNS name for the endpoint
endpoints:
  - name: front-end
    group: core
    url: "https://twin.sh/health"
    interval: 30s
    client:
      dns-resolver: "tcp://1.1.1.1:53"
    conditions:
      - "[STATUS] == 200"
      - "[BODY].status == UP"
      - "[RESPONSE_TIME] < 150"
```

## Checklist

### DIFF
--- a/client/config.go
+++ b/client/config.go
@@ -19,7 +19,7 @@ const (
 )
 
 var (
-	ErrInvalidDNSResolver = errors.New("invalid DNS resolver specified. Required format is {proto}://{ip}:{port}")
+	ErrInvalidDNSResolver        = errors.New("invalid DNS resolver specified. Required format is {proto}://{ip}:{port}")
 	ErrInvalidClientOAuth2Config = errors.New("invalid OAuth2 configuration, all fields are required")
 
 	defaultConfig = Config{
@@ -46,7 +46,7 @@ type Config struct {
 	// Timeout for the client
 	Timeout time.Duration `yaml:"timeout"`
 
-	// DNS Resolver override for the HTTPClient
+	// DNSResolver override for the HTTPClient
 	// Expected format is {protocol}://{host}:{port}
 	DNSResolver string `yaml:"dns-resolver,omitempty"`
 
--- a/client/config.go
+++ b/client/config.go
@@ -4,7 +4,10 @@ import (
 	"context"
 	"crypto/tls"
 	"errors"
+	"fmt"
+	"net"
 	"net/http"
+	"regexp"
 	"time"
 
 	"golang.org/x/oauth2"
@@ -16,6 +19,7 @@ const (
 )
 
 var (
+	ErrInvalidDNSResolver = errors.New("invalid DNS resolver specified. Required format is {proto}://{ip}:{port}")
 	ErrInvalidClientOAuth2Config = errors.New("invalid OAuth2 configuration, all fields are required")
 
 	defaultConfig = Config{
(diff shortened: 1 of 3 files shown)


tiwood:
Closes #273

tiwood:
Hello @TwiN 👋🏻️

We have some critical services we would like to monitor using a different DNS resolver, do you have any indication when/if you merge this?

Thanks for your time!

client/config.go
```diff
@@ -16,6 +19,7 @@ const (
 )
 
 var (
+	ErrInvalidDNSResolver = errors.New("invalid DNS resolver specified. Required format is {proto}://{ip}:{port}")
 	ErrInvalidClientOAuth2Config = errors.New("invalid OAuth2 configuration, all fields are required")
```
TwiN, line 23:
```suggestion
	ErrInvalidDNSResolver        = errors.New("invalid DNS resolver specified. Required format is {proto}://{ip}:{port}")
	ErrInvalidClientOAuth2Config = errors.New("invalid OAuth2 configuration, all fields are required")
```

client/config.go
```diff
@@ -42,6 +46,10 @@ type Config struct {
 	// Timeout for the client
 	Timeout time.Duration `yaml:"timeout"`
 
+	// DNS Resolver override for the HTTPClient
+	// Expected format is {protocol}://{host}:{port}
+	DNSResolver string `yaml:"dns-resolver,omitempty"`
```
TwiN, line 51:
```suggestion
	// DNSResolver override for the HTTPClient
	// Expected format is {protocol}://{host}:{port}
	DNSResolver string `yaml:"dns-resolver,omitempty"`
```

client/config.go
```diff
@@ -64,12 +79,44 @@ func (c *Config) ValidateAndSetDefaults() error {
...
+			r[k] = matches[i]
+		}
+	}
+
+	return DNSResolverConfig{
+		Protocol: r["proto"],
+		Host:     r["host"],
+		Port:     r["port"],
+	}, nil
+}
```
TwiN, line 118: Is there really a need to so thoroughly parse the string?

Also, do we really need to pass the protocol here? I'd almost rather just default to TCP as a protocol all the time.

TwiN (changes requested):

TwiN:
@tiwood Sorry for the delay, by the way!

client/config.go
tiwood: Many (corporate) DNS servers only allow UDP and (some) cloud DNS services only accept TCP, so I think we should be able to choose accordingly.

Would you like it more if we had a nested configuration?

```yaml
dns-resolver:
  protocol: udp
  host: 127.0.0.1
  port: 53
```

tiwood:
> @tiwood Sorry for the delay, by the way!

No worries, I don't expect you to be available all the time

client/config.go
TwiN: That looks a little overkill to me. Let's stay with what you currently made; after all, the whole `protocol://host:port` is used all over Gatus already. Worst case scenario, I'm planning on releasing v4.0.0 soon, so if I want to change something, I'll do it when that happens.

TwiN:
Thank you :) 
Please look at the suggestions I've made, and hopefully, I should be able to merge everything tomorrow 🤞

tiwood:
@TwiN, not sure if you got notified, but I've updated the formatting as requested.

TwiN:
Thank you for contributing, and once again, sorry for the delay.. again😅

PR #285, opened 2022-05-29
docs(cert): list valid units for CERTIFICATE_EXPIRATION

### LINKED ISSUE #246: Checking certificate lifetime in days?
### Describe the bug

Today I received a notification about an expired certificate and complaints about a non-working site, I was very surprised.

I looked at the configuration documentation and noticed that the examples are indicated in hours, and I indicated in days.

Gatus interpreted 15d as 15 minutes?

If Gatus does not support specifying y days, then there should be a startup error, similar to when there is an error in the configuration. In the parameter description, explicitly specify the supported units (hours, days, weeks)

1. Set up verification as "[CERTIFICATE_EXPIRATION] > 14d"
2. Wait for the expiration date of the certificate :)

### Version

3.5.0

### Additional information

_No response_

cgrinds:
Fixes #246

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,15 +220,15 @@ Here are some examples of conditions you can use:
 
 
 #### Placeholders
-| Placeholder                | Description                                              | Example of resolved value                    |
-|:---------------------------|:---------------------------------------------------------|:---------------------------------------------|
-| `[STATUS]`                 | Resolves into the HTTP status of the request             | 404                                          |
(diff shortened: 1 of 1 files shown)


TwiN:
To be honest, I wonder whether it's worth mentioning units smaller than s. I mean, who would realistically have a condition that checks for certification expiration at the millisecond level?

cgrinds:
Yeah, I think that's fair. Will update

TwiN (approved):

PR #291, opened 2022-06-13
feat(badge): Implement UP/DOWN status badge

### LINKED ISSUE #204: feat: Add UP/DOWN badge
Gatus give uptime and response time badges, I would love that gatus give also last status badges (UP or DOWN).
Thanks for your great work!

asymness:
## Summary
Closes #204.

Details page:
image

Status up:
image

Status down:
image

## Checklist


controller/handler/handler.go
```diff
@@ -30,6 +30,7 @@ func CreateRouter(staticFolder string, securityConfig *security.Config, uiConfig
 	unprotected.Handle("/v1/config", ConfigHandler{securityConfig: securityConfig}).Methods("GET")
 	protected.HandleFunc("/v1/endpoints/statuses", EndpointStatuses).Methods("GET") // No GzipHandler for this one, because we cache the content as Gzipped already
 	protected.HandleFunc("/v1/endpoints/{key}/statuses", GzipHandlerFunc(EndpointStatus)).Methods("GET")
+	unprotected.HandleFunc("/v1/endpoints/{key}/status/badge.svg", StatusBadge).Methods("GET")
```
TwiN, line 33: I'm a little unsure about that path.. It's not consistent with the other badges.
Maybe it should be `/v1/endpoints/{key}/current-status/badge.svg` instead? 
I was thinking `/v1/endpoints/{key}/statuses/latest/badge.svg`, but it doesn't really make sense.

At this point I'm just thinking out loud, but I think `/v1/endpoints/{key}/health/badge.svg` makes more sense, but the badge would need to be renamed to the `health` badge, so `StatusBadge` would need to be renamed to `HealthBadge` & same with the other references to it being a `status badge` instead of a `health badge`.

What do you think?

TwiN (changes requested):

controller/handler/handler.go
asymness: I think `current-status` or `status/current` could work for a status badge, but a `HealthBadge` with `/health/badge.svg` path makes more sense. I can make the changes tomorrow.

TwiN: Yeah let's go for health then :) Thanks!

TwiN:
@asymness I'm about to release v4.0.0, do you want me to wait until you've made the changes before releasing it?

asymness:
Hi @TwiN, sorry for the delay. I have made the changes so you can re-review when you're available. Thanks!

TwiN (approved):
Looks great! Thank you for the contribution!

TwiN:
Released in [v4.0.0](https://github.com/TwiN/gatus/releases/tag/v4.0.0)

PR #299, opened 2022-07-16
Add Matrix alert provider

Kalissaac:
## Summary

Adds support to use Matrix as an alert provider. Requires a user account and a room to send messages to, and optionally a custom homeserver.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ Have any feedback or questions? [Create a discussion](https://github.com/TwiN/ga
     - [Configuring Teams alerts](#configuring-teams-alerts)
     - [Configuring Telegram alerts](#configuring-telegram-alerts)
     - [Configuring Twilio alerts](#configuring-twilio-alerts)
+    - [Configuring Matrix alerts](#configuring-matrix-alerts)
     - [Configuring custom alerts](#configuring-custom-alerts)
     - [Setting a default alert](#setting-a-default-alert)
   - [Maintenance](#maintenance)
@@ -276,7 +277,7 @@ See [examples/docker-compose-postgres-storage](.examples/docker-compose-postgres
 
 
 ### Client configuration
-In order to support a wide range of environments, each monitored endpoint has a unique configuration for 
+In order to support a wide range of environments, each monitored endpoint has a unique configuration for
 the client used to send the request.
 
 | Parameter                     | Description                                                                | Default         |
@@ -377,7 +378,7 @@ ignored.
 
 ```yaml
 alerting:
-  discord: 
+  discord:
     webhook-url: "https://discord.com/api/webhooks/**********/**********"
 
 endpoints:
@@ -420,7 +421,7 @@ alerting:
     host: "mail.example.com"
     port: 587
     to: "recipient1@example.com,recipient2@example.com"
-    # You can also add group-specific to keys, which will 
+    # You can also add group-specific to keys, which will
     # override the to key above for the specified groups
     overrides:
       - group: "core"
@@ -470,7 +471,7 @@ endpoints:
 
 ```yaml
 alerting:
-  googlechat: 
+  googlechat:
     webhook-url: "https://chat.googleapis.com/v1/spaces/*******/messages?key=**********&token=********"
 
 endpoints:
@@ -501,7 +502,7 @@ endpoints:
 
 ```yaml
 alerting:
-  mattermost: 
+  mattermost:
     webhook-url: "http://**********/hooks/**********"
     client:
       insecure: true
@@ -601,9 +602,9 @@ Behavior:
 
 ```yaml
 alerting:
-  pagerduty: 
+  pagerduty:
     integration-key: "********************************"
-    # You can also add group-specific integration keys, which will 
+    # You can also add group-specific integration keys, which will
     # override the integration key above for the specified groups
     overrides:
       - group: "core"
@@ -653,7 +654,7 @@ endpoints:
 | `alerting.slack.overrides[].webhook-url`  | Slack Webhook URL                                                                          | `""`          |
 ```yaml
 alerting:
-  slack: 
+  slack:
     webhook-url: "https://hooks.slack.com/services/**********/**********/**********"
 
 endpoints:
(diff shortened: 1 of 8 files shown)


Kalissaac:
Sorry, I realized I have VS Code set to trim trailing spaces and it wrecked havoc on the README. Let me know and I can try to revert the whitespace changes.

alerting/provider/matrix/matrix.go
```diff
@@ -0,0 +1,192 @@
...
+
+	"github.com/TwiN/gatus/v4/alerting/alert"
+	"github.com/TwiN/gatus/v4/client"
+	"github.com/TwiN/gatus/v4/core"
+)
+
+// AlertProvider is the configuration necessary for sending an alert using Matrix
+type AlertProvider struct {
+	// HomeserverURL is the custom homeserver to use (optional)
+	HomeserverURL string `yaml:"homeserver-url"`
```
TwiN, line 20: Would it be possible to rename `HomeserverURL` to `ServerURL`?

alerting/provider/matrix/matrix.go
```diff
@@ -0,0 +1,192 @@
...
+					InternalRoomID: override.InternalRoomID,
+				}
+			}
+		}
+	}
+	return matrixProviderConfig{
+		HomeserverURL:  provider.HomeserverURL,
+		AccessToken:    provider.AccessToken,
+		InternalRoomID: provider.InternalRoomID,
+	}
```
TwiN, line 175: Having the same 3 fields in 3 different structs feels a bit too redundant to me.

Why not just re-use `Override` here, even if you have to add `omitempty` in the `yaml` marshaling instruction for the `Group` field. 

Either that, or what you can do is something like that:
```go
type AlertProvider struct {
	matrixProviderConfig `yaml:",inline"`

	// DefaultAlert is the default alert configuration to use for endpoints with an alert of the appropriate type
	DefaultAlert *alert.Alert `yaml:"default-alert,omitempty"`

	// Overrides is a list of Override that may be prioritized over the default configuration
	Overrides []Override `yaml:"overrides,omitempty"`
}

type Override struct {
	Group string `yaml:"group,omitempty"`

	matrixProviderConfig `yaml:",inline"`
}

const defaultServerURL = "https://matrix-client.matrix.org"

type matrixProviderConfig struct {
	ServerURL      string `yaml:"server-url"`
	AccessToken    string `yaml:"access-token"`
	InternalRoomID string `yaml:"internal-room-id"`
}
```

(you'll have to test it on your own, but the syntax should be close-ish)

alerting/provider/matrix/matrix.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+const letterBytes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
```
TwiN, line 178: Could you love this to the top with the other constant? Also maybe give it a slightly more descriptive name -- what are you trying to achieve with these letter bytes?

Alternatively, maybe you could just move it inside the `randStringBytes` function - I don't think it's an issue.

alerting/provider/provider.go
```diff
@@ -65,4 +66,5 @@ var (
 	_ AlertProvider = (*teams.AlertProvider)(nil)
 	_ AlertProvider = (*telegram.AlertProvider)(nil)
 	_ AlertProvider = (*twilio.AlertProvider)(nil)
+	_ AlertProvider = (*matrix.AlertProvider)(nil)
```
TwiN, line 69: Could you move it up a bit so it's sorted alphabetically?

config/config.go
```diff
@@ -282,6 +282,7 @@ func validateAlertingConfig(alertingConfig *alerting.Config, endpoints []*core.E
 		alert.TypeTeams,
 		alert.TypeTelegram,
 		alert.TypeTwilio,
+		alert.TypeMatrix,
```
TwiN, line 285: Could you move it up a bit so it's sorted alphabetically?

alerting/config.go
```diff
@@ -131,6 +135,12 @@ func (config Config) GetAlertingProviderByAlertType(alertType alert.Type) provid
 			return nil
 		}
 		return config.Twilio
+	case alert.TypeMatrix:
+		if config.Matrix == nil {
+			// Since we're returning an interface, we need to explicitly return nil, even if the provider itself is nil
+			return nil
+		}
+		return config.Matrix
```
TwiN, line 143: Could you move it up a bit so it's sorted alphabetically?

README.md
```diff
@@ -50,6 +50,7 @@ Have any feedback or questions? [Create a discussion](https://github.com/TwiN/ga
     - [Configuring Teams alerts](#configuring-teams-alerts)
     - [Configuring Telegram alerts](#configuring-telegram-alerts)
     - [Configuring Twilio alerts](#configuring-twilio-alerts)
+    - [Configuring Matrix alerts](#configuring-matrix-alerts)
```
TwiN, line 53: Could you move it up a bit so it's sorted alphabetically? (excl. the `custom` alert, which should stay at the bottom)

TwiN (changes requested):
Just a couple of small changes, and we should be good to go!

alerting/provider/matrix/matrix.go
Kalissaac: I renamed it to `availableCharacterBytes`, does that work? I'm not really sure what name to use. Maybe `availableTxnIdCharacters`?

alerting/provider/matrix/matrix.go
Kalissaac: I tried to do that before but I couldn't get the YAML working right, turns out I was missing the marshaling instruction! By the way, I had to export `MatrixProviderConfig` to be able to use it in the watchdog alerting test, if you would prefer that it stays within the package then I'll try to find a different way to approach it.

TwiN (approved):
Excellent work!

alerting/provider/matrix/matrix.go
TwiN: `availableCharacterBytes` is fine, moreover, now that you've move it inside the function, it's a lot easier to confuse, so that looks perfect!

PR #305, opened 2022-07-24
feat(docker): add basic healthcheck support

sparanoid:
## Summary

This can help run Gatus with Docker behind proxy like Nginx:

```yaml
  gatus:
    image: twinproduction/gatus:v4.0.0
    restart: always
    volumes:
      - ./config/gatus:/config
      - ./data/gatus:/data/
    healthcheck:
      test: ["CMD-SHELL", "curl -f http://localhost:8080 || exit 1"]
      interval: 30s
      retries: 5
      timeout: 3s
```

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # Build the go application into a binary
 FROM golang:alpine as builder
-RUN apk --update add ca-certificates
+RUN apk --update --no-cache add ca-certificates curl
 WORKDIR /app
 COPY . ./
 RUN CGO_ENABLED=0 GOOS=linux go build -mod vendor -a -installsuffix cgo -o gatus .
--- a/.examples/docker-compose-postgres-storage/docker-compose.yml
+++ b/.examples/docker-compose-postgres-storage/docker-compose.yml
@@ -12,6 +12,10 @@ services:
       - POSTGRES_PASSWORD=password
     networks:
       - web
+    healthcheck:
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Build the go application into a binary
 FROM golang:alpine as builder
-RUN apk --update add ca-certificates
+RUN apk --update --no-cache add ca-certificates curl
```
TwiN, line 3: Is there really any point in installing curl on the builder image, since the builder's only really used to generate the binary and move it to the next stage (scratch/empty image)?

TwiN (changes requested):

TwiN:
I'm not against updating the docker-compose example with Postgres, but as for adding one for Gatus, I don't think it's really useful.

Gatus ensures that all configuration is validated on start, and technically, nothing else should be able to cause Gatus to exit.
https://github.com/TwiN/gatus/blob/04a682eddc50b9eb5c456ab3eb6563e83c116f4e/main.go#L17-L20

I'd also like to keep the Docker image as small as possible, and that may be hard if we want to have a working shell (since Gatus' image uses an empty/scratch image as base).

sparanoid:
> Gatus ensures that all configuration is validated on start, and technically, nothing else should be able to cause Gatus to exit.

Yep, that make sense. Closing for now.

Issue #317, opened 2022-08-13
Config via a new GATUS_CONFIG environment variable.


tostt:
### Describe the feature request

Currently, users can specify a custom path for the configuration file by setting the `GATUS_CONFIG_FILE` environment variable. That's nice.

Going further, to make life easier for users who have no control on the infrastructure (especially volumes), a new `GATUS_CONFIG` environment variable would be useful.
This variable would contain a valid configuration (i.e. the contents of a configuration file). If set, Gatus would use it directly without reading the config file.

### Why do you personally want this feature to be implemented?

The main reason is to make life easier :
- at configuration time : keeping the configuration with the docker-compose stack,
- at runtime: avoid having to create, keep track and manage a volume for a single file.

TwiN:
Hello @tostt,

The problem with this is that some OS appear to have limitations on maximum number of bytes for environment variables.

That said, I'm not strictly against having such a feature, so if you'd like to implement it, feel free to create a PR.

TwiN:
See https://github.com/TwiN/gatus/pull/318#issuecomment-1396512518

PR #318, opened 2022-08-14
Configuration via an environment variable

tostt:
## Summary

Implementation of #317.
Notes: in the docker-compose file, version is 3.3 and image name refers to latest of twinproduction/gatus (which will need to be changed for testing)

## Checklist

### DIFF
--- a/.examples/docker-compose-in-memory-config/docker-compose.yml
+++ b/.examples/docker-compose-in-memory-config/docker-compose.yml
@@ -1,0 +1,9 @@
+version: "3.3"
+services:
+  gatus:
+    image: mygatus
+    ports:
+      - 8080:8080
+    environment:
+      GATUS_CONFIG: > 
+        {endpoints: [{name: website, url: 'https://twin.sh/health', interval: 5m, conditions: ['[STATUS] == 200', '[BODY].status == UP', '[RESPONSE_TIME] < 300']}, {name: example, url: 'https://example.org/', interval: 60s, conditions: ['[STATUS] == 200']}]}
--- a/.examples/docker-compose-in-memory-config/docker-compose.yml
+++ b/.examples/docker-compose-in-memory-config/docker-compose.yml
@@ -1,7 +1,7 @@
 version: "3.3"
 services:
   gatus:
-    image: gatus:tostt
+    image: twinproduction/gatus
     ports:
       - 8080:8080
     environment:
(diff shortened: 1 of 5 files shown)


.examples/docker-compose-in-memory-config/docker-compose.yml
```diff
@@ -0,0 +1,9 @@
+version: "3.3"
+services:
+  gatus:
+    image: twinproduction/gatus
+    ports:
+      - 8080:8080
+    environment:
+      GATUS_CONFIG: > 
+        {endpoints: [{name: website, url: 'https://twin.sh/health', interval: 5m, conditions: ['[STATUS] == 200', '[BODY].status == UP', '[RESPONSE_TIME] < 300']}, {name: example, url: 'https://example.org/', interval: 60s, conditions: ['[STATUS] == 200']}]}
```
TwiN, line 9: Is there a reason why you decided to go with this approach? 
IIRC, you should be able to have multi-line strings by using `|` instead of `>`:
```yaml
version: "3.3"
services:
  gatus:
    image: twinproduction/gatus
    ports:
      - 8080:8080
    environment: |
      endpoints:
        # ...
```

main.go
```diff
@@ -53,11 +53,16 @@ func save() {
...
+	if len(customConfig) > 0 {
+		cfg, err = config.LoadFromString(customConfig)
 	} else {
-		cfg, err = config.LoadDefaultConfiguration()
+		customConfigFile := os.Getenv("GATUS_CONFIG_FILE")
+		if len(customConfigFile) > 0 {
+			cfg, err = config.Load(customConfigFile)
+		} else {
+			cfg, err = config.LoadDefaultConfiguration()
+		}
```
TwiN, line 65: Can you write some tests for this? 🙏

TwiN (changes requested):

.examples/docker-compose-in-memory-config/docker-compose.yml
tostt: I modified the docker-compose file as per your suggestion.
The example now shows a configuration using standard multi-line YAML.
See commit f7bdf71c05e67c42754a7c2c330154691ba4468d

main.go
tostt: I'm afraid I really wouldn't know how to do that.

TwiN:
I know [I was originally open to the idea](https://github.com/TwiN/gatus/issues/317#issuecomment-1214223607), but I've changed my mind.

I'm really sorry for making you work for nothing, but I don't want to advocate bad practices, and I feel like having an entire configuration in an environment variable is a bad practice. At least, as a platform engineer, if I saw this, I would think so.

Thank you for the contribution regardless!

Issue #253, opened 2022-02-14
feat(web): HTTPS/TLS support


chr1st1ank:
### Describe the feature request

It looks like the Gatus service does not support serving over an encrypted connection by itself. Instead one would need to add a reverse proxy in front to enable HTTPS. It would be great if one could specify the path of TLS certificates and Gatus would then offer the endpoints over an encrypted connection.

### Why do you personally want this feature to be implemented?

HTTPS has become the standard for web services. Especially if any kind of credentials are involved encryption is necessary.  And Gatus supports HTTP basic auth - which is very unsecure in the absence of encryption. Using an external proxy makes the deployment unnecessarily complicated.

### Additional information

Enabling HTTPS might not be complicated. Here is an example: https://gist.github.com/denji/12b3a568f092ab951456

TwiN:
That's a good point. To be honest, I built Gatus with Kubernetes in mind, which implies a reverse proxy/ingress controller (or even a load balancer) that takes care of TLS termination, so I never bothered thinking about it as a fully stand-alone application from end to end.

I don't know when I'll be able to work on it due to how busy I am lately though, so if you or anybody wants to take a shot at it, by all means 🙏

chr1st1ank:
Thanks for the response. No worries, just opened the feature request because I found it logical to have in the backlog. Maybe I try a PR indeed at some point. But for now I added an nginx reverse proxy which handles TLS.

Darkvater:
following, have the same issue here. Cannot connect to any status-page requiring a cert as it says "remote error: tls: bad certificate". As this feature is critical I might have a go at getting it done myself, will share a pull-request if done.

iamdempa:
Is this released?

TwiN:
@iamdempa No, otherwise the issue would have been closed :P

ndlanier:
I'll add my +1 to wanting this feature.

Great tool though @TwiN, the simplicity and scalability with this is amazing and can't wait to see how this project grows.

razermuse:
Personally, I think there are likely more important features to implement than SSL support in the short term. I use NGINX with Let's Encrypt and it works great.

chr1st1ank:
Well, it is easier to implement than to configure nginx. Also, if you work with containers, you need to have two and to configure the container networking properly. It is possible, I'm also using it like this, with docker-compose, but this is such an easy catch that I don't see the point.

The implementation is also already done in #322. It just went out of sync while waiting for a final review 🤷‍♂️

PR #322, opened 2022-08-27
feat(web): Support TLS encryption

chr1st1ank:
## Summary
Implementing #253 (HTTPS/TLS support).

This is to run the server on HTTPS instead of plain HTTP.

Two considerations are worth mentioning:
1. The options are just the minimum to have TLS. There are many more options and one could add a full section for configuration of TLS. (See e.g. here: https://github.com/PowerDNS/go-tlsconfig/blob/master/config.go). But for my purposes the simple setup is more than enough. If people want more options, they can also use a reverse proxy or load balancer.
2. At first I wanted to add tests. But then I saw that the serving part of the controller is already untested now. I would have introduced a lot of additional complexity into the testing suite (create certificates, actually run the server, run queries against it) with possibly additional dependencies for creating the certificate. If we instead add hardcoded test certificates, most security scanners would remark that there are credentials in the repo.  And all of this to test just a few calls into the standard library. So I'd like to ask first whether such a test should be added or rather be left out. 

### How to test it:
Create a certificate pair:
```
openssl ecparam -genkey -name secp384r1 -out server.key
openssl req -new -x509 -sha256 -key server.key -out server.crt -days 3650
```

Configure gatus to use the keypair in the config.yaml:
```yaml
web:
  port: 4443
  cert_file: "server.crt"
  key_file: "server.key"
endpoints:
  - name: example
    url: https://example.org
    interval: 30s
    conditions:
      - "[STATUS] == 200"
```

Run gatus.

## Checklist
- [X] Tested and/or added tests to validate that the changes work as intended, if applicable. (I tested it locally)


chr1st1ank:
@TwiN Any chance of including this? I tried to make it as consistent as possible with the existing code

README.md
```diff
@@ -1028,13 +1030,15 @@ maintenance:
...
-| `security.oidc`                  | OpenID Connect configuration | `{}`          |
-
-#### Basic
+| Parameter        | Description                                    | Default |
+|:-----------------|:-----------------------------------------------|:--------|
+| `security`       | Security configuration                         | `{}`    |
+| `security.basic` | HTTP Basic configuration                       | `{}`    |
+| `security.oidc`  | OpenID Connect configuration                   | `{}`    |
+| `web.cert_file`  | Public certificate file for TLS in PEM format. | ``      |
+| `web.key_file`   | Private key file for TLS in PEM format.        | ``      |
```
TwiN, line 1039: These two don't belong there. Only configuration parameters from the `security` configuration should be there.

README.md
```diff
@@ -1052,7 +1056,7 @@ security:
 **WARNING:** Make sure to carefully select to cost of the bcrypt hash. The higher the cost, the longer it takes to compute the hash,
 and basic auth verifies the password against the hash on every request. As of 2022-01-08, I suggest a cost of 8.
 
-#### OIDC (ALPHA)
+#### OIDC Authentication (ALPHA)
```
TwiN, line 1059: Can you update the table of content to reflect this change?

README.md
```diff
@@ -1077,6 +1081,16 @@ security:
...
 
+#### TLS Encryption
+Gatus supports basic encryption with TLS. To enable this, certificate files in PEM format have to be provided.
+The example below shows an example configuration which makes gatus respond on port 4443 to HTTPS requests.
+
+```yaml
+web:
+  port: 4443
+  cert_file: "server.crt"
+  key_file: "server.key"
```
TwiN, line 1092: All configuration parameters with multiple words have said words split by `-` rather than `_`, could you update that here and on L190-L191?

I'd also prefer if it was called `certificate-file` instead of `cert-file`, and `key-file` should instead be `private-key-file`.

---

Finally, I wonder whether if it wouldn't be a good idea to just directly specify the certificate and private key directly in the configuration... Yes, I know it's a little chaotic, but that'd make things a lot easier for people.

Something like:
```yaml
web:
    private-key: |
      -----BEGIN RSA PRIVATE KEY-----
      ...
      -----END RSA PRIVATE KEY-----
    certificate: |
      -----BEGIN CERTIFICATE-----
      ...
      -----END CERTIFICATE-----

```

controller/controller.go
```diff
@@ -25,8 +25,12 @@ func Handle(cfg *config.Config) {
 	if os.Getenv("ENVIRONMENT") == "dev" {
 		router = handler.DevelopmentCORS(router)
 	}
+
+	tls_config, _ := cfg.Web.TLSConfig()
```
TwiN, line 29: Go uses camel case.

config/web/web.go
```diff
@@ -40,10 +47,27 @@ func (web *Config) ValidateAndSetDefaults() error {
...
+func (web *Config) TLSConfig() (*tls.Config, error) {
+	if web.CertFile == "" && web.KeyFile == "" {
+		return nil, nil
+	}
+	cer, err := tls.LoadX509KeyPair(web.CertFile, web.KeyFile)
+	if err != nil {
+		return nil, err
+	}
+	return &tls.Config{Certificates: []tls.Certificate{cer}}, nil
+}
```
TwiN, line 73: I would rather have a function called `loadTLSConfig()` which only gets applied if `len(web.CertFile) > 0 && len(web.KeyFile) > 0`, and it would set a private field named `tlsConfig` in the `Config` struct which we could then publicly retrieve using the `TLSConfig()` method.

This would allow  us to just read the files once rather than the current twice.

TwiN (changes requested):

README.md
chr1st1ank: I really would not mix credentials and configuration

chr1st1ank: The naming is adjusted as suggested. It really looks clearer now.

controller/controller.go
chr1st1ank: Done

config/web/web.go
chr1st1ank: Done as suggested. Only that I also had to save the possible error in a private field. Otherwise it would be necessary to attempt the loading twice on error.

README.md
```diff
@@ -1081,6 +1084,16 @@ security:
...
 
+### TLS Encryption
+Gatus supports basic encryption with TLS. To enable this, certificate files in PEM format have to be provided.
+The example below shows an example configuration which makes gatus respond on port 4443 to HTTPS requests.
+
+```yaml
+web:
+  port: 4443
+  certificate-file: "server.crt"
+  private-key-file: "server.key"
```
TwiN, line 1095: Call me nitpicky, but I think I would prefer if this was under `tls`

Something like
```yaml
web:
  port: 443
  tls:
    certificate-file: "server.crt"
    private-key-file: "server.key"
```

TwiN (changes requested):
Could you add some tests?

chr1st1ank:
> Could you add some tests?

It isn't straight-forward with the certificates. What way would you prefer:
- Add hardcoded test certificates to the repository (which is simple, but linters ususally don't like it)
- Generate a self-signed certificate on the fly with Go (which might need additional dependencies?). Here is an example of such code: https://go.dev/src/crypto/tls/generate_cert.go
- Generate a self-signed certificate locally on the dev environment & on github actions before running the test suite

TwiN:
> Generate a self-signed certificate on the fly with Go (which might need additional dependencies?). Here is an example of such code: https://go.dev/src/crypto/tls/generate_cert.go

I personally prefer this one.

Based on what I can see in the example you linked, the dependency is only used for building anyways:
```go
//go:build ignore
```

README.md
chr1st1ank: Yes, looks better. I've done that.

chr1st1ank:
Alright. Now the code is almost fully tested with certificates which are generated on the fly.
It was actually not as complicated as I expected. I didn't know that go has all the necessary tools already in the standard library.

I'm looking forward to see the CI do its job. On my machine two tests in "client" and "core" are failing. But they don't seem to have any connection with my changes.

config/web/web.go
```diff
@@ -21,6 +22,21 @@ type Config struct {
 
 	// Port to listen on (default to 8080 specified by DefaultPort)
 	Port int `yaml:"port"`
+
+	// TLS configuration
+	Tls TlsConfig `yaml:"tls"`
```
TwiN, line 27: This should be `TLS`

config/web/web.go
```diff
@@ -21,6 +22,21 @@ type Config struct {
...
+
+	tlsConfig      *tls.Config
+	tlsConfigError error
+}
+
+type TlsConfig struct {
```
TwiN, line 33: This should be `TLSConfig`.

config/web/web.go
```diff
@@ -21,6 +22,21 @@ type Config struct {
...
+
+	// Optional public certificate for TLS in PEM format.
+	CertFile string `yaml:"certificate-file,omitempty"`
+
+	// Optional private key file for TLS in PEM format.
+	KeyFile string `yaml:"private-key-file,omitempty"`
```
TwiN, line 39: Can you rename these to `CertificateFile` and `PrivateKeyFile` respectively?

config/web/web.go
```diff
@@ -21,6 +22,21 @@ type Config struct {
 
 	// Port to listen on (default to 8080 specified by DefaultPort)
 	Port int `yaml:"port"`
+
+	// TLS configuration
+	Tls TlsConfig `yaml:"tls"`
+
+	tlsConfig      *tls.Config
+	tlsConfigError error
```
TwiN, line 30: Why are we persisting the error in the struct?

config/web/web.go
```diff
@@ -40,10 +56,31 @@ func (web *Config) ValidateAndSetDefaults() error {
 	} else if web.Port < 0 || web.Port > math.MaxUint16 {
 		return fmt.Errorf("invalid port: value should be between %d and %d", 0, math.MaxUint16)
 	}
+	// Try to load the TLS certificates
+	_, err := web.TLSConfig()
+	if err != nil {
+		return fmt.Errorf("invalid tls config: %s", err)
```
TwiN, line 62:
```suggestion
		return fmt.Errorf("invalid tls config: %w", err)
```

config/web/web.go
```diff
@@ -40,10 +56,31 @@ func (web *Config) ValidateAndSetDefaults() error {
...
+		web.loadTLSConfig()
+	}
+	return web.tlsConfig, web.tlsConfigError
+}
+
+func (web *Config) loadTLSConfig() {
+	cer, err := tls.LoadX509KeyPair(web.Tls.CertFile, web.Tls.KeyFile)
+	if err != nil {
+		web.tlsConfigError = err
+	}
```
TwiN, line 84: I'd rather just return the error here instead of persisting the error in the configuration struct

TwiN (changes requested):

config/web/web.go
chr1st1ank: See below

config/web/web.go
chr1st1ank: Yes, it is a bit awkward. In my first version I directly returned the error. But the issue is that the `TLSConfig()` method is called twice, once in `ValidateAndSetDefaults()` and  once in `controller.go`. So we might try to open the same certificate files twice and create the same error twice. This can only be avoided if we somehow persist the error. Memoization would be a little bit nice because the state would be kept inside the function, but there is no good way to express that in Go, or?

If you see any better way, I'm happy to use it :-)

chr1st1ank:
@TwiN I added tests and updated from master again.

gaby:
@chr1st1ank @TwiN Any updates on this? I would rather have TLS directly with gatus than adding NGINX

gaby:
@chr1st1ank I noticed there's no support for ClientCert auth, which is very useful for enterprise environments.

It basically requires adding a field for a CA, and adding a boolean option to verify that the incoming certs are signed by the same CA as the server. 

This could probably be added later.

chr1st1ank:
I can resolve the conflicts to master and make it mergeable again, but only if there is an interest for this PR. @gaby the client certificates I'd really rather take separately

TwiN:
@chr1st1ank If you can resolve the conflicts, I'll merge it.
Sorry for the trouble 😔

chr1st1ank:
> @chr1st1ank If you can resolve the conflicts, I'll merge it. Sorry for the trouble pensive

Alright, no problem. I've merged master into my branch again and tested it locally. Looks ready to go unless you find any remaining flaws.

controller/controller.go
```diff
@@ -24,8 +24,12 @@ func Handle(cfg *config.Config) {
 	if os.Getenv("ENVIRONMENT") == "dev" {
 		router = handler.DevelopmentCORS(router)
 	}
+
+	tlsConfig, _ := cfg.Web.TLSConfig()
+
```
TwiN, line 29:
```suggestion
	tlsConfig, _ := cfg.Web.TLSConfig()
```
Also, could you check the `err` instead of skipping it?

I would rather fail fast than let the app silently run with a potentially hidden issue.

TwiN (changes requested):

controller/controller.go
chr1st1ank: Ok, I added a panic there, because `Handle()` doesn't return an error value. But this error branch is unreachable as long as there is a separate config validation before: https://github.com/TwiN/gatus/blob/5495ff8088eb05d72184f28b4dd14c94291ba0a9/config/web/web.go#L60
So it fails even earlier anyway. I tested it with an invalid certificate file. This produces:
```
❯ ./gatus
2023/04/22 14:06:00 [config][LoadConfiguration] Reading configuration from configFile=
2023/04/22 14:06:00 [config][validateAlertingConfig] Alerting is not configured
2023/04/22 14:06:00 [config][validateEndpointsConfig] Validated 1 endpoints
panic: invalid tls config: tls: failed to parse private key

goroutine 1 [running]:
main.main()
        ~/code/gatus/main.go:19 +0x168
```

TwiN: You're right, my bad. I'll remove it back and add a comment myself - you've already done more than enough work on this and I've dragged the review on for too long at this point 😂

TwiN (approved):

TwiN:
@chr1st1ank Thank you so much for your contribution (and patience 😂), I appreciate it!

PR #324, opened 2022-09-02
feat(alerting): add client config for telegram

lschloetterer:
## Summary

Adds an option to specify the client config for the telegram alerting provider.
I'm monitoring a local dns server, so the notifications should still go through if the local dns is down.
Builds on top of #284 

I have to admit I have never worked with Go before, so I'm not sure how to compile and test the project, but this change seems relatively straighforward.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -784,8 +784,6 @@ alerting:
   telegram:
     token: "123456:ABC-DEF1234ghIkl-zyx57W2v1u123ew11"
     id: "0123456789"
-    client:
-      dns-resolver: "tcp://1.1.1.1:53"
 
 endpoints:
   - name: website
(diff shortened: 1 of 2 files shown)


lschloetterer:
Well, I see now that I just have to build the docker image...
Just tried it out, works as expected

README.md
```diff
@@ -776,13 +776,16 @@ Here's an example of what the notifications look like:
...
+| `alerting.telegram.client`        | Client configuration. <br />See [Client configuration](#client-configuration).             | `{}`                       |
 | `alerting.telegram.default-alert` | Default alert configuration. <br />See [Setting a default alert](#setting-a-default-alert) | N/A                        |
 
 ```yaml
 alerting:
   telegram:
     token: "123456:ABC-DEF1234ghIkl-zyx57W2v1u123ew11"
     id: "0123456789"
+    client:
+      dns-resolver: "tcp://1.1.1.1:53"
```
TwiN, line 788:
```suggestion
```
Can you remove this?

I try to keep the example configurations as small as possible to prevent the README from getting _too_ big :P

lschloetterer:
I've removed the example as requested

TwiN (approved):
Excellent work @lschloetterer.

Since you mentioned not being familiar with Go, I'll take care of writing some tests for you.

Issue #224, opened 2021-12-14
Who's using Gatus in production?


TwiN:
If you're using Gatus in production, please let me know in the comments.

Regardless of whether you're using it for personal projects or at your company, a quick comment of what features you use would greatly benefit me _(DNS? TCP? HTTP? If you're using HTTP, are you just checking the status, or are you also validating the response body?)_

For instance:
- What company do you work for?
- What do you monitor with Gatus? (business applications? databases? DNS queries? certificate expiration?)
- How long have you been using it for?
- How many endpoints are you monitoring?
- Is it public-facing, or internal only? (e.g. can customers see it, or only developers?)
- What feature would you like to see in Gatus?
- What do you dislike the most about Gatus?

### Template
Though you don't have to use this template, here's a ready-to-use template with the aforementioned questions, if that helps convince you to spend a couple minutes to provide some feedback.
```markdown
**What company do you work for?:** 
**What do you monitor with Gatus?:**
**How long have you been using it for?:**
**How many endpoints are you monitoring?:**
**Is it public-facing, or internal only?:**
**What feature would you like to see in Gatus?:**
**What do you dislike the most about Gatus?:**
```

---

If you'd rather share feedback privately, please feel free to reach out to feedback@gatus.io

zeylos:
I do use gatus in production :upside_down_face: 

**What company do you work for?:** 
Eco CO2 https://www.ecoco2.com/
**What do you monitor with Gatus?:**
Webapps frontpages, api endpoints, internal tools, static websites, certificates expiration (I use wildcards certificates so I configure specific endpoints for each certificate only to check expiration once a day).
I only use HTTP checks, with status and response body for some api endpoints.
**How long have you been using it for?:**
6 months
**How many endpoints are you monitoring?:**
approx. 50
**Is it public-facing, or internal only?:**
Internal only
**Likes/dislikes:**
  - I really like the simplicity of the tool as is, I only use it to get alerts for our websites and apis from outside of the clusters.
  - I currently have to deploy 3 gatus instances because I need to alert on 3 different teams channels, which is not optimal but actually not that problematic cause it makes the configuration smaller for each instance and easier to read.
  - I guess the dashboards could be improved in terms of UI/UX, but I've no skill in this so I can't really help :D. For my use the current UI is totally enough because I mainly use alerts.
[...]

PR #332, opened 2022-09-18
fix(ui): center and fixed navigation arrows

shashank68:
## Summary

- Centered Navigation arrows
- keep them fixed

When the number of endpoints is large, we had to scroll down to navigate through timeline. This patch makes navigation quicker.

## Checklist


TwiN:
Can you show me an image of what this looks like?

TwiN:
To be honest, I'm not exactly sure how I feel about that. 

I do agree that the arrow on the top left corner of the endpoint details page is very poorly placed at the moment (and sometimes even covers the title), but I personally don't find the new static pagination arrows much more aesthetically pleasing than they currently are 😅 While I understand that some people are more concerned about usability (I am one of these people), a lot of people are also concerned with how the UI looks as well.

For what it's worth, I do plan on rewriting the UI soon, so hopefully, this issue will be fixed soon.

Thank you for contributing, though, I appreciate it!

---

P.S. Would you mind answering https://github.com/TwiN/gatus/issues/224 if you have some time? The screenshot you posted seems to have tons of very interesting use cases I'd love to have some details about. Of course, it's up to you :)

Issue #308, opened 2022-08-03
feat(alerting): Implement ntfy provider


TwiN:
### Describe the feature request

I'd like to create an alerting provider for [binwiederhier/ntfy](https://github.com/binwiederhier/ntfy).

For other examples, see https://github.com/TwiN/gatus/tree/master/alerting/provider

Documentation should also be added in https://github.com/TwiN/gatus/blob/master/README.md#alerting

Here's an example of a similar past pull request: #299 

### Why do you personally want this feature to be implemented?

[binwiederhier/ntfy](https://github.com/binwiederhier/ntfy) allows users to send push notifications to phone/desktop via a PUT/POST request, and I believe this would be an amazing feature to add to Gatus.

### How long have you been using this project?

N/A

### Additional information

There's already an example in ntfy's documentation, but it relies on using the `custom` alert provider: https://ntfy.sh/docs/examples/#gatus (thanks to @nickexyz for adding the example)

```yaml
alerting:
  custom:
    url: "https://ntfy.sh"
    method: "POST"
    body: |
      {
        "topic": "mytopic",
        "message": "[ENDPOINT_NAME] - [ALERT_DESCRIPTION]",
        "title": "Gatus",
        "tags": ["[ALERT_TRIGGERED_OR_RESOLVED]"],
        "priority": 3
      }
    placeholders:
      ALERT_TRIGGERED_OR_RESOLVED:
        TRIGGERED: "warning"
        RESOLVED: "white_check_mark"
```

Still, it should make the implementation easier.

PR #336, opened 2022-10-05
feat(alerting): Implement ntfy provider

TwiN:
## Summary

Closes https://github.com/TwiN/gatus/issues/308

Work remaining:
- [x] Add the documentation on the README.md
- [x] Test it with an actual Ntfy instance (I've only used https://ntfy.sh/docs/examples/#gatus as a reference; I haven't actually tested it yet)

## Checklist

### DIFF
--- a/alerting/provider/ntfy/ntfy.go
+++ b/alerting/provider/ntfy/ntfy.go
@@ -1,0 +1,73 @@
+package ntfy
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/TwiN/gatus/v4/alerting/alert"
+	"github.com/TwiN/gatus/v4/client"
+	"github.com/TwiN/gatus/v4/core"
+)
+
+// AlertProvider is the configuration necessary for sending an alert using Slack
+type AlertProvider struct {
+	URL      string `yaml:"url"`
+	Topic    string `yaml:"topic"`
+	Priority int    `yaml:"priority"`
+
+	// DefaultAlert is the default alert configuration to use for endpoints with an alert of the appropriate type
+	DefaultAlert *alert.Alert `yaml:"default-alert,omitempty"`
+}
+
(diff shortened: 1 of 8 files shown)


TwiN:
If somebody has the time to test it with Ntfy, please feel free 🙏

alerting/provider/ntfy/ntfy.go
```diff
@@ -0,0 +1,84 @@
...
+	} else {
+		message = endpoint.DisplayName()
+	}
+	if resolved {
+		tag = "white_check_mark"
+	} else {
+		tag = "x"
+	}
+	return fmt.Sprintf(`{
+  "topic": "%s",
```
binwiederhier, line 73: This seems dangerous. If the topic contains quotes then the JSON is malformation. Or am I misunderstanding?

TwiN: While you're right in that it would fail, it's not really dangerous for two reasons:
1. The configuration can only be modified by somebody with write access to the configuration file, meaning no external malicious actor can really abuse it.
2. Even if the body is malformed, the worst that could happen is that the alert would not send and an error would be logged instead.

I've tried with the following configuration:
```yaml
alerting:
  ntfy:
    topic: "this-is\"a-test"
```
and this is the error returned:
```
2022/10/09 18:44:30 [watchdog][handleAlertsToTrigger] Failed to send an alert for endpoint=test: call to provider alert returned status code 400: {"code":40017,"http":400,"error":"invalid request: request body must be message JSON","link":"https://ntfy.sh/docs/publish/#publish-as-json"}
```

That said, it'd be a pretty decent idea to add some more validation on start.

binwiederhier: My suggestion would be to just add create a struct and then use `json.EncodeString` to serialize. Then you don't have to deal with any of that.

TwiN: Resolved by c86492d

PR #234, opened 2022-01-12
feat(alerting): Add Google Chat alerting provider

kpolischuk:
1st try
need review
added google chat support
image

PR #341, opened 2022-10-14
fix: add Google Chat to list of alert types when determining valid providers

davwheat:
## Summary

Related to #234.

It seems like this was missed in the PR, rendering Google Chat alerts not working.

## Checklist


TwiN (approved):
Excellent catch!

PR #347, opened 2022-10-19
feat: Add necessary files for PWA

bondarslavik:
## Summary

Implementation of Progressive Web App (PWA)

## Checklist


web/app/public/manifest.json
```diff
@@ -0,0 +1,24 @@
...
+    {
+      "src": "/android-chrome-192x192.png",
+      "sizes": "192x192",
+      "type": "image/png"
+    },
+    {
+      "src": "/android-chrome-512x512.png",
+      "sizes": "512x512",
+      "type": "image/png"
+    }
```
TwiN, line 22: Not really a big fan of having images with names specifically targeting devices/browsers. 

Could these images just be `logo-192x192.png` and `logo-512x512.png` respectively?

web/app/public/index.html
```diff
@@ -8,10 +8,17 @@
     <title>{{ .Title }}</title>
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
+    <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png">
     <link rel="icon" type="image/png" sizes="32x32" href="/favicon-32x32.png" />
     <link rel="icon" type="image/png" sizes="16x16" href="/favicon-16x16.png" />
+    <link rel="manifest" href="/manifest.json">
+    <link rel="mask-icon" href="/safari-pinned-tab.svg" color="#5bbad5">
```
TwiN, line 15: You can call me petty, but after doing a quick search and seeing that only safari supports `mask-icon`, I would personally rather not have line and keep things minimal. 

Don't blame me, blame Safari refusing to implement things that every other browsers support (svg favicons using `rel="icon"`).

TwiN (changes requested):
Could you resolve the conflicts? 🙏 

Also, I feel like perhaps that's a bit too much?

web/app/public/manifest.json
```diff
@@ -0,0 +1,24 @@
+{
+  "id": "gatus",
+  "name": "Gatus",
+  "short_name": "Gatus",
+  "description": "Gatus is an advanced automated status page that lets you monitor your applications and configure alerts to notify you if there's an issue",
+  "lang": "en",
+  "scope": "/",
+  "start_url": "/",
+  "theme_color": "#f7f9fb",
+  "background_color": "#f7f9fb",
```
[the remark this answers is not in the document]
bondarslavik: Taken this color from https://github.com/TwiN/gatus/blob/master/web/app/src/index.css

web/app/public/index.html
bondarslavik: Agree with you, removed this line

web/app/public/index.html
```diff
@@ -8,10 +8,16 @@
...
     <link rel="icon" type="image/png" sizes="32x32" href="/favicon-32x32.png" />
     <link rel="icon" type="image/png" sizes="16x16" href="/favicon-16x16.png" />
+    <link rel="manifest" href="/manifest.json">
     <link rel="shortcut icon" href="/favicon.ico" />
-    <meta property="description" content="{{ .Description }}" />
+    <meta name="description" content="{{ .Description }}" />
+    <meta name="apple-mobile-web-app-status-bar-style" content="black-translucent">
+    <meta name="apple-mobile-web-app-title" content="{{ .Title }}">
+    <meta name="application-name" content="{{ .Title }}">
+    <meta name="theme-color" content="#f7f9fb">
```
TwiN, line 20: Sorry, I missed it during my first check: these should probably all have the same endpoint (either `/>` or `>`) for the sake of consistency.

No need to regenerate the static assets, just updating this file is sufficient.

TwiN (changes requested):

web/app/public/manifest.json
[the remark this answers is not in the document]
TwiN: Fair enough!

web/app/public/index.html
bondarslavik: Hove I can see at now this [file](https://github.com/TwiN/gatus/blob/master/web/app/public/index.html) doesn't have full consistency and use both styles.  I can make all this file in one style, what style do you prefer (`/>` or `>`) ?

TwiN: I personally prefer `/>`, but I'm fine with either :)

bondarslavik: done

TwiN:
@bondarslavik Thank you for the contribution!

Issue #128, opened 2021-07-07
Increase data retention to 30 days


mdraijer:
For the uptime badges (1h, 24h, 7d), is it possible to also support a month or 30d? Maybe even 1y?

sahil87:
Would love this feature

TwiN:
Closed in favor of #242

TwiN:
C

mdraijer:
Do you have any idea about a timeline for this improvement?

mdraijer:
By the way, @TwiN regarding your change of the title of this issue: I see that sometimes the data _is_ retained for at least 30 days, sometimes a lot longer. But in some cases a lot shorter. Probably depending on the 'MaximumNumberOfResults' and 'MaximumNumberOfEvents'? (couldn't find that in the documentation, didn't search the code yet)

So I assume that you meant to say "guarantee retention to 30 days; may be a lot longer if few events happen"?

TwiN:
@mdraijer yeah since people can set their own interval, and these interval can be extremely short (1ms), it'd be a bad idea to save for the past 30 days. What I'll do instead is make MaximumNumberOfResults configurable.

mdraijer:
Agreed, would be sufficient and a neat solution to have that value configurable.
Would that also make it possible to get badges for 30 days / 1 month?

PurseChicken:
I would like to mention that this would be extremely desired from our team as well. It would be great to get 30d, 60d, 90d and 180d intervals.

xeruf:
Yes, would love as well - we had a longer outage and an extended interval could have helped us to trace back the issue more swiftly!

TwiN:
FYI #836 will close this. I know some of you want longer than 30d, but at least it's a significant step up from the previous maximum of 7d.

I haven't had the opportunity to completely test the feature, so when #836 is merged and new latest images are pushed, I'd appreciate if some of you could give it a test (of course, I'm aware it'll take some time before the data populates, but the more testers, the faster I can release)

Thanks!

PR #373, opened 2022-11-20
feat(badge): custom duration (#128)

117503445:
Signed-off-by: 117503445 <t117503445@gmail.com>

## Summary

#128 , custom duration for badge. `{duration}` now has the pattern `<LENGTH>_<DATE_MARKER>`, e.g. 6h,2d,3m,1y.

## Checklist

### DIFF
--- a/controller/handler/badge.go
+++ b/controller/handler/badge.go
@@ -33,6 +33,31 @@ var (
 	badgeColors = []string{badgeColorHexAwesome, badgeColorHexGreat, badgeColorHexGood, badgeColorHexPassable, badgeColorHexBad}
 )
 
+var (
+	whiteList map[string]interface{}
+)
+
+func init() {
+	whiteList = map[string]interface{}{
+		"1h":  nil,
+		"24h": nil,
+		"7d":  nil,
+		"1m":  nil,
+		"3m":  nil,
+		"6m":  nil,
+		"1y":  nil,
+	}
+}
+
+func writeSVG(writer http.ResponseWriter, svg []byte) {
+	writer.Header().Set("Content-Type", "image/svg+xml")
+	writer.Header().Set("Cache-Control", "no-cache, no-store, must-revalidate")
+	writer.Header().Set("Expires", "0")
+	writer.WriteHeader(http.StatusOK)
+
+	_, _ = writer.Write(svg)
+}
+
 // UptimeBadge handles the automatic generation of badge based on the group name and endpoint name passed.
 //
 // Duration should be in the form of a number with date marker, e.g. 6h,2d,3m,1y
@@ -40,9 +65,20 @@ func UptimeBadge(writer http.ResponseWriter, request *http.Request) {
 	variables := mux.Vars(request)
 	duration := variables["duration"]
 
+	if _, exists := whiteList[duration]; !exists {
+		http.Error(writer, "not valid duration", http.StatusBadRequest)
+		return
+	}
+
 	from, err := parseFromTime(duration)
 	if err != nil {
 		http.Error(writer, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	svg, exists := cache.Get(fmt.Sprintf("badge-uptime-%s", duration))
+	if exists {
+		writeSVG(writer, svg.([]byte))
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


controller/handler/badge.go
```diff
@@ -35,22 +35,17 @@ var (
...
-	case "24h":
-		from = time.Now().Add(-24 * time.Hour)
-	case "1h":
-		from = time.Now().Add(-2 * time.Hour) // Because uptime metrics are stored by hour, we have to cheat a little
-	default:
-		http.Error(writer, "Durations supported: 7d, 24h, 1h", http.StatusBadRequest)
+
+	from, err := parseFromTime(duration)
+	if err != nil {
+		http.Error(writer, err.Error(), http.StatusBadRequest)
```
TwiN, line 45: The reason why I allowed only specific durations to be passed as parameter is because calculating the uptime can be expensive, and I'd like that to stay that way.

I'm fine with adding the following values:
- 30d
- 90d
- 180d
- 1y

but we'll have to add a caching mechanism.

TwiN (changes requested):
There's a couple changes missing:
- https://github.com/TwiN/gatus/blob/4f569b7a0e3c8752417beabc1926f17f60e880ec/storage/store/memory/uptime.go#L31-L42
- https://github.com/TwiN/gatus/blob/4f569b7a0e3c8752417beabc1926f17f60e880ec/storage/store/sql/sql.go#L330-L334

But beyond that, some before/after benchmark may be needed.

controller/handler/badge.go
117503445: You are right, unrestricted queries can lead to dos attacks

117503445:
I've come up with these strategies

1. adjust `uptimeCleanUpThreshold` to 1 year, the disadvantage is that it will take up storage space and the query will be slower
2. allow the user to specify the allowed `durations` through the config file, `uptimeCleanUpThreshold` is determined by the maximum `duration`.
3. allow the user to specify the allowed `durations` and `uptimeCleanUpThreshold` through the config file, and  user should ensure that `uptimeCleanUpThreshold` is not lower than the maximum `duration`

I think the second strategy is better, do you have any suggestions?

controller/handler/badge.go
117503445: I've come up with these strategies

1. adjust `uptimeCleanUpThreshold` to 1 year, the disadvantage is that it will take up storage space and the query will be slower
2. allow the user to specify the allowed `durations` through the config file, `uptimeCleanUpThreshold` is determined by the maximum `duration`.
3. allow the user to specify the allowed `durations` and `uptimeCleanUpThreshold` through the config file, and  user should ensure that `uptimeCleanUpThreshold` is not lower than the maximum `duration`

I think the second strategy is better, do you have any suggestions?

TwiN (changes requested):
Great work so far, but there's still some parts missing as well as some performance optimizationTo be honest, I'm still not sure if I'm comfortable with this change.

Let me start off with what's missing:
- storage-related changes are missing, it only modifies the controller. As things are, `1m`, `3m`, `6m` and `1y` will all return the same thing, because the uptime cleanup threshold is set to 10 days (and when that threshold is reached, all uptime entries older than 7 days (`uptimeRetention`) are deleted from the storage. https://github.com/TwiN/gatus/blob/9fc8374a4de2ce905d90355833c7f41deddca891/storage/store/sql/sql.go#L31-L35
- No tests for the above, and while I understand some may think there's an excessive amount of tests in Gatus, had these been a test for this new feature, the issue above would've been caught by said test
- Performance optimizations. More information after this.

As things are designed at the moment, we are storing up to 10 days grouped by hourly interval.
To help you visualize it, this is what the database looks like:
image

This means that we end up with `10 (days) * 24 (hours) = 240` rows per endpoint.

With 300 endpoints, that gives us `300 * 10 * 24 = 72000` rows.

Now, if we don't make some performance improvements, we would essentially end up with `370 (days) * 24 = 8880` rows per endpoint (370 instead of 365, because we don't want to clean up the table every hour; it's easier to just clean up a couple days worth in one go. That's why `uptimeRetention` and `uptimeCleanUpThreshold` have different values).

With 300 endpoints, that gives us `300 * 370 * 24 = 2664000` rows, which I think is quite a lot.

There's many ways we can optimize this. For instance, if we have fixed durations like 1h, 7d, 1m, 3m, 6m and 1y, then it means that once the entries are older than 7 days, we can group them by month. This would allow us to keep the number of rows per endpoint extremely low

controller/handler/badge_test.go
```diff
@@ -86,8 +86,13 @@ func TestBadge(t *testing.T) {
...
-			Path:         "/api/v1/endpoints/core_backend/uptimes/3d/badge.svg",
+			Name:         "badge-uptime-with-invalid-duration-zero-length",
+			Path:         "/api/v1/endpoints/core_backend/uptimes/0d/badge.svg",
+			ExpectedCode: http.StatusBadRequest,
+		},
+		{
+			Name:         "badge-uptime-with-invalid-duration-too-short",
+			Path:         "/api/v1/endpoints/core_backend/uptimes/d/badge.svg",
 			ExpectedCode: http.StatusBadRequest,
 		},
```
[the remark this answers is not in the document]
117503445: Thank you very much for your reply, but I am quite busy these days and it will take me a few days to continue the development of this feature.

Some of my recent commits on this branch were just to experiment with some ideas. I realized that there are a lot of design choices and the final code has to ensure quality and performance.

Thanks again for your advice and careful code review!

117503445:
I didn't realize before that the key to this feature is the performance of the storage layer. I don't think I have the energy to accomplish this feature.

Thank you very much for the code review and good luck with your project!

PR #389, opened 2022-12-19
feat: Support multiple configuration files

### LINKED ISSUE #326: split config files

Currently all config is in a single file

Splitting config into mutliple files can improve readability

for example : 
config/generic.yaml with all global config

```
metrics: true                                                                                                                                                                                                                                                                                                                                                                                                       storage:                                                                                                                                                                                                    
  type: sqlite                                                                                                                                                                                              
  path: data.db                                                                                                                                                                                           remote:                                                                                                                                                                                                     
  instances:                                                                                                                                                                                                
  - endpoint-prefix: "Remonte Node~"                                                                                                                                                                             
    url: "http://8.8.8.8:8080/api/v1/endpoints/statuses"        
```

config/host1.yaml with host1 
```
  - name: host1-ping
    url: "icmp://host1"
    conditions:
      - "[CONNECTED] == true"

  - name: host1-tcp6739
    url: "tcp://host1:6739"
    conditions:
      - "[CONNECTED] == true"
```


Adding a notion of "service"  can be great (so we can link a node with multiples services)

config/host1.yaml with host1 
```
  - name: host1
    service : ping
      url: "icmp://host1"
      conditions:
        - "[CONNECTED] == true"
    service: redis
      url: "tcp://host1:6739"
      conditions:
        - "[CONNECTED] == true"
```


Thanks !

improve readability and maintainability

_No response_

### Additional information

_No response_

henningjanssen:
## Summary
This PR allows configuration to split across multiple files using the common `config.d/`-path style. The approach reads all `*.yml` and `*.yaml` files and merges them in memory into a single file before parsing.  This seemed easier und required fewer changes than parsing each file individually and merging the objects afterwards.

Fixes #326

As I did not work with golang before, feel free to remark on obvious errors and bad patterns.

I envisaged to write tests, but did not figure out yet, how invasive tests are intended to be structured for this project. I hope the chosen way is in you interest.

## Checklist

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -376,10 +376,11 @@ func walkConfigDir(path string, fn fs.WalkDirFunc) error {
 	}
 
 	return filepath.WalkDir(path, func(path string, d fs.DirEntry, err error) error {
-		if nil == d {
+		if nil != err {
 			return nil
 		}
-		if d.IsDir() {
+		
+		if nil == d || d.IsDir() {
 			return nil
 		}
 
--- a/config/config.go
+++ b/config/config.go
@@ -31,10 +31,6 @@ const (
 	// DefaultFallbackConfigurationFilePath is the default fallback path that will be used to search for the
 	// configuration file if DefaultConfigurationFilePath didn't work
 	DefaultFallbackConfigurationFilePath = "config/config.yml"
-
-	// DefaultDistributedConfigurationDirectoryPath is the default path to search for additional configuration files
-	// if a custom directory has not been configured using the GATUS_CONFIG_DIR environment variable
-	DefaultDistributedConfigurationDirectoryPath = "config/config.d"
 )
 
 var (
@@ -93,8 +89,7 @@ type Config struct {
 	// WARNING: This is in ALPHA and may change or be completely removed in the future
 	Remote *remote.Config `yaml:"remote,omitempty"`
 
-	filePath        string    // path to the file from which config was loaded from
-	distributedPath	string	  // path to the distributed directory from which config was loaded from
+	configPath      string    // path to the file or directory from which config was loaded
 	lastFileModTime time.Time // last modification time
 }
 
@@ -113,18 +108,13 @@ func (config *Config) GetEndpointByKey(key string) *core.Endpoint {
 func (config Config) HasLoadedConfigurationFileBeenModified() bool {
 	var lastMod = config.lastFileModTime.Unix()
 
-	// check main config file
-	if fileInfo, err := os.Stat(config.filePath); err == nil {
-		if !fileInfo.ModTime().IsZero() {
-			if config.lastFileModTime.Unix() < fileInfo.ModTime().Unix() {
-				return true
-			}
-		}
+	var fileInfo, err = os.Stat(config.configPath)
+	if nil != err {
+		return false
 	}
 
-	// check distributed files
-	if len(config.distributedPath) > 0 {
-		var err = walkConfigDir(config.distributedPath, func (path string, d fs.DirEntry, err error) error {
+	if fileInfo.IsDir() {
+		err = walkConfigDir(config.configPath, func (path string, d fs.DirEntry, err error) error {
 			if info, err := d.Info(); nil == err && lastMod < info.ModTime().Unix() {
 				return ErrEarlyReturn
 			}
@@ -134,7 +124,7 @@ func (config Config) HasLoadedConfigurationFileBeenModified() bool {
 		return ErrEarlyReturn == err
 	}
 
-	return false
+	return !fileInfo.ModTime().IsZero() && config.lastFileModTime.Unix() < fileInfo.ModTime().Unix()
 }
 
 // UpdateLastFileModTime refreshes Config.lastFileModTime
@@ -144,39 +134,53 @@ func (config *Config) UpdateLastFileModTime() {
 
 // LoadConfiguration loads the full configuration composed from the main configuration file
 // and all composed configuration files
-func LoadConfiguration(filePath string, directoryPath string) (*Config, error) {
+func LoadConfiguration(configPath string) (*Config, error) {
 	var composedContents []byte
-	var usedFilePath string
+	var fileInfo os.FileInfo
+	var usedConfigPath string = ""
 
-	for _, fpath := range []string{filePath, DefaultConfigurationFilePath, DefaultFallbackConfigurationFilePath} {
-		var bytes, err = os.ReadFile(fpath)
-		if nil == err {
-			log.Printf("[config][Load] Reading configuration from configFile=%s", fpath)
-			composedContents = append(composedContents, bytes...)
-			usedFilePath = fpath
-			break
+	for _, cpath := range []string{configPath, DefaultConfigurationFilePath, DefaultFallbackConfigurationFilePath} {
+		if len(cpath) == 0 {
+			continue
 		}
+
+		var err error
+		fileInfo, err = os.Stat(cpath)
+		if nil != err {
+			continue
+		}
+
+		usedConfigPath = configPath
+		break
+	}
+	if len(usedConfigPath) == 0 {
+		return nil, ErrConfigFileNotFound
 	}
 
-	if len(directoryPath) == 0 {
-		directoryPath = DefaultDistributedConfigurationDirectoryPath
-	}
-	walkConfigDir(directoryPath, func(path string, d fs.DirEntry, err error) error {
-		var bytes, rerr = os.ReadFile(path)
-		if nil == rerr {
-			log.Printf("[config][Load] Reading configuration from configFile=%s", path)
-			composedContents = append(composedContents, bytes...)
+	if fileInfo.IsDir() {
+		walkConfigDir(configPath, func(path string, d fs.DirEntry, err error) error {
+			var bytes, rerr = os.ReadFile(path)
+			if nil == rerr {
+				log.Printf("[config][Load] Reading configuration from configFile=%s", path)
+				composedContents = append(composedContents, bytes...)
+			}
+			return nil
+		})
+	} else {
+		var bytes, serr = os.ReadFile(usedConfigPath)
+		if nil == serr {
+			log.Printf("[config][Load] Reading configuration from configFile=%s", configPath)
+			composedContents = bytes
+			usedConfigPath = usedConfigPath
 		}
-		return nil
-	})
+	}
 
 	if len(composedContents) == 0 {
 		return nil, ErrConfigFileNotFound
 	}
 
 	var config, err = parseAndValidateConfigBytes(composedContents)
-	config.filePath = usedFilePath
-	config.distributedPath = directoryPath
+	config.configPath = usedConfigPath
 	config.UpdateLastFileModTime()
 
 	return config, err
--- a/config/config.go
+++ b/config/config.go
@@ -112,10 +112,9 @@ func (config Config) HasLoadedConfigurationFileBeenModified() bool {
 	if nil != err {
 		return false
 	}
-
 	if fileInfo.IsDir() {
 		err = walkConfigDir(config.configPath, func (path string, d fs.DirEntry, err error) error {
-			if info, err := d.Info(); nil == err && lastMod < info.ModTime().Unix() {
+			if info, err := d.Info(); err == nil && lastMod < info.ModTime().Unix() {
 				return ErrEarlyReturn
 			}
 			return nil
--- a/config/config.go
+++ b/config/config.go
@@ -3,8 +3,10 @@ package config
 import (
 	"errors"
 	"fmt"
+	"io/fs"
 	"log"
 	"os"
+	"path/filepath"
 	"time"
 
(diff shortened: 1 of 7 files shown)


config/config.go
```diff
@@ -29,6 +31,10 @@ const (
 	// DefaultFallbackConfigurationFilePath is the default fallback path that will be used to search for the
 	// configuration file if DefaultConfigurationFilePath didn't work
 	DefaultFallbackConfigurationFilePath = "config/config.yml"
+
+	// DefaultDistributedConfigurationDirectoryPath is the default path to search for additional configuration files
+	// if a custom directory has not been configured using the GATUS_CONFIG_DIR environment variable
+	DefaultDistributedConfigurationDirectoryPath = "config/config.d"
```
TwiN, line 37: Does it really make sense to have a default value for this?

Personally, I feel that the best approach would be to have either `GATUS_CONFIG_FILE` or `GATUS_CONFIG_DIR`, not both.

More specifically, when `GATUS_CONFIG_DIR` is set, we'd read every file in the folder defined by `GATUS_CONFIG_DIR`, and if not, we'd fall back to `GATUS_CONFIG_FILE`.

Also, I just want to point out that for never having written Go before, you're doing an excellent job!

TwiN (changes requested):

config/config.go
henningjanssen: I had the default setup for many applications (f.e. nginx, sshd) in mind that use a structure like this
```
.
├── config.d
│   └── my.config
└── main.config
```
and implement distribution with `include config.d/*`-statements which are not supported by YAML. To use only `GATUS_CONFIG_DIR` with fallback to `GATUS_CONFIG_FILE`, use both `GATUS_CONFIG_DIR` and `GATUS_CONFIG_FILE` or use both but without defining a default for `GATUS_CONFIG_DIR` all seem valid to me.
I will adjust this PR during the next few days.

henningjanssen: Now it's reduced to single env `GATUS_CONFIG_PATH` that can contain either a file or a directory path. Are you fine with this approach?

TwiN: That's excellent!

main.go
```diff
@@ -52,14 +52,15 @@ func save() {
...
-	if len(customConfigFile) > 0 {
-		cfg, err = config.Load(customConfigFile)
-	} else {
-		cfg, err = config.LoadDefaultConfiguration()
+func loadConfiguration() (*config.Config, error) {
+	var configPath = os.GetEnv("GATUS_CONFIG_PATH")
+
+	// backwards compatibility
+	if len(configPath) == 0 {
+		configPath = os.GetEnv("GATUS_CONFIG_FILE")
```
TwiN, line 60:
```suggestion
func loadConfiguration() (*config.Config, error) {
	var configPath = os.Getenv("GATUS_CONFIG_PATH")
	// Backwards compatibility
	if len(configPath) == 0 {
		configPath = os.Getenv("GATUS_CONFIG_FILE")
```

TwiN: I was going to commit it myself, but I think you have "allow edits from maintainer" unchecked 😔

config/config.go
```diff
@@ -98,63 +103,87 @@ func (config *Config) GetEndpointByKey(key string) *core.Endpoint {
...
+	var fileInfo, err = os.Stat(config.configPath)
+	if nil != err {
+		return false
 	}
-	return false
+
+	if fileInfo.IsDir() {
+		err = walkConfigDir(config.configPath, func (path string, d fs.DirEntry, err error) error {
+			if info, err := d.Info(); nil == err && lastMod < info.ModTime().Unix() {
+				return ErrEarlyReturn
```
TwiN, line 119:
```suggestion
	if fileInfo.IsDir() {
		err = walkConfigDir(config.configPath, func (path string, d fs.DirEntry, err error) error {
			if info, err := d.Info(); err == nil && lastMod < info.ModTime().Unix() {
				return ErrEarlyReturn
```

config/config.go
```diff
@@ -98,63 +103,87 @@ func (config *Config) GetEndpointByKey(key string) *core.Endpoint {
...
 func (config Config) HasLoadedConfigurationFileBeenModified() bool {
-	if fileInfo, err := os.Stat(config.filePath); err == nil {
-		if !fileInfo.ModTime().IsZero() {
-			return config.lastFileModTime.Unix() != fileInfo.ModTime().Unix()
-		}
+	var lastMod = config.lastFileModTime.Unix()
+
+	var fileInfo, err = os.Stat(config.configPath)
+	if nil != err {
+		return false
```
TwiN, line 113:
```suggestion
	lastMod := config.lastFileModTime.Unix()
	fileInfo, err : = os.Stat(config.configPath)
	if err != nil {
		return false
```

config/config.go
```diff
@@ -98,63 +103,87 @@ func (config *Config) GetEndpointByKey(key string) *core.Endpoint {
...
+			if info, err := d.Info(); nil == err && lastMod < info.ModTime().Unix() {
+				return ErrEarlyReturn
+			}
+			return nil
+		})
+
+		return ErrEarlyReturn == err
+	}
+
+	return !fileInfo.ModTime().IsZero() && config.lastFileModTime.Unix() < fileInfo.ModTime().Unix()
```
TwiN, line 127:
```suggestion
			return nil
		})
		return err == ErrEarlyReturn
	}
	return !fileInfo.ModTime().IsZero() && config.lastFileModTime.Unix() < fileInfo.ModTime().Unix()
```

config/config.go
```diff
@@ -98,63 +103,87 @@ func (config *Config) GetEndpointByKey(key string) *core.Endpoint {
...
+
+	for _, cpath := range []string{configPath, DefaultConfigurationFilePath, DefaultFallbackConfigurationFilePath} {
+		if len(cpath) == 0 {
+			continue
+		}
+
+		var err error
+		fileInfo, err = os.Stat(cpath)
+		if nil != err {
+			continue
```
TwiN, line 150:
```suggestion
		var err error
		fileInfo, err = os.Stat(cpath)
		if err != nil {
			continue
```

config/config.go
```diff
@@ -98,63 +103,87 @@ func (config *Config) GetEndpointByKey(key string) *core.Endpoint {
...
-	cfg, err := Load(DefaultConfigurationFilePath)
-	if err != nil {
-		if err == ErrConfigFileNotFound {
-			return Load(DefaultFallbackConfigurationFilePath)
+	if fileInfo.IsDir() {
+		walkConfigDir(configPath, func(path string, d fs.DirEntry, err error) error {
+			var bytes, rerr = os.ReadFile(path)
+			if nil == rerr {
+				log.Printf("[config][Load] Reading configuration from configFile=%s", path)
+				composedContents = append(composedContents, bytes...)
```
TwiN, line 165:
```suggestion
	if fileInfo.IsDir() {
		walkConfigDir(configPath, func(path string, d fs.DirEntry, err error) error {
			bytes, rerr := os.ReadFile(path)
			if rerr == nil {
				log.Printf("[config][Load] Reading configuration from configFile=%s", path)
				composedContents = append(composedContents, bytes...)
```

config/config.go
```diff
@@ -98,63 +103,87 @@ func (config *Config) GetEndpointByKey(key string) *core.Endpoint {
...
+			}
+			return nil
+		})
+	} else {
+		var bytes, serr = os.ReadFile(usedConfigPath)
+		if nil == serr {
+			log.Printf("[config][Load] Reading configuration from configFile=%s", configPath)
+			composedContents = bytes
+			usedConfigPath = usedConfigPath
```
TwiN, line 174:
```suggestion
		bytes, serr := os.ReadFile(usedConfigPath)
		if serr == nil {
			log.Printf("[config][Load] Reading configuration from configFile=%s", configPath)
			composedContents = bytes
```

config/config.go
```diff
@@ -98,63 +103,87 @@ func (config *Config) GetEndpointByKey(key string) *core.Endpoint {
...
+	if len(composedContents) == 0 {
+		return nil, ErrConfigFileNotFound
 	}
-	return
+
+	var config, err = parseAndValidateConfigBytes(composedContents)
+	config.configPath = usedConfigPath
+	config.UpdateLastFileModTime()
+
+	return config, err
```
TwiN, line 186:
```suggestion
	config, err := parseAndValidateConfigBytes(composedContents)
	config.configPath = usedConfigPath
	config.UpdateLastFileModTime()
	return config, err
```

config/config.go
```diff
@@ -332,3 +361,29 @@ func validateAlertingConfig(alertingConfig *alerting.Config, endpoints []*core.E
...
+		}
+
+		ext := filepath.Ext(path)
+		if ".yml" != ext && ".yaml" != ext {
+			return nil
+		}
+
+		return fn(path, d, err)
+	})
+}
```
TwiN, line 389:
```suggestion
// walkConfigDir is a wrapper for filepath.WalkDir that strips directories and non-config files
func walkConfigDir(path string, fn fs.WalkDirFunc) error {
	if len(path) == 0 {
		// If the user didn't provide a directory, we'll just use the default config file, so we can return nil now.
		return nil
	}
	return filepath.WalkDir(path, func(path string, d fs.DirEntry, err error) error {
		if err != nil {
			return nil
		}
		if d == nil || d.IsDir() {
			return nil
		}
		ext := filepath.Ext(path)
		if ext != ".yml" && ext != ".yaml" {
			return nil
		}
		return fn(path, d, err)
	})
}
```

config/config_test.go
```diff
@@ -29,19 +29,40 @@ import (
...
+		t.Error("Should not have returned an error, because configuration files exist at the provided path for distributed files")
+	}
+}
+
+func TestLoadCombinedConfiguration(t *testing.T) {
+	_, err := LoadConfiguration("../test-conf/empty-conf.d/")
+	if nil == err {
+		t.Error("Should have returned an error, because the configuration directory does not contain any configuration files")
+	}
+}
```
TwiN, line 64: Note to self: rewrite that in a single Test (use test table method) and dynamically generate the config directories and files instead of using the `test-conf` folder. Just make sure to defer clean up

TwiN (changes requested):

config/config.go
```diff
@@ -98,63 +103,80 @@ func (config *Config) GetEndpointByKey(key string) *core.Endpoint {
...
-// HasLoadedConfigurationFileBeenModified returns whether the file that the
+// HasLoadedConfigurationFileBeenModified returns whether one of the file that the
 // configuration has been loaded from has been modified since it was last read
 func (config Config) HasLoadedConfigurationFileBeenModified() bool {
-	if fileInfo, err := os.Stat(config.filePath); err == nil {
-		if !fileInfo.ModTime().IsZero() {
-			return config.lastFileModTime.Unix() != fileInfo.ModTime().Unix()
-		}
+	lastMod := config.lastFileModTime.Unix()
+	fileInfo, err : = os.Stat(config.configPath)
```
TwiN, line 110:
```suggestion
	fileInfo, err := os.Stat(config.configPath)
```

TwiN (approved):

TwiN:
@henningjanssen Thank you for your contribution!

TwiN:
By the way, I did some testing locally and it turns out your implementation doesn't work 😓 
I've prevented it from making it to latest and I'll see if I can fix it, but long story short, two files with `endpoints:` doesn't actually merge the endpoints, the second `endpoints:` just overwrite the first one.

TwiN:
@henningjanssen I'm really sorry; I had to revert it because I wasn't able to fix it before signing off for the day (it's nearly 4AM for me) and I don't want to leave the master branch in an inaccurate state. I'll have a look at it later.

henningjanssen:
Yeah, because there is no "correct" way of handling duplicate nested keys, especially if they have different types. But this should also be an issue with duplicating keys inside the same yaml-file as this implementation sees no difference between a single and multiple files due to concatenation of byte streams before parsing.

The advantage of this approach is to have f.e. alert-config (containing access tokens) separated from endpoints so that the latter can be checked into version control while the former is deployed on the host directly.

While probably breaking the YAML protocol, splitting endpoints into multiple files is possible by omitting the top-level key. The load order can be influenced by filename (`10-xxx.yml`, `70-yyy.yml`). A possibility to make this usage of this hack more obvious could be to additionally support usage of a file extension like `.pyml` (partial yml).

TwiN:
Oh it's definitely possible, don't worry :P 

Personally, I think there would be no advantage implementing this feature if we can't define endpoints in multiple files -- that's what people want the most, because the endpoint configuration is _the_ longest configuration.

It's going to require some testing, but long story short, I'll make it deep merge the yaml files, append keys that have a slice as value rather than overwrite other slices, and prevent duplicate keys with primitive as value.

I'll re-add your change with the fixes in #396

TwiN:
Surprisingly, outside of yq (which had too many dependencies for my liking), I haven't found any decent libraries to deep merge YAML files, so I made one: https://github.com/TwiN/deepmerge

Granted it could be because I haven't searched for _that_ long, but 🤷‍♂️

TwiN:
Superseded by #396

PR #403, opened 2023-01-16
feat(condition): Support floating point

jmcker:
## Summary

- Adds support for floating point numbers in conditions.
- Fixes comparison of large integers.
- Adds the `str()` function.
- Resolves non-numeric values to `NaN` for numeric-only operators.

Support for `<`, `<=`, `>`, and `>=` was relatively straightforward since numbers were already parsed.
`==` and `!=` now check if both operands are numeric before falling back to string comparison.

Large integers from `[BODY]` were being parsed to [`float64_t` by `json.Unmarshal`][1] and then formatted in scientific notation by [`Sprintf("%v")` ][2], which made them unparseable as an integer in `sanitizeAndResolveNumerical`.

[1]: https://pkg.go.dev/encoding/json#Unmarshal:~:text=for%20JSON%20booleans-,float64%2C%20for%20JSON%20numbers,-string%2C%20for%20JSON]
[2]: https://pkg.go.dev/fmt#:~:text=float32%2C%20complex64%2C%20etc%3A%20%25g

Added the `str()` function to opt out of the numeric parsing during `==` and `!=` and keep the string comparison behavior.

Numbers are now passed around as `float64_t`, which cannot represent every `int64_t`.
`[BODY]` values were already `float64_t` during deserialization, and `float64_t` can represent `> 285,000` years in milliseconds before losing precision, so I think this should be OK.

## Checklist

### DIFF
--- a/core/condition.go
+++ b/core/condition.go
@@ -3,6 +3,7 @@ package core
 import (
 	"errors"
 	"fmt"
+	"math"
 	"strconv"
 	"strings"
 	"time"
@@ -328,8 +329,8 @@ func sanitizeAndResolveNumerical(list []string, result *Result) (parameters []st
 		if duration, err := time.ParseDuration(element); duration != 0 && err == nil {
 			resolvedNumericalParameters = append(resolvedNumericalParameters, float64(duration.Milliseconds()))
 		} else if number, err := strconv.ParseFloat(element, 64); err != nil {
-			// Default to 0 if the string couldn't be converted to an integer
-			resolvedNumericalParameters = append(resolvedNumericalParameters, 0)
+			// Default to NaN if the string couldn't be converted to a number
+			resolvedNumericalParameters = append(resolvedNumericalParameters, math.NaN())
 		} else {
 			resolvedNumericalParameters = append(resolvedNumericalParameters, number)
 		}
--- a/core/condition.go
+++ b/core/condition.go
@@ -70,6 +70,12 @@ const (
 	// Usage: [IP] == any(1.1.1.1, 1.0.0.1)
 	AnyFunctionPrefix = "any("
 
+	// StrFunctionPrefix is the prefix for the str function
+	//
+	// Usage: [BODY].val == str(55e3)
+	//        55 != str(55.0)
+	StrFunctionPrefix = "str("
+
 	// FunctionSuffix is the suffix for all functions
 	FunctionSuffix = ")"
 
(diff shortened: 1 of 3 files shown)


TwiN (commented):
While I completely agree that Gatus should support floating point numbers (and I cannot believe I've never implemented nor tested that before, as it should be a fairly common use case 🤦‍♂️), I have mixed feelings about the new `str` function you also added.

If you truly _need_ the `str` function, we can discuss it in a separate issue/PR, but I'd prefer to keep it outside of the changes you've made to support floating point.

jmcker:
I personally don't need it as long as you're ok with the new behavior of `isEqual`. Even if someone were checking for the string `"1e3"`, it would be parsed and compared as `1000` with the new changes. Probably not a common case, but `str` was an escape hatch if someone needed to avoid the parsing.

I'll revert the `str` commit later today or tomorrow and can move it to a separate PR if you want to include it.

core/condition.go
```diff
@@ -216,7 +223,28 @@ func isEqual(first, second string) bool {
...
+		}
 	}
+
+	// If both are numeric, force a numeric comparison.
+	// Ensures valid comparison for numbers expressed in decimal, scientific, etc.
+	firstNumeric, firstParseErr := strconv.ParseFloat(first, 64)
+	secondNumeric, secondParseErr := strconv.ParseFloat(second, 64)
+	if firstParseErr == nil && secondParseErr == nil {
+		return firstNumeric == secondNumeric
+	}
```
TwiN, line 246: This comment seems to imply that `1e3` would only be parsed as `1000` if both are numeric. Am I misunderstanding something?

If it `1e3` is always parsed as `1000`, then this isn't a change I can accept right now, because that would technically be a breaking change, even if the `str()` function was added.

jmcker:
I apologize that I haven't been able to get back to this. Thanks for releasing #434. That fixes the "large integer" comparisons that initially brought me here.

core/condition.go
jmcker: I think `{ "value": 1e3 }` always gets parsed as 1000. The behavior does change for `{ "value": "1e3" }` (the string `"1e3"`).

If both sides of the comparison came in as actual numbers (which get converted to strings by [jsonpath/Sprintf](https://github.com/TwiN/gatus/blob/0c92534432f4e3bd2561810b6a38d6c1ab6ce67f/jsonpath/jsonpath.go#L61-L63)) or are strings that can be parsed as numbers, they'll trigger [line 245](https://github.com/TwiN/gatus/pull/403/files#diff-c24e9639766cd2b5cdce77b082fa734e67b7d518d7e4b66354ec373bd47e51deR245) right now. If either can't be parsed as a number, the behavior is the same as before and will continue to use string comparison.

The change to handling `"1e3"` and the reduction in range from `int64` to `float64` are probably breaking changes.

TwiN:
No worries!

I'm going to close this for now, as we at the very least have support for the integer portion of floating numbers thanks to #434.
I'll keep this PR in mind and might re-visit later to add decimal support.

PR #427, opened 2023-02-12
fix(config): Support `$$` in config for literal `$`

### LINKED ISSUE #422: Allow inserting a literal $ in the endpoints.body

We have a graphQL endpoint that we monitor sending a specific query with some variables. But when Gatus parses the configuration file, it expands the variables from this query as if they were environment variables.

An example of query with variables from [Graphql.org](https://graphql.org/learn/queries/#variables):

```graphql
query HeroNameAndFriends($episode: Episode) {
  hero(episode: $episode) {
    name
    friends {
      name
    }
  }
}
```

I ended up with a hacky way to have the literal `$` by creating an environment variable `DOLLAR=$` and changing the query to:

```graphql
query HeroNameAndFriends(${DOLLAR}episode: Episode) {
  hero(episode: ${DOLLAR}episode) {
    name
    friends {
      name
    }
  }
}
```

But I'd be more comfortable without having to circumvent the env variable expansion. I tried multiple ways to write the string in YAML, all expanded the `$episode` to ``.

Ugly hacks are cool and and stuff. But it is still an ugly hack...

~1 month

### Additional information

_No response_

TwiN:
## Summary

This allows escaping `$` in the configuration file by prefixing it with another `$` (i.e. to escape a dollar sign, users have to use `$$`).

Example:
```yaml
endpoints:
  - name: example
    url: https://example.org/docs/how-to-create-config-file-without-including-literal-password-in-it
    conditions:
      - "[BODY] == pat(*$$PASSWORD*)"
```
(for verbosity, the above endpoint assumes that there's `$PASSWORD` in the body)

Fixes #422 

## Checklist

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/TwiN/deepmerge"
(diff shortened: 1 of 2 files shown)


PurseChicken:
I am glad I found this. I was banging my head forever on why I kept getting a 422 status code back. The variables in the body needed to be escaped with another $. It would be helpful if this was in the documentation.

Issue #433, opened 2023-02-15
JSON Parser is not working as expected


antikilahdjs:
### Describe the bug

Hello team, sorry to ask it but am facing some issues to use my condition from json result to monitor my application.

My JSON have this informations:

##########################################
```json
{
  "status": "success",
  "data": {
    "resultType": "vector",
    "result": [
      {
        "metric": {
          "consumergroup": "data-4",
          "topic": "sonar"
        },
        "value": [
          1676481020.849,
          "217.20000000000002"
        ]
      }
    ]
  }
}
```
##########################################

### What do you see?

I would like to monitor using my condition based if my value is higher than 400 then health check will show something in GUI but got offline and not monitor it.

My config is:

config:
  endpoints:
  - name: data-logger-status
    url: "http://myjson-url:9090/api/v1/query?"
    interval: 30s
    conditions:
      - "[BODY].status == success"
      - "[BODY].data.result[0].value[1] > 400"
      - "[RESPONSE_TIME] < 600"

### What do you expect to see?

Not monitorinig my application as expected

![image](https://user-images.githubusercontent.com/33559260/219108917-339891d1-f5e7-443b-b277-9bb668b44e1d.png)

![image](https://user-images.githubusercontent.com/33559260/219108966-6515cca6-f492-452d-a291-782d79f47a03.png)

### List the steps that must be taken to reproduce this issue

1 - Install using normal way
2 - Add the Endpoint to get a API result

### Version

v5.3.0

### Additional information

Somebody can help me to understand where I am failed to configure or if my conditions has been created using wrong way.

I would like expected monitoring using my condition based in my JSON if my value > 400 then got some error in GUI or something like that.

Thank you

TwiN:
Looks to be related to #403.

I can't work on supporting full decimal support right now, but I can at least support the non-decimal part of it (since it's better than just defaulting every numbers with a decimal value to 0).

PR #434, opened 2023-02-16
fix(condition): Partially support numbers with floating point

TwiN:
## Summary

Fixes https://github.com/TwiN/gatus/issues/433

Does not add support for decimal numbers, but it converts float64 to int64.
The reason why I'm not just using float64 instead of int64 is because float64 does not support all the numbers that int64 supports, which means this would be a breaking change. Instead, this change at least supports the non-decimal part of floating point numbers.

This is an improvement over the current implementation, as right now, numbers with decimals are just converted to 0 when compared using a non-equal operator

## Checklist


TwiN:
Related to #403

PR #442, opened 2023-02-22
add cron-schedule support to alerts by changing IsEnabled result

umputun:
The change is back-compatible. It suppresses alerts outside the user-defined crontab format schedule by setting enabled to false.

The goal is to suppress alerts with user-provided crontab formatted settings.  It is somewhat similar to the current `maintenance` but allows more flexibility in setting the schedule.  For example, if `cron-schedule` is set to `* 9-18 * *` the alert will be disabled before 9:00 and after 18:59. I think, adding it  on the `Alert` level is the simplest solution here with the minimal code change.

See discussion https://github.com/TwiN/gatus/discussions/250. 

## Checklist

_Note: I'm not sure what is the proper section in the readme to add this info_


TwiN:
I will close this, because the implementation I'd like to see for endpoint-specific alerts is the same as the one used for maintenance. For the sake of consistency.

PR #444, opened 2023-03-06
update go-ping dependency to pro-ping (maintained fork)

floj:
Hi, thanks for this great project. We use it to monitor internal service availability. We did this before using blackbox_exporter from Prometheus, but Gatus is much easier to configure and understand. Great work!

I recently wanted to add a health check via ICMP and noticed that it does not work on Linux out of the box. Quick check in the code referred to https://github.com/go-ping/ping#linux and this solved the problem.
By checking go-ping I noticed the comment on top if the README stating that the library is deprecated and one should switch to the maintained fork https://github.com/prometheus-community/pro-bing. This is what this PR does.

I additionally added a short sentence to the Gatus READMEs ICMP section, pointing Linux users to the relevant section of pro-ping.

## Checklist


TwiN (approved):
LGTM! Thank you for the contribution

PR #457, opened 2023-04-22
refactor: Modify implementation of TLS

TwiN:
## Summary

This PR modifies the implementation of TLS, which was originally implemented in #322

## Checklist

---

@chr1st1ank if you could test it on your end, I would appreciate it (just to make sure I didn't accidentally break anything 😂), but no pressure.


chr1st1ank:
@TwiN , yes I tested the new version locally and it works perfectly with the same config and certificates as before.
Nice move mit the `omitempty` for the whole TLSConfig object. That makes the error handling easier.

Issue #466, opened 2023-05-05
Make MaximumNumberOfResults and MaximumNumberOfEvents configurable


skhokhlov:
### Describe the feature request

Right now only 100 checks and 50 events are stored. If we have frequent checks, for instance with 10s internal, we will store results only for 16 last minutes which is not enough for retrospective analysis. It would be very great to configure the numbers. 

### Why do you personally want this feature to be implemented?

We would like to store old data. Currently, for this we have to do checks less frequently.

BassSingh:
I like this :)

mensinda:
I also ran into this, and decided to work around this limitation manually. Turns out, increasing the hard coded values is relatively easy. For anyone else who runs into this limitation until #476 is merged:

```diff
diff --git a/storage/store/common/limits.go b/storage/store/common/limits.go
index 92007220..a5187bb8 100644
--- a/storage/store/common/limits.go
+++ b/storage/store/common/limits.go
@@ -2,8 +2,8 @@ package common

 const (
 	// MaximumNumberOfResults is the maximum number of results that an endpoint can have
-	MaximumNumberOfResults = 100
+	MaximumNumberOfResults = 1000

 	// MaximumNumberOfEvents is the maximum number of events that an endpoint can have
-	MaximumNumberOfEvents = 50
+	MaximumNumberOfEvents = 500
 )
diff --git a/web/app/src/components/EndpointGroup.vue b/web/app/src/components/EndpointGroup.vue
index dbc1fa63..aa801014 100644
--- a/web/app/src/components/EndpointGroup.vue
+++ b/web/app/src/components/EndpointGroup.vue
@@ -18,7 +18,7 @@
       <slot v-for="(endpoint, idx) in endpoints" :key="idx">
         <Endpoint
             :data="endpoint"
-            :maximumNumberOfResults="20"
+            :maximumNumberOfResults="40"
             @showTooltip="showTooltip"
             @toggleShowAverageResponseTime="toggleShowAverageResponseTime" :showAverageResponseTime="showAverageResponseTime"
         />
diff --git a/web/app/src/components/Pagination.vue b/web/app/src/components/Pagination.vue
index 3f909e1f..d88ab31d 100644
--- a/web/app/src/components/Pagination.vue
+++ b/web/app/src/components/Pagination.vue
@@ -1,7 +1,7 @@
 <template>
   <div class="mt-3 flex">
     <div class="flex-1">
[...]

PR #476, opened 2023-05-12
feat(config): Make MaximumNumberOfResults and MaximumNumberOfEvents configurable

skhokhlov:
## Summary

#466 

## Checklist


config/config.go
```diff
@@ -260,6 +260,7 @@ func parseAndValidateConfigBytes(yamlBytes []byte) (config *Config, err error) {
 			return nil, err
 		}
 	}
+	config.UI.MaximumNumberOfResults = config.Storage.MaximumNumberOfResults
```
skhokhlov, line 263: I'm not sure about this part, but I think that it's a bad idea to have separate configuration for MaximumNumberOfResults in UI

BassSingh:
@TwiN When you get a chance could you review this please? This looks so awesome and I think it will really benefit Gatus!

README.md
```diff
@@ -296,12 +296,14 @@ Here are some examples of conditions you can use:
...
+| `storage`                        | Storage configuration                                                                                                                              | `{}`       |
+| `storage.path`                   | Path to persist the data in. Only supported for types `sqlite` and `postgres`.                                                                     | `""`       |
+| `storage.type`                   | Type of storage. Valid types: `memory`, `sqlite`, `postgres`.                                                                                      | `"memory"` |
+| `storage.caching`                | Whether to use write-through caching. Improves loading time for large dashboards. <br />Only supported if `storage.type` is `sqlite` or `postgres` | `false`    |
+| `storage.maximumNumberOfResults` | The maximum number of results that an endpoint can have                                                                                            | `100`      |
+| `storage.maximumNumberOfEvents`  | The maximum number of events that an endpoint can have                                                                                             | `50`       |
```
TwiN, line 306: For consistency, `maximumNumberOfResults` and `maximumNumberOfEvents` should be `maximum-number-of-results` and `maximum-number-of-events` respectively.

config/ui/ui.go
```diff
@@ -22,12 +23,13 @@ var (
...
-	Logo        string   `yaml:"logo,omitempty"`        // Logo to display on the page
-	Link        string   `yaml:"link,omitempty"`        // Link to open when clicking on the logo
-	Buttons     []Button `yaml:"buttons,omitempty"`     // Buttons to display below the header
+	Title                  string   `yaml:"title,omitempty"`       // Title of the page
+	Description            string   `yaml:"description,omitempty"` // Meta description of the page
+	Header                 string   `yaml:"header,omitempty"`      // Header is the text at the top of the page
+	Logo                   string   `yaml:"logo,omitempty"`        // Logo to display on the page
+	Link                   string   `yaml:"link,omitempty"`        // Link to open when clicking on the logo
+	Buttons                []Button `yaml:"buttons,omitempty"`     // Buttons to display below the header
+	MaximumNumberOfResults int      // MaximumNumberOfResults to display on the page
```
TwiN, line 32: I'd rather not make this configurable at the UI level & let it be handled by the storage configuration. Since the UI supports pagination, we don't have to worry about this at the UI.

controller/handler/endpoint_status_test.go
```diff
@@ -153,7 +158,13 @@ func TestEndpointStatuses(t *testing.T) {
 	// Can't be bothered dealing with timezone issues on the worker that runs the automated tests
 	firstResult.Timestamp = time.Time{}
 	secondResult.Timestamp = time.Time{}
-	router := CreateRouter(&config.Config{Metrics: true})
+	router := CreateRouter(&config.Config{
+		Metrics: true,
+		Storage: &storage.Config{
+			MaximumNumberOfResults: storage.DefaultMaximumNumberOfResults,
+			MaximumNumberOfEvents:  storage.DefaultMaximumNumberOfEvents,
+		},
```
TwiN, line 166: This should not be necessary. 

For test cases, the default values should be set in `storage`'s `Initialize` function, which is called by `https://github.com/TwiN/gatus/blob/3350e8144376c1fbe002469248af0427cfe985fa/storage/store/store.go#L75-L85 for tests.

As for how it should be handled for non-test cases: https://github.com/TwiN/gatus/blob/3350e8144376c1fbe002469248af0427cfe985fa/storage/config.go#L31 (the latter seems to have already been done)

controller/handler/util.go
```diff
@@ -13,12 +12,9 @@ const (
 
 	// DefaultPageSize is the default page siZE to use if none is specified or an invalid value is provided
 	DefaultPageSize = 20
-
-	// MaximumPageSize is the maximum page size allowed
-	MaximumPageSize = common.MaximumNumberOfResults
 )
 
-func extractPageAndPageSizeFromRequest(r *http.Request) (page int, pageSize int) {
+func extractPageAndPageSizeFromRequest(r *http.Request, cfg *config.Config) (page int, pageSize int) {
```
TwiN, line 17: It feels kind of overkill to pass the encore configuration when all you need is the `MaximumNumberOfResults`

storage/config.go
```diff
@@ -25,6 +30,12 @@ type Config struct {
 	// as they happen, also known as the write-through caching strategy.
 	// Does not apply if Config.Type is not TypePostgres or TypeSQLite.
 	Caching bool `yaml:"caching,omitempty"`
+
+	// MaximumNumberOfResults is the maximum number of results that an endpoint can have
+	MaximumNumberOfResults int `yaml:"maximumNumberOfResults,omitempty"`
+
+	// MaximumNumberOfEvents is the maximum number of events that an endpoint can have
+	MaximumNumberOfEvents int `yaml:"maximumNumberOfEvents,omitempty"`
```
TwiN, line 38: As mentioned in another review comment, `maximumNumberOfResults` and `maximumNumberOfEvents` should be `maximum-number-of-results` and `maximum-number-of-events` respectively.

controller/handler/util_test.go
```diff
@@ -2,6 +2,8 @@ package handler
 
 import (
 	"fmt"
+	"github.com/TwiN/gatus/v5/config"
+	"github.com/TwiN/gatus/v5/storage"
```
TwiN, line 6: The imports should be split in two groups: imports from the standard library, and external imports.

In other words, this should be 
```go
import (
	"fmt"
	"net/http"
	"testing"

	"github.com/TwiN/gatus/v5/config"
	"github.com/TwiN/gatus/v5/storage"
)
```

storage/store/sql/sql.go
```diff
@@ -56,17 +54,34 @@ type Store struct {
...
+
+	// maximumNumberOfResults is the maximum number of results that an endpoint can have
+	maximumNumberOfResults int
+	// maximumNumberOfEvents is the maximum number of events that an endpoint can have
+	maximumNumberOfEvents int
+
+	// eventsCleanUpThreshold is a maximum number of events before triggering a clean up
+	eventsCleanUpThreshold int
+	// resultsCleanUpThreshold is a maximum number of results before triggering a clean up
+	resultsCleanUpThreshold int
```
TwiN, line 66: Instead of defining `eventsCleanUpThreshold` and `resultsCleanUpThreshold` in the store, we should create constants named something like `eventsAboveMaximumCleanUpThreshold` and `resultsAboveMaximumCleanUpThreshold` & set both of these to `10`.

Then we can just compute the threshold based on `store.maximumNumberOfEvents + eventsAboveMaximumCleanUpThreshold` and `store.maximumNumberOfResults + resultsAboveMaximumCleanUpThreshold`.

controller/handler/endpoint_status_test.go
TwiN: After reviewing changes further down, it seems like you've also done the change in `storage.Initialize`, so I'm not sure defining a storage config here is even necessary 🤔

TwiN (changes requested):
Awesome work!

There's still a few more changes I want to look closer at, but let's start with this.

config/ui/ui.go
skhokhlov: Yes, I'm agree, that's why I did this https://github.com/TwiN/gatus/pull/476/files#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R262 It's a workaround to pass this value to ui, we need to to calculate number of pages, because it's hardcoded in the ui https://github.com/TwiN/gatus/pull/476/files#diff-1c806791cec1d6fff630b04440dfb90fff5a34ef3a17cadadfbeafa2c26dcfe2L4

controller/handler/endpoint_status_test.go
skhokhlov: So, I cannot just rely on `Initialize` function because I need the config initialised before `store.Get()` call here https://github.com/TwiN/gatus/pull/476/files#diff-8cfab31fa8754c8f5da1e4f0222ea131e70dead2ebfb3a77b14950919627026dR103

config/ui/ui.go
TwiN: Ah that makes sense! Could you perhaps add a newline above `MaximumNumberOfResults` to separate it from the other parameters?

Also perhaps updating the documentation to make it clear why this field is different from the others (i.e. it's not configurable because we're passing it from the UI?)

---

_Alternatively, I wonder if we should just not pass this at all & when a user tries to go to a page with no results, just mentioned that there's no results._

controller/handler/util_test.go
```diff
@@ -52,10 +55,15 @@ func TestExtractPageAndPageSizeFromRequest(t *testing.T) {
...
+	storageCfg := &storage.Config{}
+	storageCfg.ValidateAndSetDefaults()
+	cfg := &config.Config{
+		Storage: storageCfg,
+	}
 	for _, scenario := range scenarios {
 		t.Run("page-"+scenario.Page+"-pageSize-"+scenario.PageSize, func(t *testing.T) {
 			request, _ := http.NewRequest("GET", fmt.Sprintf("/api/v1/statuses?page=%s&pageSize=%s", scenario.Page, scenario.PageSize), http.NoBody)
-			actualPage, actualPageSize := extractPageAndPageSizeFromRequest(request)
+			actualPage, actualPageSize := extractPageAndPageSizeFromRequest(request, cfg.Storage.MaximumNumberOfResults)
```
TwiN, line 66: Is there really a need to initialize a storage config here since we're just passing `cfg.Storage.MaximumNumberOfResults`?

I think we'd be better off removing 
```
storageCfg := &storage.Config{}
	storageCfg.ValidateAndSetDefaults()
	cfg := &config.Config{
		Storage: storageCfg,
	}
```
adding `MaximumNumberOfResults` to the scenario struct definition and just doing
```
actualPage, actualPageSize := extractPageAndPageSizeFromRequest(request, scenario.MaximumNumberOfResults)
```
That way we can properly test the behavior of `extractPageAndPageSizeFromRequest` with different `MaximumNumberOfResults` values.

storage/config.go
```diff
@@ -38,5 +49,11 @@ func (c *Config) ValidateAndSetDefaults() error {
 	if c.Type == TypeMemory && len(c.Path) > 0 {
 		return ErrMemoryStorageDoesNotSupportPath
 	}
+	if c.MaximumNumberOfResults == 0 {
+		c.MaximumNumberOfResults = DefaultMaximumNumberOfResults
+	}
+	if c.MaximumNumberOfEvents == 0 {
+		c.MaximumNumberOfEvents = DefaultMaximumNumberOfEvents
+	}
```
TwiN, line 57:
```suggestion
	if c.MaximumNumberOfResults <= 0 {
		c.MaximumNumberOfResults = DefaultMaximumNumberOfResults
	}
	if c.MaximumNumberOfEvents <= 0 {
		c.MaximumNumberOfEvents = DefaultMaximumNumberOfEvents
	}
```

storage/store/sql/sql_test.go
```diff
@@ -390,7 +393,7 @@ func TestStore_BrokenSchema(t *testing.T) {
 		t.Fatal("expected no error, got", err.Error())
 	}
 	// Break
-	_, _ = store.db.Exec("DROP TABLE endpoints")
+	_, _ = store.db.Exec("drop table endpoints")
```
TwiN, line 396: Can you revert this change?

storage/store/sql/sql_test.go
```diff
@@ -419,7 +422,7 @@ func TestStore_BrokenSchema(t *testing.T) {
 		t.Fatal("expected no error, got", err.Error())
 	}
 	// Break
-	_, _ = store.db.Exec("DROP TABLE endpoint_events")
+	_, _ = store.db.Exec("drop table endpoint_events")
```
TwiN, line 425: Can you revert this change?

storage/store/sql/sql_test.go
```diff
@@ -435,7 +438,7 @@ func TestStore_BrokenSchema(t *testing.T) {
 		t.Fatal("expected no error, got", err.Error())
 	}
 	// Break
-	_, _ = store.db.Exec("DROP TABLE endpoint_results")
+	_, _ = store.db.Exec("drop table endpoint_results")
```
TwiN, line 441: Can you revert this change?

storage/store/sql/sql_test.go
```diff
@@ -464,7 +467,7 @@ func TestStore_BrokenSchema(t *testing.T) {
 		t.Fatal("expected no error, got", err.Error())
 	}
 	// Break
-	_, _ = store.db.Exec("DROP TABLE endpoint_uptimes")
+	_, _ = store.db.Exec("drop table endpoint_uptimes")
```
TwiN, line 470: Can you revert this change?

storage/store/sql/sql_test.go
```diff
@@ -451,7 +454,7 @@ func TestStore_BrokenSchema(t *testing.T) {
 		t.Fatal("expected no error, got", err.Error())
 	}
 	// Break
-	_, _ = store.db.Exec("DROP TABLE endpoint_result_conditions")
+	_, _ = store.db.Exec("drop table endpoint_result_conditions")
```
TwiN, line 457: Can you revert this change?

TwiN (changes requested):

storage/store/sql/sql_test.go
skhokhlov: Sorry, it was auto formatting in my IDE

skhokhlov:
@TwiN how does the process of PR work in this project? Who should resolve conversations, the one asking or the one answering? You don't have contributing guidelines in this repo so it's not clear.

BassSingh:
Hey lovely @TwiN is there any chance you could review this sir? It would be amazing if we could get this feature as we absolutely love Gatus!!! <3

heitorPB:
What exactly is an event in this context? What's the difference to a result?

emouawad:
Reviving.. is this still planned?

jonisapp:
Any update on this ? Gatus does so many things, so it's a bit disappointing not to be able to configure the data retention period...

TwiN:
Looks like there's a couple of merge conflicts due to how long it has been.

I've always been holding back about this, mostly because the way the information is displayed would provide a bad user experience.

Right now, you have to go one page at a time on the dashboard, which is 20 results each. If there were 1000 results, you'd have to click 50 times to get to the last page.

Either way, since this has been requested so many times, we may as well give it a shot.

P.S. due to how many code changes have been made, it may not be possible to easily resolve the conflicts in this PR

Issue #88, opened 2021-02-21
web.context-root is needed if the URL can not be rewritten.


chi-bd:
At a1679ddc5e6c36e58fcfc75ece86bbe557e27709, `web.context-root` is deleted.
But this option is needed in the environment which the URL can not be rewritten, for example Amazon ALB ingress controller.
Would you please revive it?

TwinProduction:
As mentioned in v2.0.0's release notes:
> Removed web.context-root (Vue.js' router doesn't easily support context roots. May be re-implemented in the future if enough people ask for it)

The main reason for removing it is due to my inability to make a dynamic context root work with Vue.js' router (without ugly hacks) despite several attempts. 

If you want to have a go at it, feel free, but unless a few more people request this, I won't personally try to re-implement it.

That being said, for your specific issue, most people create a separate CNAME (e.g. `status.example.com`) for their status page. 
Is there something preventing you from doing this?

KarlMW:
What about this?  https://stackoverflow.com/questions/43879418/deploy-vuejs-app-in-a-subdirectory
It seems that it can be done with an edit of `web/app/vue.config.js` like this:
```
module.exports = {
  ...
  publicPath: "./"
  ...
};
```
(bear in mind that I know nothing of what I write - I just googled the problem - hopefully not wasting your time here)

In terms of the usefulness of being able to host content in a sub-directory - it's one of the first things I look for in a service I want to privately self-host (ie. purely for my own use).  Being in a subdirectory means that its existence is not visible, whereas the existence of subdomains is publicly displayed in DNS.

KarlMW:
I hope that's not one of the "ugly" hacks you refer to  :-)

TwinProduction:
@KarlMW Hahaha no, if it was just that, it couldn't even be called a hack.

In fact, this would work fine if I weren't using vue router for the service detail page.

Long story short, one of `publicPath`'s limitations is that it cannot a relative path cannot be used in conjunction with `history.pushState`, which is unfortunately used by Vue.js' router. I could fix this by using hash history (e.g. `http://localhost:8080/#/services/core_twinnation-sitemap`), but I just can't bring myself to like that style of routing.

[reference](https://cli.vuejs.org/config/#publicpath):
> ### Limitations of relative publicPath
>
> Relative publicPath has some limitations and should be avoided when:
> - You are using HTML5 history.pushState routing;
> - You are using the pages option to build a multi-paged app.

[...]

PR #507, opened 2023-06-20
docs: Specify that exposing Gatus through a subpath is not supported

ansilh:
## Summary
Added FAQ to address issue https://github.com/TwiN/gatus/issues/88.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ Have any feedback or questions? [Create a discussion](https://github.com/TwiN/ga
   - [disable-monitoring-lock](#disable-monitoring-lock)
   - [Reloading configuration on the fly](#reloading-configuration-on-the-fly)
   - [Endpoint groups](#endpoint-groups)
+  - [Exposing Gatus on a custom path](#exposing-gatus-on-a-custom-path)
   - [Exposing Gatus on a custom port](#exposing-gatus-on-a-custom-port)
   - [Keeping your configuration small](#keeping-your-configuration-small)
   - [Badges](#badges)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1712,7 +1712,11 @@ variable instead, you can use that environment variable directly in the configur
 web:
   port: ${PORT}
 ```
+### Exposing Gatus UI on a custom URL.
 
+Currently, you can expose the Gatus UI using a custom fully qualified domain name (FQDN) such as `gatus.customdomain.com`. However, it does not support path-based routing, which means you cannot access it through a URL like `common.customdomain.com/gatus`.
```
TwiN, line 1717:
```suggestion
Currently, you can expose the Gatus UI using a custom fully qualified domain name (FQDN) such as `status.example.org`. However, it does not support path-based routing, which means you cannot access it through a URL like `example.org/status/`. 
```

README.md
```diff
@@ -1712,7 +1712,11 @@ variable instead, you can use that environment variable directly in the configur
 web:
   port: ${PORT}
 ```
+### Exposing Gatus UI on a custom URL.
 
+Currently, you can expose the Gatus UI using a custom fully qualified domain name (FQDN) such as `status.example.org`. However, it does not support path-based routing, which means you cannot access it through a URL like `example.org/status/`.
```
TwiN, line 1717:
```suggestion

### Exposing Gatus UI on a custom URL.
Currently, you can expose the Gatus UI using a fully qualified domain name (FQDN) such as `status.example.org`. However, it does not support path-based routing, which means you cannot expose it through a URL like `example.org/status/`. 
```

TwiN (approved):

PR #510, opened 2023-06-26
feat(config): Make RefreshInterval configurable

### LINKED ISSUE #509: Persist refresh interval on change

In the ui part of the YAML config I want to set the default refresh value of the frontend. 

I am to lazy to switch to the 10s refresh interval again and again, a preconfigured one would be nice.

_No response_

### Additional information

_No response_

macmoritz:
## Summary
This PR adds a config entry for the default value of the refresh interval of the ui.
Default is 300s/5m as before
Closes #509

## Checklist


config/ui/ui.go
```diff
@@ -22,12 +23,13 @@ var (
...
-	Logo        string   `yaml:"logo,omitempty"`        // Logo to display on the page
-	Link        string   `yaml:"link,omitempty"`        // Link to open when clicking on the logo
-	Buttons     []Button `yaml:"buttons,omitempty"`     // Buttons to display below the header
+	Title           string   `yaml:"title,omitempty"`           // Title of the page
+	Description     string   `yaml:"description,omitempty"`     // Meta description of the page
+	Header          string   `yaml:"header,omitempty"`          // Header is the text at the top of the page
+	Logo            string   `yaml:"logo,omitempty"`            // Logo to display on the page
+	Link            string   `yaml:"link,omitempty"`            // Link to open when clicking on the logo
+	Buttons         []Button `yaml:"buttons,omitempty"`         // Buttons to display below the header
+	RefreshInterval int      `yaml:"refreshInterval,omitempty"` // RefreshInterval is the interval in seconds at which the page gets refreshed
```
TwiN, line 32:
```suggestion
	RefreshInterval int      `yaml:"refresh-interval,omitempty"` // Interval in seconds at which the page gets refreshed
```
For consistency

README.md
```diff
@@ -240,6 +240,7 @@ If you want to test it locally, see [Docker](#docker).
 | `ui.header`                                     | Header at the top of the dashboard.                                                                                                         | `Health Status`            |
 | `ui.logo`                                       | URL to the logo to display.                                                                                                                 | `""`                       |
 | `ui.link`                                       | Link to open when the logo is clicked.                                                                                                      | `""`                       |
+| `ui.refreshInterval`                            | Default interval to refresh the page. Available values: 10, 30, 60, 120, 300, 600 (in seconds)                                              | `300`                      |
```
TwiN, line 243:
```suggestion
| `ui.refresh-interval`                            | Default interval to refresh the page. Available values: 10, 30, 60, 120, 300, 600 (in seconds)                                              | `300`                      |
```

In hindsight, perhaps this variable should be named `default-refresh-interval` instead 🤔

TwiN (changes requested):

README.md
macmoritz: Just changing the naming of the yaml key?
I would say so, otherwise we have a conflict in the go code with the default value.

macmoritz: Changed only the yaml key to `default-refresh-interval` cause of the conflicts mentioned above.

macmoritz:
@TwiN I changed the naming of the YAML key. Could you rereview and give me some feedback or merge?

config/ui/ui.go
```diff
@@ -22,12 +23,13 @@ var (
...
-	Logo        string   `yaml:"logo,omitempty"`        // Logo to display on the page
-	Link        string   `yaml:"link,omitempty"`        // Link to open when clicking on the logo
-	Buttons     []Button `yaml:"buttons,omitempty"`     // Buttons to display below the header
+	Title           string   `yaml:"title,omitempty"`                    // Title of the page
+	Description     string   `yaml:"description,omitempty"`              // Meta description of the page
+	Header          string   `yaml:"header,omitempty"`                   // Header is the text at the top of the page
+	Logo            string   `yaml:"logo,omitempty"`                     // Logo to display on the page
+	Link            string   `yaml:"link,omitempty"`                     // Link to open when clicking on the logo
+	Buttons         []Button `yaml:"buttons,omitempty"`                  // Buttons to display below the header
+	RefreshInterval int      `yaml:"default-refresh-interval,omitempty"` // Default Interval in seconds at which the page gets refreshed
```
TwiN, line 32: Can you rename `RefreshInterval` to `DefaultRefreshInterval` as well?

TwiN (changes requested):

config/ui/ui.go
macmoritz: I though about it, but there would be a naming conflict with the in-code default value:
```go
const (
    ...
    defaultRefreshInterval = 300
)
```
The only difference would be the _d_, this is confusing in my point of view.
Is there a better naming?

TwiN: variables/constants/types/methods/functions starting with a lowercase letter are not exported outside the package, so it's not too unusual imo.

The `default-refresh-interval` in the YAML is to make it more obvious for users that this value can be overriden, while the one from the constant is just for us. Honestly, you could remove the constant and set the value directly to 300 on L76 if you prefer.

macmoritz: Fixed

macmoritz:
Hey @TwiN, could you please check again and update me?

TwiN:
@macmoritz I know it's been a while since this has been open, but looking at the issue you created, why not just replace sessionStorage by localStorage instead?

https://github.com/TwiN/gatus/blob/bde30b2efb82982e6b77592cc0c690760906e8d1/web/app/src/components/Settings.vue#L36

https://github.com/TwiN/gatus/blob/bde30b2efb82982e6b77592cc0c690760906e8d1/web/app/src/components/Settings.vue#L81

I'm suggesting this, because it seems I accidentally used sessionStorage instead & that's why you have to keep re-setting the interval 😅 

Let me see if I can do something for you

TwiN:
I believe I've implemented a satisfactory fix for #509 - let me know what you think.

I do appreciate you attempting to fix it via a PR, though!

macmoritz:
This looks like a proper solution. Thanks for fixin.

TwiN:
@macmoritz Glad you're satisfied! Thanks again for contributing.

PR #514, opened 2023-06-30
fix(css): disable hover effects when not supported

macmoritz:
## Summary
Hover effects only should been applied when they are supported.
For example on smartphones hover gets triggered on touch, which is unhandy if you click on the response times.

## Checklist
- [x] Updated documentation in `README.md`, if applicable. => no need from my point of view

### DIFF
--- a/web/app/tailwind.config.js
+++ b/web/app/tailwind.config.js
@@ -14,4 +14,7 @@ module.exports = {
     extend: {},
   },
   plugins: [],
+  future: {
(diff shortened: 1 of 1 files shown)


TwiN:
I'm fine with merging this, but since you haven't regenerated the static assets, the change won't be there until the next time there's a change on the UI.

macmoritz:
Makes sense. I need to run `make frontend-build`, correct?

TwiN (approved):

PR #522, opened 2023-07-10
chore: use strings.EqualFold instead

testwill:
## Checklist


security/oidc.go
```diff
@@ -117,7 +117,7 @@ func (c *OIDCConfig) callbackHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	for _, subject := range c.AllowedSubjects {
-		if strings.ToLower(subject) == strings.ToLower(idToken.Subject) {
+		if strings.EqualFold(subject, idToken.Subject) {
```
TwiN, line 120: Was there a specific reason why this change was needed? Did you encounter an issue with a specific subject?

Issue #533, opened 2023-07-24
Add Schema for config file


amai2012:
### Describe the feature request

A JSON schema could be used to validate the config.yml for gatus.

### Why do you personally want this feature to be implemented?

Aid in writing my own configuration files (see https://github.com/TwiN/gatus/issues/57)

Currently my pipeline deploying gatus has a just a yamllint step, which could be replaced by validating against the schema.

### How long have you been using this project?

10 days

amai2012:
Also it would be convenient to upload it to https://www.schemastore.org/json/ so users can comfortably use it in their IDE.

PR #534, opened 2023-07-24
Create gatus-schema.json

amai2012:
Provide a JSON schema for the config.yaml file

## Summary
#533

## Checklist


TwiN:
I would rather have something automated that generates the schema instead of having to update it manually every time the schema changes

amai2012:
Ok, I am not familiar with the Implementation.
Any ideas on how to generate it?

TwiN:
First step would be finding a programmatic way to generate a schema using something like reflection on the primary `Config` struct https://github.com/TwiN/gatus/blob/7d570ce148010e2cea4d9748a3474f4b2c15509a/config/config.go#L54
That file would have a main function triggering that schema generation and would be somewhere like `scripts/generate-schema.go`.

Once you find a way to do this, you'll have to create a new GitHub workflow to call `scripts/generate-schema.go` and then commit the file with the generated schema to the PR that triggered the workflow. If you're not familiar with GHA, you can skip the GHA part for now and just focus on generating the schema programmatically.

PR #546, opened 2023-08-07
fix(client): Honor client.insecure when doing TLS checks

### LINKED ISSUE #541: client.insecure Does Not Affect TLS/STARTTLS Checks
### Describe the bug

I am trying to monitor TLS endpoints that use certificates signed by an internal CA.  For the purposes of this monitor, TLS validation is not strictly necessary.  For HTTPS endpoints, setting `client.insecure` to `true` will prevent certificate validation errors from causing a failure response.  This doesn't seem to be honored for TLS/STARTTLS checks.

In the interim, I am just using TCP checks which are not able to provide additional useful information such as [CERTIFICATE_EXPIRATION]

### What do you see?

When doing a TLS or STARTTLS check, the `client.insecure` option does not seem to be honored and the checks fail if the cert is unable to be validated for any reason (no IP SAN, not trusted, etc)

When setting `client.insecure` to `true` for an endpoint, TLS validation for TLS/STARTTLS checks should be disabled to allow the check to complete.

1. Configure an endpoint that uses a TLS check
2. Either use a target that has a self-signed cert or replace the hostname with its IP address (assuming the cert does not have an IP SAN)
3. Set `client.insecure` to `true` for that endpoint

The check will fail for one or more reasons related to TLS validation.

### Version

5.5.1

### Additional information

_No response_

vax-r:
## Summary
This PR enables gatus to honor client.insecure in the TLS checking
it will detect client.insecure in config.yaml and perform the relevant operation like ignore bad certificate and unknown certificates authority and so on.

Fixes #541 

## Checklist

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -142,14 +142,20 @@ func CanPerformStartTLS(address string, config *Config) (connected bool, certifi
 
 // CanPerformTLS checks whether a connection can be established to an address using the TLS protocol
 func CanPerformTLS(address string, config *Config) (connected bool, certificate *x509.Certificate, err error) {
-	connection, err := tls.DialWithDialer(&net.Dialer{Timeout: config.Timeout}, "tcp", address, nil)
+	connection, err := tls.DialWithDialer(&net.Dialer{Timeout: config.Timeout}, "tcp", address, &tls.Config{
+		InsecureSkipVerify: config.Insecure,
+	})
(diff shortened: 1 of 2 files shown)


vax-r:
I forgot to remove the testing changes in config.yaml
I'll remove it later, sorry.

PR #551, opened 2023-08-15
fix(alerting): correct success threshold for telegram

plantatorbob:
## Summary

There is a typo in Telegram alerting provider implementation: for resolved alerts, instead of success count, failure count is printed instead.

This PR fixes the typo and adjusts relevant test cases.

## Checklist


TwiN (approved):
Good catch!

TwiN:

PR #552, opened 2023-08-18
feat(endpoint): shorten interval on error

### LINKED ISSUE #112: Shorten interval on error
Feature emrequest: It would be great to be able to decrease interval time on error. E.g.: check every hour if everything is ok but when an error happens start checking every ten minutes.

vax-r:
## Summary

This PR enables user to set a shorter interval when an endpoint status check isn't success
_Note that you have to add a field called failed-interval in endpoints under config.yaml_

fix #112 

@TwiN , I haven't write the document yet, after your review I'll write this feature into README.md and alter the example in config.yaml

## Checklist


core/endpoint.go
```diff
@@ -104,6 +104,12 @@ type Endpoint struct {
 	// Interval is the duration to wait between every status check
 	Interval time.Duration `yaml:"interval,omitempty"`
 
+	// FailedInterval is the duration we'll use to replace Interval when an endpoint stauts check is failed
+	FailedInterval time.Duration `yaml:"failed-interval,omitempty"`
+
+	// IntervalChanged is whether the endpoint's Interval has changed before
+	IntervalChanged bool `"omitempty"`
```
heitorPB, line 111: This bool doesn't seem to be useful? Is starts as `false` and is set to `true` if there's an error. Maybe i'm missing something?

watchdog/watchdog.go
```diff
@@ -90,6 +90,12 @@ func execute(endpoint *core.Endpoint, alertingConfig *alerting.Config, maintenan
 
 // UpdateEndpointStatuses updates the slice of endpoint statuses
 func UpdateEndpointStatuses(endpoint *core.Endpoint, result *core.Result) {
+	// If user has set the failed-interval field, update enpoint interval due to its previous failure
+	if !endpoint.IntervalChanged && !result.Success && endpoint.FailedInterval.Seconds() > 0 {
+		endpoint.Interval = endpoint.FailedInterval
```
heitorPB, line 95: doesn't this value need to be changed back to the original value when the check is successful again?

vax-r: @heitorPB I am not sure about this part, I thought the issue owner want it to be shorten forever?

core/endpoint.go
vax-r: @heitorPB I'll make sure what is it for and fix it later thank you

watchdog/watchdog.go
heitorPB: > @heitorPB I am not sure about this part, I thought the issue owner want it to be shorten forever?

If someone wants to shorten the interval between checks forever, why change the code instead of the configuration file?

What I understood about this feature request is to have shorter intervals (more checks) _during an outage_, and after the system is healthy again go back to the regular intervals.

Does that make sense?

TwiN:
So there's a reason why I haven't implemented this sooner. My biggest concern is, if we have a retry interval (or failed interval, as you named it), its interaction with the alert configuration is no longer implicitly obvious/evident, if that makes sense.

For instance, let's say you set the normal interval to 3m, and the retry interval to 5 seconds. 
This would imply that once every 3 minutes, the endpoint's health is evaluated, and if it fails, it will retry 5 seconds later. If that fails again, then it would presumably stop retrying and the next check would be in 3 minutes.

The problems with the statement above are the following:
- I made an assumption that there would only be one retry. This isn't exactly clear, and that is reinforced by the fact that your implementation seems to retry forever until it succeeds (but what if it doesn't? The increase in load would likely not help the situation, especially if the underlying endpoint being retried upon performs intensive tasks behind the scenes)
- What about the alerts?
  - Does the `failure-threshold` include the retries?
    - One could assume that it should, as it may allow us to trigger an alert sooner which users could then fix sooner, but what about the `success-threshold` then? Would it make sense for alerts to be resolved faster too?
- How should this be displayed on the UI? Should successful retries show the failed retries? If multiple retries fail, wouldn't that not pollute the UI by showing multiple failures in a row when these failures in fact happened in the shorter time frame defined by the retry interval?

What I'm trying to get at is, I'm really not sure I want to add this because it makes the configuration less intuitive.

TwiN:
As there has been no reply for several months, I'm going to close this.

PR #563, opened 2023-09-06
fix: Support hexadecimal integers in conditions

heitorPB:
Remove the assumption that JSON contains only decimal integers and instead identify the base from the prefix of the data.

[`strconv.ParseInt`](https://pkg.go.dev/strconv#ParseInt) can identify the base of the integer based on its "prefix":
- 0x -> hexadecimal
- 0o -> octal
- 0b -> binary
- decimal otherwise.

## Checklist


core/condition_test.go
```diff
@@ -287,6 +287,27 @@ func TestCondition_evaluate(t *testing.T) {
 			ExpectedSuccess: true,
 			ExpectedOutput:  "[BODY].data.id > 0",
 		},
+		{
+			Name:            "body-jsonpath-hexadecimal-int-using-greater-than",
+			Condition:       Condition("[BODY].data > 0"),
+			Result:          &Result{Body: []byte("{\"data\": \"0x1\"}")},
+			ExpectedSuccess: true,
+			ExpectedOutput:  "[BODY].data > 0",
+		},
```
TwiN, line 296: Could you add one more test that makes sure `len([BODY].data) == 3` passes? 

I'm fine with the change, but I'd like to ensure that the length of a string, even if that string is a hexadecimal representation of a number, is treated like a string.

TwiN (changes requested):

core/condition_test.go
heitorPB: Sure. Can I have multiple conditions in these test scenarios, like in the YAML? Or do you prefer new test case? I can add more tests as well, any ideas of what could go wrong here? Maybe longer numbers?

TwiN: Up to you; I'd personally make a separate test case just for `len([BODY].data) == 3`.
No need for testing longer numbers imo.

heitorPB: Done! Let me know what you think about it.

TwiN: It's perfect

core/condition_test.go
```diff
@@ -287,6 +287,48 @@ func TestCondition_evaluate(t *testing.T) {
 			ExpectedSuccess: true,
 			ExpectedOutput:  "[BODY].data.id > 0",
 		},
+		{
+			Name:            "body-jsonpath-hexadecimal-int-using-greater-than",
+			Condition:       Condition("[BODY].data > 0"),
+			Result:          &Result{Body: []byte("{\"data\": \"0x1\"}")},
```
TwiN, line 293: In hindsight, shouldn't the data here be `0x1` instead of `"0x1"`?

Could you add a test with the same body, but the condition `[BODY].data == 0x1`?

TwiN (changes requested):

core/condition_test.go
heitorPB: I tried that, than it crashed.

JSON does not support hex numbers directly, we need to wrap them as strings. See this discussion: https://stackoverflow.com/questions/52671719/can-hex-format-be-used-with-json-files-if-so-how

We can only have decimal integers in a JSON element.

core/condition_test.go
heitorPB: Found an issue. I added a test to compare `[BODY].data == 2` with `{\"data\": \"0x02\"}` and I got this error:

```
--- FAIL: TestCondition_evaluate (0.00s)
    --- FAIL: TestCondition_evaluate/body-jsonpath-hexadecimal-int-using-greater-than-without-strings (0.00s)
        condition_test.go:683: Condition '[BODY].dataa == 2' should have been success=true
        condition_test.go:686: Condition '[BODY].dataa == 2' should have resolved to '[BODY].dataa == 2', got '[BODY].dataa (INVALID) == 2'
```

For some reason, `sanitazeAndResolve` (`core/endpoint.go:232`) fails to extract the hex string from the JSON. I suspect something on the function `walk` in `jsonpath/jsonpath.go` is not right.

heitorPB: Oh, actually it was an error on my test case... ignore the previous errors. THe issue is on the isEqual function. It tries to test if `"0x02"` is equal to `"2"`, of course it isn't:

```
--- FAIL: TestCondition_evaluate (0.00s)
    --- FAIL: TestCondition_evaluate/body-jsonpath-hexadecimal-int-using-greater-than-without-strings (0.00s)
        condition_test.go:683: Condition '[BODY].dataa == 2' should have been success=true
        condition_test.go:686: Condition '[BODY].dataa == 2' should have resolved to '[BODY].dataa == 2', got '[BODY].dataa (0x02) == 2'
```

heitorPB: Oh, found the issue. Please check my 2 most recent commits. Result of the test suite did not change for me.

core/condition_test.go
TwiN: That's alright. Could you still create a test to see if the condition `[BODY].data == 0x1` would work with the the body `[]byte("{\"data\": \"0x1\"}")`?

heitorPB: It works!

I also added other tests of equality as well. Let me know what you think.

TwiN: I think you're missing my point. I'd like to make sure if the condition `[BODY].data == 0x1` passes if the body is `{"data": "0x1"}`.

All the tests you made have conditions that are numerical while the body is non-numerical (e.g. hexadecimal), but there are no tests with, for instance, an hexadecimal condition.

The test should look something like this:
```go
{
	Name:            "...",
	Condition:       Condition("[BODY].data == 0x1"),
	Result:          &Result{Body: []byte("{\"data\": \"0x1\"}")},
	ExpectedSuccess: true,
	ExpectedOutput:  "[BODY].data == 0x1",
},
```

heitorPB: Btw, I added a bunch of tests, if you want I can remove some of them. Let me know what you think!

TwiN (approved):

core/condition_test.go
TwiN: Awesome, thank you!

TwiN:

heitorPB:
Thank you for Gatus and mentoring me in this issue :)

Issue #75, opened 2021-01-20
Config in env variables is weak


pollosp:
First of all nice work with Gatus folks 👏 , thank for the effort here and doing it opensource.

When I was setting up Gatus and I missed the availability of write some config as ENV variable , I thought it could be nice  specially for the alerting configuration with secrets tokens and private urls. 
Using environment variables way maybe you can use the config yml file as template or base for multiple environments and set those settings independently of the environment. 

I'm thinking in a scenario where some config is populated to a container using ENV variables like with ECS + [Chamber](https://github.com/segmentio/chamber) or ECS + [ssm-env](https://github.com/remind101/ssm-env) overriding or complementing the yml file. 

For example you are doing it with `GATUS_CONFIG_FILE` which is very useful if you maintain and build those yml files , per environment

Is something like this in your roadmap? Do you think this could be useful?
Thanks again.

TwinProduction:
@pollosp In fact, you can use environment variables in your configuration.

For instance:
```yaml
services:
  - name: twinnation
    url: "http://${ENVIRONMENT}.example.org/"
    interval: 30s
    conditions:
      - "[STATUS] == 200"
```

If you want to use environment variables in your alerting configuration, you could do the following:
```yaml
alerting:
  slack: 
    webhook-url: "${SLACK_WEBHOOK_URL}"
```

pollosp:
@TwinProduction Oh I didn't know it I was checking the code but I didn't infer that 👏  I will try and I'm going to close the issue!

PR #571, opened 2023-09-20
docs: demonstrate use of environment variables in config

dan-hill2802:
## Summary
The example for Postgres shows the database credentials in plain text in `config.yaml`, which isn't an ideal way to provide credentials. I didn't initaly realise environment variables could be used in config files until I saw #75 . I later noticed the note in the README. Hopefully this makes it clearer and provides a better example for configuring database connections.

## Checklist

### DIFF
--- a/.examples/docker-compose-postgres-storage/docker-compose.yml
+++ b/.examples/docker-compose-postgres-storage/docker-compose.yml
@@ -18,6 +18,10 @@ services:
     restart: always
     ports:
       - "8080:8080"
+    environment:
+      - POSTGRES_DB=gatus
+      - POSTGRES_USER=username
+      - POSTGRES_PASSWORD=password
     volumes:
       - ./config:/config
     networks:
--- a/.examples/docker-compose-postgres-storage/docker-compose.yml
+++ b/.examples/docker-compose-postgres-storage/docker-compose.yml
@@ -19,9 +19,9 @@ services:
     ports:
       - "8080:8080"
     environment:
-      - POSTGRES_DB=gatus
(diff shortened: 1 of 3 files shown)


.examples/docker-compose-postgres-storage/docker-compose.yml
```diff
@@ -18,6 +18,10 @@ services:
     restart: always
     ports:
       - "8080:8080"
+    environment:
+      - POSTGRES_DB=gatus
+      - POSTGRES_USER=username
+      - POSTGRES_PASSWORD=password
```
TwiN, line 24:
```suggestion
      - POSTGRES_USER=username
      - POSTGRES_PASSWORD=password
      - POSTGRES_DB=gatus
```

TwiN: Nitpicking, but just to match the order of the env vars as they're used in `postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@postgres:5432/${POSTGRES_DB}?sslmode=disable`

dan-hill2802: I matched the order from the postgres container in this file. Do you want postgres changing too or leaving as it is?

TwiN: No need, just changing the order of the env vars as I suggested will be enough.
I'd have committed it directly in your PR, but I think you disabled the option allowing contributors to contribute to your PR/fork

TwiN (approved):

PR #582, opened 2023-09-30
fix(alerting): add condition results to ntfy

wei840222:
show the condition results in ntfy notifications like below.  
image
image


TwiN (approved):

TwiN:

Issue #554, opened 2023-08-23
SQL Server monitoring support


vanhoutenbos:
### Describe the feature request

Hi!

It would be awesome to be able to monitor my SQL Server to see if my database is reachable and 'works'.
This could be using the ODBC protocol for example.
A bit more extensive but Datadog uses this method to collect all their data, in this case I would only need to know if the database 'is alive'.

https://docs.datadoghq.com/database_monitoring/setup_sql_server/

### Why do you personally want this feature to be implemented?

It would be nice to check if the database is alive and is working.

TwiN:
Not a big fan of proprietary systems. If you can get the ODBC protocol working for mysql/postgres/etc and it happens to also support SQL server, I don't mind the feature, but I don't want to add a feature that only works for a proprietary system.

vanhoutenbos:
Understandable, I would be more a fan of a protocol that supports many vendors not just one.

The reason for requesting is a lot of my clients use a variant of Microsoft SQL Server and want to monitor it using a 'uptime' or connectivity tool.
I almost always recommend Gatus!
This is a missing feature for some of them that makes it hard to check connectivity to the database.

Issue #437, opened 2023-02-17
Support service agents


epoberezkin:
### Describe the feature request

Currently the service itself should be queried by Gatus in order to determine its response time / availability.

There are more complex services that require a standalone "agent" application to determine service availability from "business logic" point of view – e.g., it can be determined as a sequence of service requests rather than just one request, and the service "delay" is the time of transaction and not the time of one request.

The proposal is NOT to complicate the logic of Gatus to allow all possible scenarios, but rather add support for endpoint agents when:
- the server state is displayed based on the content of the response from the agent, and not the response time.
- the time of agent response is not taken into account, the lack of agent response is treated as "no information" (grey color) rather than as "service down"

### Why do you personally want this feature to be implemented?

We are looking for an open-source solution to present the status of preset servers in https://simplex.chat apps to the end users - this can only be done via agent, as we need to test that the message is actually delivered and show e2e delivery time, rather than just that the server responds to PING or anything else.

### How long have you been using this project?

evaluating

### Additional information

We can sponsor this feature - not to cover all costs though, but we would sponsor say $100.

epoberezkin:
alternatively it could be status update API I guess to achieve the same...

TwiN:
Because of the fact that users could be monitoring pretty much anything, I don't think creating agents for each individual use case would be appropriate. Instead, as you've pointed out, something like #229 would make more sense, as even if we decide to create agents in the future, #229 would still likely be necessary.

TwiN:
This will become possible once #724 is merged

TwiN:
#724 has already been merged some time ago. Closing.

PR #596, opened 2023-10-18
poc: custom scripts

jcsaaddupuy:
Hi,

Using a custom script in a endpoint would allow anyone to support any custom target.

This MR is a poc to execute a custom script, and be able to check in conditions the exit code, the script stdout/stderr.

Example configuration : 
```yaml
  - name: custom script
    script:
      command: "echo -n HELLO"
    interval: 10s
    conditions:
      - "[EXIT_CODE] == 0"
      - "[STDOUT] == HELLO"
```

I did not implement tests yet, and the MR may not be clean.
Let me know what you think !

### DIFF
--- a/core/condition.go
+++ b/core/condition.go
@@ -50,6 +50,10 @@ const (
 
 	// DomainExpirationPlaceholder is a placeholder for the duration before the domain expires, in milliseconds.
 	DomainExpirationPlaceholder = "[DOMAIN_EXPIRATION]"
+
+	ScriptExitCode = "[EXIT_CODE]"
+	ScriptStdOut   = "[STDOUT]"
+	ScriptStdErr   = "[STDERR]"
 )
 
 // Functions
@@ -261,6 +265,12 @@ func sanitizeAndResolve(elements []string, result *Result) ([]string, []string) 
 			element = strconv.FormatInt(result.CertificateExpiration.Milliseconds(), 10)
 		case DomainExpirationPlaceholder:
 			element = strconv.FormatInt(result.DomainExpiration.Milliseconds(), 10)
+		case ScriptExitCode:
+			element = strconv.FormatInt(int64(result.ScriptExitCode), 10)
+		case ScriptStdErr:
+			element = string(result.ScriptStderr)
+		case ScriptStdOut:
+			element = string(result.ScriptStdout)
 		default:
 			// if contains the BodyPlaceholder, then evaluate json path
 			if strings.Contains(element, BodyPlaceholder) {
--- a/core/endpoint.go
+++ b/core/endpoint.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os/exec"
 	"strings"
 	"time"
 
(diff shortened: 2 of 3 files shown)


TwiN:
Gatus' Docker image doesn't have bash installed though, nor does it have curl, wget, echo, etc.

jcsaaddupuy:
Indeed.
It's a good thing to keep the default image as light as possible.

But isn't it only a runtime issue?

As a gatus user, it's totally possible to either extend the base image, or build a docker image with a more complex setup and install gatus in it.
This feature can stay totally optional.

Supporting custom script would be a simple way to let end-users implement usecases that are not covered out of the box by gatus.

For example, in my case, I need to check the correct execution of some pipelines persisting data in a bigquery database.
With this feature, I'm able to implement a SQL check based on a python script, delegating all the setup/authentication mechanism outside of gatus.

That could be extended to others cases (for ex: having checks based on complex api calls, proprietary systems/...) while keeping the impact on gatus codebase low - we just check for the exit code after all :)

jcsaaddupuy:
To add more material on this proposal, one could build a custom image but still using gatus from the last official image :

```Dockerfile
FROM busybox # or any other base image, for ex python

COPY --from=twinproduction/gatus /gatus /gatus

COPY ./config /config # inject custom local config defining scripts endpoints
# also should copy custom scripts in PATH

ENV PORT=8080
EXPOSE ${PORT}
ENTRYPOINT ["/gatus"]
```

I see at least one or two issues (in addition to my own usecase) this would allow to easily support :
- https://github.com/TwiN/gatus/issues/554
- https://github.com/TwiN/gatus/issues/437

Let me know if you find interest in this feature, I can make a proper PR with unittests / clean a little the code.

TwiN:
I'm conflicted, because this feature may cause people to be unhappy if it doesn't work out of the box, which it won't unless people create their own Docker image -- something not everybody knows how to do or is willing to spend the time doing.

Instead, I feel like the new [external endpoint](https://github.com/TwiN/gatus?tab=readme-ov-file#external-endpoints) feature may be sufficient to achieve the same goal.

kadaan:
The external endpoint feature is nice, but it does require developing additional services to support those custom checks.  If, for example, I want to support MQTT checks I need to write a service and host it in kubernetes so that it call call Gatus.  This is much more complex than extending the Gatus image with the tooling necessary to run the additional check you want to support.

kadaan:
Another advantage is that the community could author non-core checks (i.e. GPRC) that would be able to be used without integrating into the core of Gatus.  This is beneficial because there will always be cases where people want support for specific systems that will not have enough adoption to justify the support cost of adding them to Gatus.

kadaan:
@jcsaaddupuy Another use case would be MQTT support.

kadaan:
@TwiN I know that this is retired, but something like this would get us custom check support but allow it to be external: https://github.com/kubernetes-retired/contrib/blob/master/exec-healthz/cmd/exechealthz/exechealthz.go

TwiN:
This is no longer required, as you can now use external endpoints to push statuses.

You can run any custom scripts you want on any system you want with access to tools, and then push them.

Issue #607, opened 2023-11-02
feat: make authentication optional for email alert provider


calvinhenderson:
### Describe the feature request

Make authentication optional for the email alert provider.

### Why do you personally want this feature to be implemented?

Currently, the email alert provider requires a password and always performs authentication. This is incompatible with an environment utilizing IP-based authentication without a username/password for internal messages.

For example, a [Google Workspace](https://support.google.com/a/answer/2956491) may not enforce authentication for specific IPs utilizing the Gmail SMTP relay:
<img width="594" alt="image" src="https://github.com/TwiN/gatus/assets/23652636/58e332e8-c491-41b2-a1a3-d7695d8b1f68">

### How long have you been using this project?

< 1 week

### Additional information

An alternative to the opt-out flag is to disable authentication automatically when a username and password are not present. I figured a _hard opt-out_ makes more sense as this situation is far less common and automatically disabling authentication could make debugging more difficult for the average user.

PR #608, opened 2023-11-02
feat(alerting): make authentication optional for email provider

calvinhenderson:
## Summary

#607 

## Checklist


calvinhenderson:
I noticed an issue where emails are rejected when using a non-local smtp server due to the _from hostname_ being set to `localhost`. I updated the `gomail.Dialer` initialization to set the `LocalName` attribute to the `From` domain when disabling auth. This resolved the issue and email alerts are now working correctly utilizing the Gmail SMTP relay.

README.md
```diff
@@ -468,20 +468,21 @@ endpoints:
...
+| `alerting.email.from`                   | Email used to send the alert                                                               | Required `""` |
+| `alerting.email.username`               | Username of the SMTP server used to send the alert. If empty, uses `alerting.email.from`.  | `""`          |
+| `alerting.email.password`               | Password of the SMTP server used to send the alert                                         | Required `""` |
+| `alerting.email.host`                   | Host of the mail server (e.g. `smtp.gmail.com`)                                            | Required `""` |
+| `alerting.email.port`                   | Port the mail server is listening to (e.g. `587`)                                          | Required `0`  |
+| `alerting.email.to`                     | Email(s) to send the alerts to                                                             | Required `""` |
+| `alerting.email.disable-authentictaion` | Disables authentication. Useful for SMTP relays using IP-based authentication              | `false`       |
```
TwiN, line 480:
```suggestion
| `alerting.email.disable-authentication` | Disables authentication. Useful for SMTP relays using IP-based authentication              | `false`       |
```

TwiN: Is this really necessary though? Could we not just check if `alerting.email.password` is blank, and if it is, assume that the authn is disabled?

TwiN (changes requested):

README.md
calvinhenderson: Yep, I suggested that as an alternative in #607. I just wasn’t sure if that should be the default behavior. If you’re okay with that being the case then I can change it!

alerting/provider/email/email.go
```diff
@@ -66,7 +66,22 @@ func (provider *AlertProvider) Send(endpoint *core.Endpoint, alert *alert.Alert,
...
+
+	var d *gomail.Dialer
+	if len(provider.Password) == 0 {
+		// Get the domain in the From address
+		localName := "localhost"
+		fromParts := strings.Split(provider.From, `@`)
+		if len(fromParts) == 2 {
+			localName = fromParts[1]
+		}
+
```
TwiN, line 78:
```suggestion
	var d *gomail.Dialer
	if len(provider.Password) == 0 {
		// Get the domain in the From address
		localName := "localhost"
		fromParts := strings.Split(provider.From, `@`)
		if len(fromParts) == 2 {
			localName = fromParts[1]
		}
```

alerting/provider/email/email.go
```diff
@@ -14,12 +14,12 @@ import (
...
-	Password string `yaml:"password"`
-	Host     string `yaml:"host"`
-	Port     int    `yaml:"port"`
-	To       string `yaml:"to"`
+	From        string `yaml:"from"`
+	Username    string `yaml:"username"`
+	Password    string `yaml:"password"`
+	Host        string `yaml:"host"`
+	Port        int    `yaml:"port"`
+	To          string `yaml:"to"`
```
TwiN, line 22:
```suggestion
	From     string `yaml:"from"`
	Username string `yaml:"username"`
	Password string `yaml:"password"`
	Host     string `yaml:"host"`
	Port     int    `yaml:"port"`
	To       string `yaml:"to"`
```

TwiN:
Excellent work @calvinhenderson, thank you for contributing!

PR #631, opened 2023-11-28
feat: add ability to delay startup

### LINKED ISSUE #619: feat: Add ability to delay startup

I would like for there to be the ability to add a delay to the image starting up before executing /gatus.

I imagine this can be done internally based on an environment variable (e.g. GATUS_SLEEP). What might be easier is to modify the dockerfile to build from golang:alpine vs scratch. This will allow the use of sh or /bin/ash so that in Kubernetes, for example, we can overwrite the entrypoint be be something like `/bin/ash -c sleep 5 /gatus`.

The reason why I am looking into this is because when using a sidecar in Kubernetes (for example cloud-sql-proxy), the gatus image comes up before the sidecar does and that causes an error since the database is not available yet. A simple way to solve this is to add a configurable delay to the gatus image to give the sidecar enough time to be ready for connections.

vax-r:
## Summary

* Enable ability to set delay seconds before start up gatus
* ~~Currently the default example value is set to 5 seconds~~
* ~~enable configuration in Dockerfile and Makefile to set delay seconds before startup~~

fix #619 

## Checklist


main.go
```diff
@@ -14,6 +15,15 @@ import (
...
 func main() {
+	delay_sec, err := strconv.Atoi(os.Getenv("GATUS_DELAY_START_SECONDS"))
+	if err != nil {
+		// If we can't get GATUS_DELAY_START_SECONDS from environment variable
+		// set delay_sec to 0
+		log.Println("Error when parsing the value of GATUS_DELAY_START_SECONDS")
+		delay_sec = 0
+	}
+	// Wait for delay_sec
+	time.Sleep(time.Duration(delay_sec) * time.Second)
```
TwiN, line 26:
```suggestion
	if delayInSeconds, _ := strconv.Atoi(os.Getenv("GATUS_DELAY_START_SECONDS")); delayInSeconds > 0 {
		log.Printf("Delaying start by %d seconds", delayInSeconds)
		time.Sleep(time.Duration(delayInSeconds) * time.Second)
	}
```

TwiN (changes requested):

main.go
vax-r: @TwiN 
I've made the requested changes, and add an example under .example/ directory
Please take a look when you have time, thank you!

.examples/delay-startup/Makefile
```diff
@@ -0,0 +1,43 @@
...
+
+#############
+# Front end #
+#############
+
+frontend-build:
+	npm --prefix web/app run build
+
+frontend-run:
+	npm --prefix web/app run serve
```
TwiN, line 43: By example, I meant more like a Dockerfile similar to https://github.com/TwiN/gatus/tree/master/.examples/docker-minimal but with the environment variable you just added.

That being said, now that I've given it more thoughts, perhaps it would be easier to just add a section in the README under FAQ with a title like `Configuring a startup delay` that mentions the new `GATUS_DELAY_START_SECONDS` environment variable.

TwiN (changes requested):

.examples/delay-startup/Makefile
TwiN: Done in 636b570 and 7b224ca

TwiN (approved):

Issue #506, opened 2023-06-19
feat: support ping size + do not fragment


antoinekh:
### Describe the feature request

Hello,

Since probing support ping  size and do not fragment (dnf) it would be a nice to have these features also in gatus.

code in probing:
```
type Pinger struct {
        ...
        // Size of packet being sent
	Size int
       ...
	// df when true sets the do-not-fragment bit in the outer IP or IPv6 header
	df bool

```
something like in Gatus:
  - name: MY_SERVICE
    group: INTERNAL_CAAS_PRD
    url: "icmp://10.157.76.1"
    interval: 1s
    size:  2000
    dnf: true
    conditions:
      - "[CONNECTED] == true"
    client:
      timeout: 1s

### Why do you personally want this feature to be implemented?

It would be possible to test the ping with a load and validate the MTU for the network flow.

DNF is supported by probing since
https://github.com/prometheus-community/pro-bing/commit/235b493eb3b80aa5ecce268fc3b7b18538c7a361

### How long have you been using this project?

month

### Additional information

low priority, a nice to have

Could again be a edge use case and useless for most people, I'd understand if this wasn't Gatus's target.

PR #633, opened 2023-12-04
feat(icmp) support DoNotFragment + Size

antoinekh:
## Summary
It's really a network-related need.
We use Gatus for temporary tests and we need to do MTU validation.

This pull request adds 2 capabilities to ICMP:
- Ability to set the DoNotFragement bit
- Ability to configure the size

This will my issue #506 

Example of configuration with these options:
```
  - name: 10.35.206.132_df
    group: static
    url: "icmp://10.35.206.132"
    interval: 3s
    conditions:
      - "[CONNECTED] == true"
    client:
      timeout: 1s
      df: true
      size: 3500
```

## Checklist

- [X ] Tested and/or added tests to validate that the changes work as intended, if applicable.
- [X ] Updated documentation in `README.md`, if applicable.


README.md
```diff
@@ -38,6 +38,7 @@ Have any feedback or questions? [Create a discussion](https://github.com/TwiN/ga
 
 
 ## Table of Contents
+- [Table of Contents](#table-of-contents)
```
TwiN, line 41:
```suggestion
```

README.md
```diff
@@ -202,65 +203,65 @@ If you want to test it locally, see [Docker](#docker).
...
+| `ui.buttons[].link`                             | Link to open when the button is clicked.                                                                                                    | Required `""`              |
+| `maintenance`                                   | [Maintenance configuration](#maintenance).                                                                                                  | `{}`                       |
 
 
 ### Conditions
 Here are some examples of conditions you can use:
 
 | Condition                        | Description                                         | Passing values             | Failing values   |
-|:---------------------------------|:----------------------------------------------------|:---------------------------|------------------|
+| :------------------------------- | :-------------------------------------------------- | :------------------------- | ---------------- |
```
TwiN, line 264: Could you revert this & other similar changes?

client/config.go
```diff
@@ -50,6 +54,12 @@ type Config struct {
 	// Timeout for the client
 	Timeout time.Duration `yaml:"timeout"`
 
+	// Df set for the client
+	Df bool `yaml:"df"`
```
TwiN, line 58: What does `Df` mean? `DoFragment`?
Could you clarify that in the comment?

TwiN (changes requested):

antoinekh:
Ah that's my markdown auto format in my vscode, I'll revert.
And add info on the Df bit
I’ll take explanation for this good https://stackoverflow.com/questions/45953716/when-to-set-dont-fragment-flag-in-ip-header

antoinekh:
Should be better now.
Let me know if you need more informations.

I'm not sure if this is really a feature that should be added to Gatus, but it doesn't cost much to add it.

q:q!
TwiN, line 1: looks like this file has been added by accident?

TwiN (changes requested):

q:q!
antoinekh: My bad, fixed !

README.md
```diff
@@ -351,6 +350,8 @@ the client used to send the request.
 | `client.insecure`                      | Whether to skip verifying the server's certificate chain and host name.     | `false`         |
 | `client.ignore-redirect`               | Whether to ignore redirects (true) or follow them (false, default).         | `false`         |
 | `client.timeout`                       | Duration before timing out.                                                 | `10s`           |
+| `client.df`                            | Option to set the DoNotFragement bit for ICMP packet                        | `false`         |
+| `client.size`                          | Size of the ICMP packet                                                     | `25`            |
```
TwiN, line 354: I'In hindsight, maybe this should be under `icmp`, i.e. 

```
client:
  icmp:
    df: true
    size: 25
```

What do you think?

antoinekh: It's done, as you said, it makes more sense that way

client/config.go
```diff
@@ -73,6 +82,14 @@ type DNSResolverConfig struct {
 	Port     string
 }
 
+// ICMP is the configuration for the ICMP client specific config
+type Icmp struct {
+	// Don't Fragment flag (DF) for the client
+	Df bool `yaml:"df"`
+	// Size of the packet
+	Size int `yaml:"size"`
+}
```
TwiN, line 91:
```suggestion
type ICMPConfig struct {
	DF bool `yaml:"df"`    // Don't Fragment flag used for the ICMP client
	Size int `yaml:"size"` // Size of the packet
}
```
To match other configurations, `Icmp` should be renamed to `ICMPConfig`. 

`Df` should also be refactored to `DF` so that it follows Go's standard naming convention.

TwiN (changes requested):

client/config.go
```diff
@@ -83,11 +83,9 @@ type DNSResolverConfig struct {
 }
 
 // ICMP is the configuration for the ICMP client specific config
-type Icmp struct {
-	// Don't Fragment flag (DF) for the client
-	Df bool `yaml:"df"`
-	// Size of the packet
-	Size int `yaml:"size"`
+type ICMPConfig struct {
+	DF bool `yaml:"df"`    // Don't Fragment flag used for the ICMP client
```
TwiN, line 87:
```suggestion
	DF bool  `yaml:"df"`   // Don't Fragment flag used for the ICMP client
```

client/config.go
```diff
@@ -110,15 +108,15 @@ func (c *Config) ValidateAndSetDefaults() error {
...
+        if c.ICMPConfig.Size < 25 {
+            c.ICMPConfig.Size = 25
         }
     } else {
         // If Icmp is nil, let's initialize it with default values
-        c.Icmp = &Icmp{
-            Df:   false,
+        c.ICMPConfig = &ICMPConfig{
+            DF:   false,
             Size: 25,
```
TwiN, line 120: Based on what I just looked at, the current default value is 25 (on my machine, at least 🤣)

image

Side note, I also see that there's a comment on L112 that may no longer be valid - it says `// limit for pro-ping, below 25 it's not working`.

TwiN (changes requested):

client/config.go
antoinekh: I've set this configuration because it's the only way I've found to make it work.
But maybe I'm missing something in Go.

I did the tests by validating with network captures:
- A classic linux ping can have a data of 0.
- If I don't specify a Size or defaultSize by default the Size is 24.
*So you are right I can remove defaultSize = 24*

But then, If I don't specifiy anything in icmp in my yaml file, the ping is not working.

That's why I put this code with the Size in it.
```
        // If Icmp is nil, let's initialize it with default values
        c.ICMPConfig = &ICMPConfig{
            DF:   false,
            Size: 24,
        }
```
But maybe this is not the good way to initialise it.

If I test with a Size <24 it doesn't work either.
I notice that this limitation exists in other tools that use pro-bing
For example
https://github.com/SuperQ/smokeping_prober
`Packetdata size in bytes. Default 56 (Range: 24 - 65535) ` 

That's why I put
```
    if c.ICMPConfig != nil {
        // limit for pro-ping, below 24 it's not working
        if c.ICMPConfig.Size < 24 {
            c.ICMPConfig.Size = 24
        }
```

And my mistake: the limit is 24, not 25.

antoinekh:
I close this PR, this is an overkill feature, not necessary anymore

Issue #532, opened 2023-07-23
Binding to IPv6 fails or binds to the wrong address


tesinormed:
# Description

When trying to bind to an IPv6 address, whether that may be `::` or a specified address (such as `fdac:8d3d:d2de:aaaa::7`), Gatus fails to start or binds to the wrong address.

# Observations

*Note: the port shown in the log messages will change because of dynamic port allocation.*

## Binding to a valid IPv6 address (with square brackets)

The program fails to start with the following log output.

stderr:
```
2023/07/23 00:16:46 [controller][Handle] Listening on [fdac:8d3d:d2de:aaaa::7]:28215
2023/07/23 00:16:46 [controller][Handle]failed to listen: listen tcp4: address fdac:8d3d:d2de:aaaa::7: no suitable address found
```

## Binding to `::` (with square brackets)
This allows the program to start. However, it binds to 0.0.0.0 (IPv4), as seen in the log output and `netstat`.

stderr:
```
2023/07/23 00:22:29 [controller][Handle] Listening on [::]:25694
```

stdout:
```
 ┌───────────────────────────────────────────────────┐ 
 │                   Fiber v2.46.0                   │ 
 │              https://127.0.0.1:25694              │ 
 │      (bound on host 0.0.0.0 and port 25694)       │ 
 │                                                   │ 
 │ Handlers ............ 26  Processes ........... 1 │ 
 │ Prefork ....... Disabled  PID ................. 1 │ 
 └───────────────────────────────────────────────────┘ 
 ```
 
 netstat:
 ```
 # netstat --all --numeric --inet --inet6 --programs | grep ':25694'
tcp        0      0 0.0.0.0:25694           0.0.0.0:*               LISTEN      558986/gatus    
```

## Binding to a valid IPv6 address (without square brackets)

The program fails to start with the following log output.

stderr:
```
2023/07/23 00:27:06 [controller][Handle] Listening on fdac:8d3d:d2de:aaaa::7:25318
2023/07/23 00:27:06 [controller][Handle]failed to listen: listen tcp4: address fdac:8d3d:d2de:aaaa::7:25318: too many colons in address
```

## Binding to `::` (without square brackets)

The program fails to start with the following log output.

stderr:
```
2023/07/23 00:28:22 [controller][Handle] Listening on :::29233
2023/07/23 00:28:22 [controller][Handle]failed to listen: listen tcp4: address :::29233: too many colons in address
```

# Expectations

[...]

PR #650, opened 2024-01-09
fix: Allow binding IPv6 addresses for web

n-thumann:
## Summary
This PR fixes https://github.com/TwiN/gatus/issues/532.

According to https://docs.gofiber.io/api/fiber#config, Fiber listens on IPv4-only by default. This is the reason why no IPv6 addresses can be bound.

This PR changes the `Network` from `NetworkTCP4` to `NetworkTCP`, allowing both IPv4 and IPv6 (see https://pkg.go.dev/net#Listen).
Binding to the unspecified IPv4 or IPv6 address (`0.0.0.0` or `[::]`) will make it listen on both IPv4 and IPv6. Binding a specific IPv4 or IPv6 address will make it listen on that specific address only.

## Checklist


TwiN (approved):
Makes sense!

TwiN:
@n-thumann Thank you for your contribution!

Issue #185, opened 2021-10-13
Feature Request: choose which IP protocol is used


Carlotronics:
Following the current OVH outage, which only partially affected the services accessible in IPv6, I think it would be a nice feature to be able to distinguish on Gatus the services accessible only in v6 or in v4.

An implementation could be by setting `services[].ip_version` to `inet` or `inet6`:

```yaml
services:
  - name: website IPv4
    url: "https://twin.sh/health"
    ip_version: inet
    conditions:
      - "[STATUS] == 200"

  - name: website IPv6
    url: "https://twin.sh/health"
    ip_version: inet6
    conditions:
      - "[STATUS] == 200"
```

TwiN:
Fixed by #661

ribtoks:
@TwiN Why this is closed? The issue is about chosing IP protocol in general and the author's use-case clearly indicated the HTTP GET request (**not** ICMP request) to "https://twin.sh/health", while the pull request introduced an unrelated ICMP functionality. Can we reopen it until the network mode for HTTP is implemented?

TwiN:
My bad, it was probably a busy day.

Though given nobody pointed this out in over a year,  it seems this feature isn't really useful for most.

Could you create a separate issue? Just FYI, I'd expect the configuration to be under the client configuration, that way, we could make it apply to multiple endpoint types as well.

ribtoks:
@TwiN I can create another issue. But just to clarify, applying it to multiple endpoint types is actuall NOT the intended behavior. The intended behavior is to test a single endpoint in two different ways by resolving domain to IPv4 and IPv6 addresses only. This is the actual source of network problems: service was accessible over IPv4 address, but not via IPv6 address which is hard to detect until some customers complain.

TwiN:
@ribtoks I'm talking about the endpoint-level client configuration. I'm not suggesting this configuration would be applied to all endpoints at the same time, it will still be configurable per-endpoint.

The goal is simply code re-use. I don't want there to be 50 different ways of configuring the same thing.

PR #661, opened 2024-01-20
feat(client): Add client network configuration for ICMP endpoint clients

n-thumann:
## Summary
This PR adds support to explicitly specify the network to use for ICMP endpoint clients, e.g. IPv6.
In the future, this may be extended to also support HTTPS endpoints (requested https://github.com/TwiN/gatus/issues/185).

## Details
I decided to use `ping.NewPinger` instead of `ping.New`, because the latter automatically [resolves](https://github.com/prometheus-community/pro-bing/blob/6bee2459eb4c8ab44f90837a700ae8e1dc0e035c/ping.go#L336) the address before we can set the desired network.
I set `ip` as the default config value. This is technically not needed and could be an empty string, because [`SetNetwork`](https://github.com/prometheus-community/pro-bing/blob/6bee2459eb4c8ab44f90837a700ae8e1dc0e035c/ping.go#L374) takes care of it anyways, but I think it's a cleaner approach.

## Background
I am running Gatus on an IPv6-only machine. Sadly, `net.ResolveIPAddr` still returns a single IPv4 address, when being asked to resolve a dual-stacked host (see https://github.com/golang/go/issues/28666). Pinging the returned IPv4 address from Gatus will obviously fail on an IPv6-only host.
I confirmed this by:
- Adding `ipv6.google.com` as an ICMP endpoint: This works by default on an IPv6-only host, because this target is IPv6-only anyways
- Adding `google.com` as an ICMP endpoint: This doesn't work by default on an IPv6-only host, because this target is dual-stacked and Gatus tries to ping via IPv4. Adding `network: "ip6"` to the endpoints client config fixes this.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -426,17 +426,6 @@ endpoints:
     conditions:
       - "[STATUS] == 200"
 ```
-
-This example shows how you can use the `client.network` configuration to enforce IPv6 for ICMP endpoints:
-```yaml
-endpoints:
-  - name: ping-example
-    url: "icmp://example.com"
-    client:
-      network: "ip6"
-    conditions:
-      - "[CONNECTED] == true"
-```
 > 📝 Note that Gatus will use the [gcloud default credentials](https://cloud.google.com/docs/authentication/application-default-credentials) within its environment to generate the token.
 
 ### Alerting
--- a/README.md
+++ b/README.md
@@ -360,6 +360,7 @@ the client used to send the request.
 | `client.oauth2.scopes[]`               | A list of `scopes` which should be used for the `Client credentials flow`.  | required `[""]` |
 | `client.identity-aware-proxy`          | Google Identity-Aware-Proxy client configuration.                           | `{}`            |
 | `client.identity-aware-proxy.audience` | The Identity-Aware-Proxy audience. (client-id of the IAP oauth2 credential) | required `""`   |
+| `client.network`                       | The network to use for ICMP endpoint client (`ip`, `ip4` or `ip6`).         | `"ip"`          |
 
 > 📝 Some of these parameters are ignored based on the type of endpoint. For instance, there's no certificate involved
 in ICMP requests (ping), therefore, setting `client.insecure` to `true` for an endpoint of that type will not do anything.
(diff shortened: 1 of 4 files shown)


TwiN (changes requested):
Could you add some tests?

n-thumann:
> Could you add some tests?

Done :v:

TwiN (approved):

client/client_test.go
```diff
@@ -83,6 +83,30 @@ func TestPing(t *testing.T) {
...
+		}
+	}
+	if success, rtt := Ping("ipv6.google.com", &Config{Timeout: 500 * time.Millisecond, Network: "ip6"}); !success {
+		t.Error("expected true, because the host is IPv6-only")
+		if rtt == 0 {
+			t.Error("Round-trip time returned on success should've higher than 0")
+		}
+	}
+	if success, rtt := Ping("ipv4.google.com", &Config{Timeout: 500 * time.Millisecond, Network: "ip4"}); !success {
+		t.Error("expected true, because the host is IPv4-only")
```
TwiN, line 105:
```
--- FAIL: TestPing (1.34s)
    client_test.go:99: expected true, because the host is IPv6-only
    client_test.go:101: Round-trip time returned on success should've higher than 0
    client_test.go:105: expected true, because the host is IPv4-only
```
Looks like the integration tests are failing 😔

TwiN (changes requested):

client/client_test.go
n-thumann: I added those two unit tests, because they're for IPv6-only and IPv4-only domains, respectively and it would have been an ideal check to confirm functionality.
Due to the apparent limitation of ICMP being dropped in the network of GitHub-hosted runners (just confirmed this in a test repo) we have two options:

1. Drop these two tests
2. Use a domain that is single-stacked and points to only `::1` or `127.0.0.1` using e.g. sslip.io or backname.io (adds maybe unwanted dependence on theses services)

What would you prefer?

TwiN: Ah, this is tricky.. At least the local icmp check works.

I'm okay with commenting the two integration tests as well as adding a short comment explaining why we can't perform integration tests, however, could you add an additional test verifying the default behavior (neither ip6 nor ip4 -- so the default `ip`) works?

At least this should ensure that future changes do not alter the default behavior.

n-thumann: Sounds good, like this?

> Also, just want to point out that I appreciate you taking the time to be so thorough :)

🤝

TwiN (approved):

README.md
```diff
@@ -425,6 +426,17 @@ endpoints:
...
+This example shows how you can use the `client.network` configuration to enforce IPv6 for ICMP endpoints:
+```yaml
+endpoints:
+  - name: ping-example
+    url: "icmp://example.com"
+    client:
+      network: "ip6"
+    conditions:
+      - "[CONNECTED] == true"
+```
```
TwiN, line 439:
```suggestion

```

TwiN (approved):

TwiN (approved):

TwiN:

PR #637, opened 2023-12-08
fix(api): Add option to set custom fiber ReadBufferSize param

arturkasperek:
## Summary
This PR adds option to setup custom parameter `GATUS_API_READ_BUFFER_SIZE`. This param tweaks fiber `ReadBufferSize` that control max header size (read more here https://github.com/gofiber/fiber/blob/master/docs/api/fiber.md#config).

The problem occurred in my company cause we use gatus on the same domain as app and this causes sending large cookies.

Issue: https://github.com/TwiN/gatus/issues/636

## Checklist

- [x] Tested and/or added tests to validate that the changes work as intended, if applicable.
- [x] Updated documentation in `README.md`, if applicable.

Issue #636, opened 2023-12-08
Request Header Fields Too Large 431


arturkasperek:
### Describe the bug

I have very large `Cookie` header send in request and as a result I get `431` errora 

### What do you see?

Request Header Fields Too Large

### What do you expect to see?

App should work

### List the steps that must be taken to reproduce this issue

1. Set very large header that has 3.5k bytes
2. Notice the 431 error

### Version

v5.7.0

lefes:
Second on this!
Waiting for fix, cuz have the problem with keycloak auth cookies on gatus application

PR #663, opened 2024-01-25
Fix request header fields too large

lifez:
I cannot edit on the previous PR so that I copy and open the new PR #637 
> ## Summary
> This PR adds option to setup custom parameter `GATUS_API_READ_BUFFER_SIZE`. This param tweaks fiber `ReadBufferSize` that control max header size (read more here https://github.com/gofiber/fiber/blob/master/docs/api/fiber.md#config).
> 
> The problem occurred in my company cause we use gatus on the same domain as app and this causes sending large cookies.
> 
> Issue: #636
> 
> ## Checklist
> * [x]  Tested and/or added tests to validate that the changes work as intended, if applicable.
> * [x]  Updated documentation in `README.md`, if applicable.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -40,6 +40,7 @@ func (a *API) createRouter(cfg *config.Config) *fiber.App {
 			log.Printf("[api.ErrorHandler] %s", err.Error())
 			return fiber.DefaultErrorHandler(c, err)
 		},
+		ReadBufferSize: getReadBufferSize(),
 		Network: fiber.NetworkTCP,
 	})
 	if os.Getenv("ENVIRONMENT") == "dev" {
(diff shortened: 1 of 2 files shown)


api/api.go
```diff
@@ -111,3 +112,16 @@ func (a *API) createRouter(cfg *config.Config) *fiber.App {
 	protectedAPIRouter.Get("/v1/endpoints/:key/statuses", EndpointStatus)
 	return app
 }
+
+func getReadBufferSize() int {
```
TwiN, line 116: Any chance you could add a test that ensures this function returns 4096 by default as well as whatever custom value if the env var is set?

TwiN (changes requested):

TwiN:
Superseded by #675

PR #675, opened 2024-02-07
fix(web): Allow configuration of read-buffer-size

### LINKED ISSUE #636: Request Header Fields Too Large 431

### LINKED ISSUE #674: Request Header Fields Too Large
### Describe the bug

I have a small monitors like 6. The all alerts is so simple like the screenshot below. To works I need restart the Gatus each 10 minutes due the error "[api.ErrorHandler] Request Header Fields Too Large"

image

### What do you see?

[api.ErrorHandler] Request Header Fields Too Large

image

I don't remember but some version in the past I never got this error but now I am not able to say that version

1. Just install using normal way
2. In my case my monitors came from Prometheus like that http://thanos-query-frontend.thanos:9090/api/v1/query?query=sum%28avg_over_time%28xxxxxxxxx%7Bxxxxxxx%3D~%22.%2A%22%2C%20xxxxxx%3D~%22xxxxxxxxx%22%2C%20xxxxxx%21%3D%22xxxxxxxxx%22%7D%5B10m%5D%29%29%20by%20%28xxxxxxx%2C%20topic%29

***I guess the issue is the long URL because I tried to removed and the application run without any errors. There is possible fix it internally ? I need use the Prometheus Queries and almost time the URL is too long because I need convert it to works in Gatus as well

***** Update: It worked for some hours and the error back it again so, is not the issue with long URLs.

### Version

5.7.0

### Additional information

 2024/02/07 16:54:48 [watchdog][execute] Monitored group=Applications; endpoint=xxxx success=true; errors=0; duration=2ms

 2024/02/07 16:54:49 [watchdog][execute] Monitored group=Integrations; endpoint=xxxx; success=true; errors=0; duration=33ms

 2024/02/07 16:54:49 [watchdog][execute] Monitored group=Integrations; endpoint=xxxx; success=true; errors=0; duration=50ms

 2024/02/07 16:54:50 [watchdog][execute] Monitored group=API's; endpoint=xxxxx; success=true; errors=0; duration=13ms

 2024/02/07 16:54:51 [watchdog][execute] Monitored group=API's; endpoint=xxxxxx; success=true; errors=0; duration=42ms

 2024/02/07 16:54:52 [api.ErrorHandler] Request Header Fields Too Large

 2024/02/07 16:54:52 [api.ErrorHandler] Request Header Fields Too Large

 2024/02/07 16:54:54 [api.ErrorHandler] Request Header Fields Too Large

 2024/02/07 16:54:54 [api.ErrorHandler] Request Header Fields Too Large

TwiN:
## Summary
This fixes the `431 Request Header Fields Too Large` error

By default, the read-buffer-size is 8192, up from fiber's default of 4096.

Fixes #674
Fixes #636
Supersedes #637
Supersedes #663

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,57 +206,58 @@ If you want to test it locally, see [Docker](#docker).
 ## Configuration
 | Parameter                                       | Description                                                                                                                                 | Default                    |
 |:------------------------------------------------|:--------------------------------------------------------------------------------------------------------------------------------------------|:---------------------------|
-| `debug`                                         | Whether to enable debug logs.                                                                                                                   | `false`                    |
-| `metrics`                                       | Whether to expose metrics at /metrics.                                                                                                          | `false`                    |
-| `storage`                                       | [Storage configuration](#storage)                                                                                                               | `{}`                       |
-| `endpoints`                                     | List of endpoints to monitor.                                                                                                                   | Required `[]`              |
-| `endpoints[].enabled`                           | Whether to monitor the endpoint.                                                                                                                | `true`                     |
-| `endpoints[].name`                              | Name of the endpoint. Can be anything.                                                                                                          | Required `""`              |
-| `endpoints[].group`                             | Group name. Used to group multiple endpoints together on the dashboard. <br />See [Endpoint groups](#endpoint-groups).                          | `""`                       |
-| `endpoints[].url`                               | URL to send the request to.                                                                                                                     | Required `""`              |
-| `endpoints[].method`                            | Request method.                                                                                                                                 | `GET`                      |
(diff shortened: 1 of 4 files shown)


mingchuno:
@TwiN  Can we make a new release with this change? Our production env is seeing the "Request Header Too Large" error and we cannot use the `latest` docker tag due to policy constrain. 

What can I do to help expedite? Thank You.

PR #690, opened 2024-02-28
feat: Added support for Docker secrets

durandj:
## Summary

This adds the ability to use [Docker
secrets](https://docs.docker.com/compose/use-secrets/) in configuration files. In Docker this is done by creating a secret resource, adding that secret to the container when you're starting it, and specifying an environment variable that has the `_FILE` suffix. The environment variable should point to the file which contains the secret (`/run/secrets/<secret-name>`).

Typically Docker images are setup so that they will try and find any environment variables that end in `_FILE` and set new environment variables with the same name minus the `_FILE` suffix in the running process. This is beneficial since environment variables that are set by the user when creating the container are visible to anyone who is able to run `docker container inspect <container>` on the host. For secrets this could be really damaging and leak sensitive information. Instead it is recommended to use Docker secrets.

Because Gatus uses the `scratch` base image I wasn't able to just use a Bash script to convert the secret file path into a normal environment variable like many other images do. Instead I opted to just modify the configuration logic so that it checks the environment variable name and changes its behavior based on that. This seems to work well enough.

As far as error handling, I opted _not_ to crash the service when it's unable to read the secret file and instead just pretend its a normal environment variable and return an empty string. This follows the conventions of the rest of the configuration handling and leaves the error reporting to the configuration validation.

I've also updated the readme to mention this feature with a link to an example.

## Checklist

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -3,6 +3,7 @@ package config
 import (
 	"errors"
 	"fmt"
+	"io"
 	"io/fs"
 	"log"
 	"os"
@@ -223,7 +224,7 @@ func parseAndValidateConfigBytes(yamlBytes []byte) (config *Config, err error) {
 	// environment variable. This allows Gatus to support literal "$" in the configuration file.
 	yamlBytes = []byte(strings.ReplaceAll(string(yamlBytes), "$$", "__GATUS_LITERAL_DOLLAR_SIGN__"))
 	// Expand environment variables
-	yamlBytes = []byte(os.ExpandEnv(string(yamlBytes)))
+	yamlBytes = []byte(os.Expand(string(yamlBytes), expandEnvironmentVariable))
 	// Replace __GATUS_LITERAL_DOLLAR_SIGN__ with "$" to restore the literal "$" in the configuration file
 	yamlBytes = []byte(strings.ReplaceAll(string(yamlBytes), "__GATUS_LITERAL_DOLLAR_SIGN__", "$"))
 	// Parse configuration file
@@ -261,6 +262,31 @@ func parseAndValidateConfigBytes(yamlBytes []byte) (config *Config, err error) {
 		}
 	}
 	return
+}
+
+func expandEnvironmentVariable(name string) string {
+	secretPathVarName := name + "_FILE"
+	secretPath, secretIsSet := os.LookupEnv(secretPathVarName)
+
+	if !secretIsSet {
+		return os.Getenv(name)
+	}
+
+	secretFile, err := os.Open(secretPath)
+	if err != nil {
+		return ""
+	}
+
+	defer func() {
+		_ = secretFile.Close()
+	}()
+
+	secretBytes, err := io.ReadAll(secretFile)
+	if err != nil {
+		return ""
+	}
+
+	return strings.TrimSpace(string(secretBytes))
 }
 
 func validateConnectivityConfig(config *Config) error {
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -3,6 +3,7 @@ package config
 import (
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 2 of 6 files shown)


onedr0p:
> This is beneficial since environment variables that are set by the user when creating the container are visible to anyone who is able to run docker container inspect <container> on the host. For secrets this could be really damaging and leak sensitive information. Instead it is recommended to use Docker secrets.

If an attacker can run `docker container inspect` to get the secrets, surely they also have access to the secret file on disk on the host unless a person is dumb enough to expose the Docker tcp socket over their network.

durandj:
It depends on how you initialize the secrets. Yes, at minimum you've moved your secrets to a different file that can be owned by some other user and not readable however you can also setup secrets from other sources. I want to not have my secrets live on disk and in an unencrypted for. That's completely nullified if someone just needs to then have access to the Docker daemon and they can now see anything that I've configured.

config/config.go
```diff
@@ -263,6 +264,31 @@ func parseAndValidateConfigBytes(yamlBytes []byte) (config *Config, err error) {
...
+		_ = secretFile.Close()
+	}()
+
+	secretBytes, err := io.ReadAll(secretFile)
+	if err != nil {
+		return ""
+	}
+
+	return strings.TrimSpace(string(secretBytes))
+}
```
TwiN, line 290:
```suggestion
func expandEnvironmentVariable(name string) string {
	secretPath, secretIsSet := os.LookupEnv(name + "_FILE")
	if !secretIsSet {
		return os.Getenv(name)
	}
	secretFile, err := os.Open(secretPath)
	if err != nil {
		return ""
	}
	defer secretFile.Close()
	secretBytes, err := io.ReadAll(secretFile)
	if err != nil {
		return ""
	}
	return strings.TrimSpace(string(secretBytes))
}
```

config/config_test.go
```diff
@@ -1616,3 +1617,129 @@ func TestGetAlertingProviderByAlertType(t *testing.T) {
...
+
+	t.Cleanup(func() {
+		_ = os.Remove(tempFile.Name())
+	})
+
+	if _, err := io.WriteString(tempFile, secretValue); err != nil {
+		t.Errorf("unable to write secret to temporary file: %s", err.Error())
+	}
+
+	os.Setenv("GATUS_TestExpandingOfFileEnvironmentVariables_FILE", tempFile.Name())
```
TwiN, line 1637: can you unset the env var after this test using defer? Just to prevent surprises.

config/config_test.go
```diff
@@ -1616,3 +1617,129 @@ func TestGetAlertingProviderByAlertType(t *testing.T) {
...
+		t.Errorf(
+			"should default to empty string when variables aren't set, expected: %s but got %s",
+			"http://localhost:8080",
+			actualValue,
+		)
+	}
+}
+
+func TestExpandingOfFileEnvironmentVariablesMissingFile(t *testing.T) {
+	os.Setenv("GATUS_TestExpandingOfFileEnvironmentVariables_FILE", "TestExpandingOfFileEnvironmentVariablesMissingFile.txt")
```
TwiN, line 1683: Same here

config/config_test.go
```diff
@@ -1616,3 +1617,129 @@ func TestGetAlertingProviderByAlertType(t *testing.T) {
...
+	t.Cleanup(func() {
+		_ = os.Remove(tempFile.Name())
+	})
+
+	if _, err := io.WriteString(tempFile, secretValue); err != nil {
+		t.Errorf("unable to write secret to temporary file: %s", err.Error())
+	}
+
+	os.Setenv("GATUS_TestExpandingOfFileEnvironmentVariables_FILE", tempFile.Name())
+	os.Setenv("GATUS_TestExpandingOfFileEnvironmentVariables", otherSecretValue)
```
TwiN, line 1724: Here too

TwiN (changes requested):

durandj:
@TwiN I've addressed your comments, thanks for reviewing!

ribtoks:
This is a great functionality, can we get this merged? Looks like we need to resolve conflicts once again

Issue #641, opened 2023-12-23
XMPP WSS Bad Status


roughnecks:
### Describe the bug

Not sure if a bug but I cannot get the wss endpoint to work with my prosody instance.

This is my configuration:

```
  - name: prosody
    group: core
    url: "wss://woodpeckersnest.space:5281/xmpp-websocket"
    body: "status"
    conditions:
      - "[CONNECTED] == true"
      - "[BODY].result >= 0"
```

### What do you see?

![image](https://github.com/TwiN/gatus/assets/1666536/79bc6088-80da-4afb-bf4d-9c58312538fb)

### List the steps that must be taken to reproduce this issue

1. Configure wss endpoint
2. Start application
3. Look at the webpage - error in the wss endpoint

### Version

I installed with GO

TwiN:
To be honest, I don't currently monitor any websocket endpoint myself, but perhaps @heitorPB will see something I don't as they're the one who implemented the feature 🤔

roughnecks:
Hi @heitorPB What logs are you referring to?

I tried enabling debug logs but `journalctl -xeu gatus.service` doesn't say much
And I don't know why there's no nginx logs, like access and error logs are not even created.

heitorPB:
I was referring to the Gatus logs, in debug mode. Something like `journalctl --unit gatus.service --grep "wss"`.

Can you connect directly to your websocket? Like `websocat woodpeckersnest.space:5281/xmpp-websocket`?

I can't reach it via websocket:
```
$ websocat -v -H='Sec-Websocket-Protocol: xmpp' -H='seb-websocket-version: 13' -H='upgrade: websocket' wss://woodpeckersnest.space/xmpp-websocket
[INFO  websocat::lints] Auto-inserting the line mode
[INFO  websocat::stdio_threaded_peer] get_stdio_peer (threaded)
[INFO  websocat::ws_client_peer] get_ws_client_peer
websocat: WebSocketError: WebSocketError: Received unexpected status code (404 Not Found)
websocat: error running
```

roughnecks:
The only thing I can say to you now is that I have a bot connected to wss (IPv4 only).

PR #702, opened 2024-03-07
feat: Allow custom origin header for Websocket

hamann:
## Summary

The issue described in https://github.com/TwiN/gatus/issues/641#issuecomment-1980446434 was caused by an hardcoded origin header at https://github.com/TwiN/gatus/blob/1e431c797ad265b43269d1dfe270f1acd37d24fa/client/client.go#L265 which got validated by the websocket server and returend 403 (`bad status`, from https://github.com/golang/net/blob/ab271c317248ea0f18481852f96d12d5eca05cf8/websocket/hybi.go#L440-L445)
This PR allows to customize the origin header.
Example
```
endpoints:
  - name: websocket
    url: "wss://example.org"
    interval: 10s
    headers:
      Origin: "http://gatus.example.org"
    conditions:
      - "[CONNECTED] == true"
```

## Checklist


client/client.go
```diff
@@ -260,12 +260,11 @@ func Ping(address string, config *Config) (bool, time.Duration) {
 }
 
 // QueryWebSocket opens a websocket connection, write `body` and return a message from the server
-func QueryWebSocket(address, body string, config *Config) (bool, []byte, error) {
+func QueryWebSocket(address, body string, config *Config, origin string) (bool, []byte, error) {
```
TwiN, line 263:
```suggestion
func QueryWebSocket(address, body, origin string, config *Config) (bool, []byte, error) {
```

Can you move the origin before the config parameter? 🙏

TwiN (changes requested):

core/endpoint.go
```diff
@@ -221,6 +224,14 @@ func (endpoint *Endpoint) ValidateAndSetDefaults() error {
 	if _, contentTypeHeaderExists := endpoint.Headers[ContentTypeHeader]; !contentTypeHeaderExists && endpoint.GraphQL {
 		endpoint.Headers[ContentTypeHeader] = "application/json"
 	}
+
+	// Automatically add Origin header for websocket endpoints if there isn't one specified
+	if endpoint.Type() == EndpointTypeWS {
+		if _, originHeaderExists := endpoint.Headers[OriginHeader]; !originHeaderExists {
+			endpoint.Headers[OriginHeader] = Origin
+		}
+	}
```
TwiN, line 233: While I agree that adding support for passing the origin header is worth it, I think perhaps we shouldn't add it by default here but instead add a note in the docs mentioning that users _may_ have to specify an `Origin` header.

kadaan: @TwiN It seems as the code as in this PR ensures that the current behavior, which users might rely on, continues to functions as it has.  I'm unsure if ceasing the specify the Origin header will end up being a backwards incompatible change. What is the benefit of no longer specifying the Origin header on any websocket call unless it is explicitly specified?

TwiN: @kadaan I don't personally have any websocket endpoint handy to test this on; I'm just trying to avoid backward incompatible changes to the best of my ability. Could you investigate this? Once the conflicts are resolved, I wouldn't mind merging this.

config/endpoint/endpoint.go
```diff
@@ -48,6 +48,9 @@ const (
 	TypeWS       Type = "WEBSOCKET"
 	TypeSSH      Type = "SSH"
 	TypeUNKNOWN  Type = "UNKNOWN"
+
+	Origin       = "http://localhost"
```
TwiN: We should specify that this is the default origin for websocket endpoints. The name should be something like DefaultWebsocketOrigin

config/endpoint/endpoint.go
```diff
@@ -203,6 +206,14 @@ func (e *Endpoint) ValidateAndSetDefaults() error {
 	if _, contentTypeHeaderExists := e.Headers[ContentTypeHeader]; !contentTypeHeaderExists && e.GraphQL {
 		e.Headers[ContentTypeHeader] = "application/json"
 	}
+
+	// Automatically add Origin header for websocket endpoints if there isn't one specified
+	if e.Type() == TypeWS {
```
TwiN, line 211: Why is this necessary? Shouldn't it already be possible to specify the Origin header? All your change seems to be doing is adding a default Origin header of `http://localhost` if the endpoint is a websocket endpoint and there is no origin header already specified

TwiN (changes requested):

config/endpoint/endpoint.go
hamann: Yes, the default is set in `ValidateAndSetDefaults` to stay backwards compatible if there isn't one. Later it's passed to `client.QueryWebSocket`. How else would you do it?

TwiN: My question is more that if you know what you want, wouldn't you just be able to set the headers configuration without this change? If so, then I'd expect the PR to be a change to the README, not a change to the default behavior.

Your change sets the `Origin` header value to `http://localhost`, but personally, I don't agree with that default value. I'd much rather document that websocket endpoints probably want the `Origin` header defined

PR #712, opened 2024-03-21
feat(alerting): discord provider message content config

GarrettRobinson:
## Summary
Discord provider config entry for the Content field when posting messages to the service.

Discord limitations don't allow mentions inside embeds to ping/highlight.

Using the content field allows a mention to:
- Ping the user/role with an alert
- Increase the notification counter on the server/channel
- Add a message to the notification inbox

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -483,7 +483,7 @@ alerting:
   discord:
     webhook-url: "https://discord.com/api/webhooks/**********/**********"
     title: "Example Title"
-    content: "Hey <@123456789012345678" look!"
+    content: "Hey <@123456789012345678> look!"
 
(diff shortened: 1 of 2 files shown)


TwiN (changes requested):
Couple things:
1. Why not just have a channel specifically for these types of alerts, and have people edit the notification settings for that channel so that they get notified for every message?
2. Some of the indentations you made use spaces instead of tabs
3. You'll have to update the tests or they'll fail
4. In the documentation, I'd suggest you add a note that says:
> optionally, you can use the `alerting.discord.content` field to ping specific users.

Issue #679, opened 2024-02-17
Alerts triggered status is not persistent


macmoritz:
### Describe the bug

The `triggered` status of a monitor is only momentary. After a restart of gatus the information is gone.  
   
Such restarts are daily business in cloud environments. 

### What do you expect to see?

Gatus should only send one alert for a failing monitor. 

### List the steps that must be taken to reproduce this issue

1. gatus starts
2. after some time a monitored service fails
3. gatus sends an alert after checking if alert was sended before
4. gatus will not send more notifications for this monitor, which is correct (log: `[watchdog][handleAlertsToTrigger] Sending matrix alert because alert for endpoint=my-endpoint='' has been TRIGGERED`)
5. gatus restarts
6. gatus will send an alert for the monitor again

### Additional information

Fix could be to use the events from the database for checking if an alert was already sended.

TwiN:
Yes, you're right. It's also something I've been wanting to take care of for a while.

ser:
I would also add something like UUID of the sent alert and keep the same UUID when the alert gets resolved or sending is repeated by any reason.

TwiN:
PR is up @ #764 but further testing is necessary

PR #724, opened 2024-04-07
feat: Implement push-based external endpoints

### LINKED ISSUE #95: Feature Request: client health update
Is there a way or better tool that can have the client send updates to Gatus.

For example having a cron job that executes a curl against http(s)://gatus.example.com/client/{computer name}/{service name}/{check}

With a monitor that if the check for heartbeat  hasn’t been received in 10 minutes it will generate an alert?  This would make it possible for systems only allowed to send out but not in based on network segmentation to provide updates.
Thank you,
Kensel

### LINKED ISSUE #673: Push-based monitoring

First, thanks for gatus, a very useful tool!
Is it possible to add push-based monitoring, where external services can report to gatus that they are up ? 
Example of such feature are [uptime-kuma](https://github.com/louislam/uptime-kuma) push, and [healthchecks.io](https://github.com/healthchecks/healthchecks). 

It could be : services would call an exposed gatus URL (with a specific `<id>` set in the config file) which would record the time of last access. Then the periodic monitoring function would set the monitoring "unhealthy"  if the heartbeat was missed in the last interval.

This will allow for monitoring the absence of events, eg use at the end of a periodic backup script, to be be able to monitor that the backup has not happened.

3 months

### Additional information

I looked at #229 : the title "Expose an API for external applications to push statuses" seems to be similar to the above request, but the description is about something else (API to add/modify/delete endpoints).

### LINKED ISSUE #722: Implement endpoints whose results are pushed into Gatus externally

This feature has been requested dozens of times to me, and I'm feeling motivated to implement it right now, so here goes nothing.

The crux of it is, let's say Gatus doesn't support something, but you want to use Gatus to display and alert on that "unsupported something" anyways. Well, this is the feature that enables it.

The steps involved are:
1. Figure out a way to classify an endpoint as "external"
  1.1. said endpoint must have an individually configurable "token" which is used for authentication purposes. 
2. Create an API that allows an HTTP request to "report" a result about an endpoint
3. Said endpoint must automatically check if an alert has been triggered

_No response_

Forever

### Additional information

_No response_

TwiN:
## Summary

Implements push-based external endpoints

Also refactored some code here and there & fixed formatting on the docs

Fixes #722 
Fixes #673
Fixes #95 

## Checklist


TwiN:
There's one more issue to fix - `NumberOfFailuresInARow` and `NumberOfSuccessesInARow` are not persisted in the database (sql storage), which means that tomorrow's me will have to ensure that the `external-endpoints[].alerts` cfg is respected properly.

I _may_ have to persist NumberOfFailuresInARow and NumberOfSuccessesInARow -- it's extra work, _but_ it would also fix #679

Issue #742, opened 2024-04-22
Provide Backend Error messages via Gatus frontend


PhilippKolmann:
### Describe the feature request

I have found your project today and think it suits my needs perfectly. I would love to have a possibitliy to get any errors, the backend put in the [BODY].errors JSON part, that these could be displayed at the frontend in the errors section.

That would make debugging much easier.

If the Body is not a JSON, a config switch to add the whole Body to the Error Messages would be great as well.

### Why do you personally want this feature to be implemented?

As stated above, I think it would make debugging much easier.

### How long have you been using this project?

just found today

PR #743, opened 2024-04-23
Add endpoints.onErrorAdd

PhilippKolmann:
## Summary
Add endpoints[].onErrorAdd to add response body or parts of it as error, when the conditions fail for easier debugging.

Implements https://github.com/TwiN/gatus/issues/742

## Checklist


TwiN:
I would be more okay with having the `debug` parameter print the body in the logs honestly. 
The big problem is that it's pretty common for firewalls or outages to cause normally JSON or plain text endpoints to return HTML pages, and those HTML pages can be pretty lengthy, something that may cause issues both at the storage level and at the display level on the UI

PhilippKolmann:
My intention was, that I see immediately what my micro services are lacking because they write it in the body.

But I see your point.

PhilippKolmann:
Would you prefer to have a limit of text that could be added, and if the body is over the limit, then to truncate or not print it?

TwiN:
There is now support for logging levels on Gatus. I'm going to close this for now,

PR #748, opened 2024-04-28
fix: Use privileged mode for ICMP only on Windows

### LINKED ISSUE #697: ICMP not working on Kubernetes even if sysctl -w net.ipv4.ping_group_range="0 2147483647"
### Describe the bug

Ping not working even if sysctl -w net.ipv4.ping_group_range="0 2147483647" is configured.
net.ipv4.ping_group_range is set in initContainer which have privileged role an run as root.

I tried to mount an alpine container in same pod where gatus is. I can successfully use ping in this alpine container which have same securitycontext as gatus container, is running a non root user, on uid and guid 65534 (nobody) 

I tried to ping google.com

Used config in config.yaml
endpoints:
  - name: TEST
    enabled: true
    url: "icmp://google.com"
    interval: 60s
    conditions:
    - "[CONNECTED] == true"
    client:
      timeout: 30s

### What do you see?

Endpoint is returning not ok

Ping success

Use icmp on endpoint running in Kubernetes using this endpoints config
endpoints:
  - name: TEST
    enabled: true
    url: "icmp://google.com"
    interval: 60s
    conditions:
    - "[CONNECTED] == true"
    client:
      timeout: 30s
      
Use initContainer with root privlege using alpine image and execute this command:
 sysctl -w net.ipv4.ping_group_range="0 2147483647"

### Version

twinproduction/gatus:v5.7.0

### Additional information

_No response_

TwiN:
## Summary

Use privileged mode for ICMP only on Windows

Fixes #697 

## Checklist

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -238,12 +238,8 @@ func Ping(address string, config *Config) (bool, time.Duration) {
 	pinger := ping.New(address)
 	pinger.Count = 1
 	pinger.Timeout = config.Timeout
-	// Set the pinger's privileged mode to true for every GOOS except darwin
-	// See https://github.com/TwiN/gatus/issues/132
-	//
-	// Note that for this to work on Linux, Gatus must run with sudo privileges.
-	// See https://github.com/prometheus-community/pro-bing#linux
-	pinger.SetPrivileged(runtime.GOOS != "darwin")
+	// See https://github.com/TwiN/gatus/issues/132 and https://github.com/TwiN/gatus/issues/697
+	pinger.SetPrivileged(runtime.GOOS == "windows")
 	pinger.SetNetwork(config.Network)
 	err := pinger.Run()
 	if err != nil {


TwiN:
Temporarily published container image as `twinproduction/gatus:experimental` for testing purposes.

h3mmy:
I've verified that this works in my environment for the icmp check. Details here: https://github.com/TwiN/gatus/issues/697#issuecomment-2141077524

h3mmy (approved):
LGTM. Thanks for making this change!

TwiN:
FYI this is on-hold and is being discussed in #697

ignisf:
Hya, just chiming in that this fixes ICMP with rootless `podman` containers.

h3mmy:
Is there anything specifically preventing this from being merged?

TwiN:
@h3mmy https://github.com/TwiN/gatus/issues/697#issuecomment-2081700989 is why I'm not merging it.

https://github.com/TwiN/gatus/issues/697#issuecomment-2201176988 is the reason why this needs more investigation.

The fix itself resolves an issue for some people, but causes more issues for other people. This still needs some investigation, and I don't have the time to spend on this at the moment. Also worth pointing out that for me, the current release works just fine, but the fix does not _unless_ I add 
```
      securityContext:
        sysctls:
          - name: net.ipv4.ping_group_range
            value: 0 65536
```
to the pod security context, which means that this would technically be a regression for my cloud environment, and likely others.

TwiN:
Superseded by #1056

Issue #310, opened 2022-08-04
Custom alert 


sicute:
### Describe the bug

i have webhook with ryver or any webhook  which need describe content type such as "text/plain" or anything .
and i check logs was content type '' not support .

call to provider alert returned status code 400: {"error":{"code":"client.http.error","lang":"en-US","value":"content type '' not supported"}

### What do you see?

```
custom:
    url: "https://aichatco.ryver.com/application/webhok/xxx"
    method: "POST"
    body: |
      {
        "text":"[ALERT_TRIGGERED_OR_RESOLVED]: [ENDPOINT_GROUP] - [ENDPOINT_NAME] - [ALERT_DESCRIPTION]"
      }
```

### What do you expect to see?

Here example webhook was working in console

curl -H "Content-Type: text/plain; charset=utf-8" -X POST  -d "**hiii**" https://aichatco.ryver.com/application/webhook/xxxx

### List the steps that must be taken to reproduce this issue

no metrics , 404 .

### Version

latest

### Additional information

i want replace statping with your apps .
i using docker-compose with sqldatabase

TwiN:
Have you looked at this? https://github.com/TwiN/gatus#configuring-custom-alerts

You can specify your own headers like so:
```yaml
custom:
  # ...
  headers:
    Content-Type: "text/plain"
```

By default, if you don't specify any header with the name `Content-Type`, the `Content-Type` header will default to `application/json`.

TwiN:
Since you haven't replied, I'm going to close this issue.

TwiN:
@ser You're right, I apologize for the misunderstanding.

PR #751, opened 2024-05-01
fix(alerting): Set Content-Type to application/json by default for custom provider if none are defined

TwiN:
## Summary

In #310, I said that by default, if you don't specify any header with the name Content-Type, the Content-Type header will default to application/json.

That statement was wrong, but after this is merged, it won't be.

## Checklist


TwiN:
Nevermind, this isn't necessary.

PR #763, opened 2024-05-14
feat(security): added the ability to use a different claim field to match against the allow-subjects list.

LCerebo:
## Summary

You can use a custom claim field to identify users i.e. 'email' or 'preferred_username'.
If not specified the default is 'subject'.

## Checklist

##Notes
Tested using a keycloak server as OIDC provider.
I tried to add some unit tests but I'm not sure how to mock a valid OIDC server.


LCerebo:
@TwiN can you take a look in this PR? 
Thanks in advance.

TwiN (changes requested):
Is there any way you could add a test to ensure that the feature works as intended?

TwiN:
Just checking up again - before I can merge this, I'd like a few tests to prove that it does work as intended, and also to ensure that future changes won't break the feature you're adding.

I am sorry for requiring this before merging, but my hands are tied as I have a duty to Gatus users to not neglect security.

LCerebo:
> Just checking up again - before I can merge this, I'd like a few tests to prove that it does work as intended, and also to ensure that future changes won't break the feature you're adding.
> 
> I am sorry for requiring this before merging, but my hands are tied as I have a duty to Gatus users to not neglect security.

@TwiN sorry for the long delay, I've added the tests as required. Let me know if it's enough.
Thanks for the patience.

security/oidc.go
```diff
@@ -125,17 +126,46 @@ func (c *OIDCConfig) callbackHandler(w http.ResponseWriter, r *http.Request) { /
...
+		http.Redirect(w, r, "/", http.StatusFound)
+		return
 	}
-	logr.Debugf("[security.callbackHandler] Subject %s is not in the list of allowed subjects", idToken.Subject)
 	http.Redirect(w, r, "/?error=access_denied", http.StatusFound)
 }
 
+func (c *OIDCConfig) isAuthorized(subject string, claimsMap map[string]any) bool {
+	claimToCheck := c.ClaimToCheck
+	if len(claimToCheck) > 0 {
```
TwiN, line 147: Instead of doing it like this, extract the claim to compare (sub, or the custom one), and then iterate over the list of allowed subjects. It makes the code easier to read.

README.md
```diff
@@ -2633,6 +2633,7 @@ security:
 | `security.oidc.client-secret`    | Client secret                                                  | Required `""` |
 | `security.oidc.scopes`           | Scopes to request. The only scope you need is `openid`.        | Required `[]` |
 | `security.oidc.allowed-subjects` | List of subjects to allow. If empty, all subjects are allowed. | `[]`          |
+| `security.oidc.claim-to-check`   | Name of the field to use to match against `allowed-subjects`   | `""`            |
```
TwiN: I think I'd rather this be called `custom-claim-to-check`, and if it's not defined, then we just use the subject. The description should also be  `Name of the claim to use to match against allowed-subjects`.

My reasoning for this is that if we leave it as `claim-to-check`, people may be tempted to set it to `sub` or `subject`, which wouldn't work under the current implementation.

Now that I say it out loud, it may be prudent to also use idToken.Subject if the user set the parameter to `sub`, just in case

README.md
```diff
@@ -2645,6 +2646,8 @@ security:
     scopes: ["openid"]
     # You may optionally specify a list of allowed subjects. If this is not specified, all subjects will be allowed.
     #allowed-subjects: ["johndoe@example.com"]
+    # You can specify a different claim field to match against allowed-subjects. If not specified "subject" is used.
```
TwiN: This isn't accurate. The field name is `sub`, not `subject`. This also kind of implies that the value of the claim-to-check parameter will be set to that value, which isn't correct, as that may in fact not work (not sure, but I think the claims map is just the custom claims, not the default fielde like `sub`)

security/oidc.go
LCerebo: Change as requested in 675347a29caaf5bcf0d9de6233e635d43ca581c3

README.md
LCerebo: Change as requested in 675347a29caaf5bcf0d9de6233e635d43ca581c3

README.md
LCerebo: Change as requested in 675347a29caaf5bcf0d9de6233e635d43ca581c3

LCerebo:
@TwiN  let me know if everything is good and ready to merge.
Thanks!

LCerebo:
@TwiN sorry to bother you again, just let me know if everything is ok for you.
Thanks.

PR #764, opened 2024-05-15
feat(alerting): Persist triggered alerts across application restart

### LINKED ISSUE #679: Alerts triggered status is not persistent

TwiN:
## Summary

Fixes #679

## Checklist

### DIFF
--- a/alerting/alert/alert.go
+++ b/alerting/alert/alert.go
@@ -1,7 +1,10 @@
 package alert
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"errors"
+	"strconv"
 	"strings"
 )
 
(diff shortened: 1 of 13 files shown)


LCerebo:
@TwiN  I tried you PR since I was trying to look at this feature myself. When I restart gatus for the second time I re-receive all the OK notification.  The KO ones seems fine. Later today I will try to look more in detail your commit to give you some more details.

TwiN:
@LCerebo Much appreciated! Due to the nature of the change, I will probably let it sit in latest for a few weeks in case somebody reports an issue with it.

PR #834, opened 2024-08-07
fix: Logo on Login screen

praveen-livspace:
## Summary
Logo on the login screen is hardcoded. The proposed change uses the configured logo if available.

## Checklist


TwiN:
Does that really make sense though?
I understand wanting to have to user logo on the page, but if there's an authentication page, wouldn't users want to know what they're authenticating to?

I'd be willing to meet in the middle and have some kind of `<custom logo> | <Gatus' logo>` design, but I feel just having the custom logo may be a bit confusing for the end user

PR #875, opened 2024-10-02
feat(metrics): Add URL to metrics labels

danlsgiga:
Small feature addition.

Sometimes is useful to have the endpoint URL in metrics labels so we can include that as part of alert messages when these metrics are scraped via prometheus.


j6nca (approved):

danlsgiga:
@TwiN any feedback on this PR?

TwiN:

TwiN:
@danlsgiga I don't do this usually, but I had to revert the PR because I don't have the time to fix this right now.

After merging, I found out that the URL doesn't discard the query parameters, and that's an issue because unfortunately, many people pass tokens as query parameter (e.g. `https://example.org/api/v1/example?token=123`), which would make releasing this a security risk.

If you can make another PR that discards all query parameters (`?token=123`, in the example above), I could merge your feature.

danlsgiga:
Oh, I'm sorry @TwiN, I totally missed this message! I'll look into this and open a new PR! thanks!

Issue #767, opened 2024-05-18
Allow setting a priority on resolved Pushover notifications


onedr0p:
### Describe the feature request

It would be nice if I could set a separate priority for when alerts are resolved using Pushover. 

Ideally I want alerts when triggered to be 1 (which is possible now) and when resolved 0 (not currently possible).

This is how Prometheus and Alertmanager currently operate for pushover for me:

```yaml
priority: '{{ if eq .Status "firing" }}1{{ else }}0{{ end }}'
```

I would be over the moon if gatus could do something like that too.

### Why do you personally want this feature to be implemented?

This would help cut down on alerting fatigue since some people might only be interested in getting alerted on problems at a higher priority and alerted on a lower priority when resolved.

### How long have you been using this project?

18 months

onedr0p:
@TwiN I would like to try and implement this but I am curious if you would be fine with the following config? The way `alerting.pushover.priority-on-resolved` would work is that it wouldn't be a breaking change to those using pushover already.

#### Configuring Pushover alerts
| Parameter                              | Description                                                                                     | Default                      |
|:---------------------------------------|:------------------------------------------------------------------------------------------------|:-----------------------------|
| `alerting.pushover`                    | Configuration for alerts of type `pushover`                                                     | `{}`                         |
| `alerting.pushover.application-token`  | Pushover application token                                                                      | `""`                         |
| `alerting.pushover.user-key`           | User or group key                                                                               | `""`                         |
| `alerting.pushover.title`              | Fixed title for all messages sent via Pushover                                                  | Name of your App in Pushover |
| `alerting.pushover.priority`           | Priority of all messages, ranging from -2 (very low) to 2 (emergency)                           | `0`                          |
| `alerting.pushover.priority-on-resolved`           | Priority of message on resolved, ranging from -2 (very low) to 2 (emergency)                           | `alerting.pushover.priority`                          |
| `alerting.pushover.sound`              | Sound of all messages<br />See [sounds](https://pushover.net/api#sounds) for all valid choices. | `""`                         |
| `alerting.pushover.default-alert`      | Default alert configuration. <br />See [Setting a default alert](#setting-a-default-alert)      | N/A                          |

Marck:
I am glad I Googled and found this question, I also would like this very much.
It seems there isn't a PR for it. @onedr0p care to create a PR for it?
Otherwise I could give a go at writing go

onedr0p:
[...]

PR #879, opened 2024-10-17
feat(pushover): priority on resolved

onedr0p:
## Summary

The logic can be handle a number of different ways, please advise if this solution is not ideal. Thanks!

https://github.com/TwiN/gatus/issues/767

## Checklist


README.md
```diff
@@ -111,6 +111,7 @@ Have any feedback or questions? [Create a discussion](https://github.com/TwiN/ga
   - [Configuring a startup delay](#configuring-a-startup-delay)
   - [Keeping your configuration small](#keeping-your-configuration-small)
   - [Proxy client configuration](#proxy-client-configuration)
+  - [Proxy client configuration](#proxy-client-configuration-1)
```
onedr0p, line 114:
```suggestion
```

README.md
```diff
@@ -73,8 +73,8 @@ Have any feedback or questions? [Create a discussion](https://github.com/TwiN/ga
     - [Configuring Twilio alerts](#configuring-twilio-alerts)
     - [Configuring AWS SES alerts](#configuring-aws-ses-alerts)
     - [Configuring custom alerts](#configuring-custom-alerts)
-    - [Configuring Zulip alerts](#configuring-zulip-alerts)
     - [Setting a default alert](#setting-a-default-alert)
+    - [Configuring Zulip alerts](#configuring-zulip-alerts)
```
onedr0p, line 77:
```suggestion
    - [Configuring Zulip alerts](#configuring-zulip-alerts)
    - [Setting a default alert](#setting-a-default-alert)
```

onedr0p:
I'm also fine with changing `priority-on-resolved` to `resolved-priority` if that's desirable.

alerting/provider/pushover/pushover.go
```diff
@@ -13,8 +13,9 @@ import (
 )
 
 const (
-	restAPIURL      = "https://api.pushover.net/1/messages.json"
-	defaultPriority = 0
+	restAPIURL                = "https://api.pushover.net/1/messages.json"
+	defaultPriority           = 0
+	defaultPriorityOnResolved = defaultPriority
```
TwiN, line 18: Instead of creating a separate constant for defaultPriorityOnResolved that uses defaultPriority anyways, let's just use defaultPriority

onedr0p: Done!

onedr0p:
I went ahead and renamed this setting to `ResolvedPriority`, hope that's okay!

alerting/provider/pushover/pushover.go
```diff
@@ -34,6 +34,10 @@ type AlertProvider struct {
 	// default: 0
 	Priority int `yaml:"priority,omitempty"`
 
+	// Priority of all messages, ranging from -2 (very low) to 2 (Emergency)
```
onedr0p, line 37:
```suggestion
	// Priority of resolved messages, ranging from -2 (very low) to 2 (Emergency)
```

alerting/provider/pushover/pushover.go
```diff
@@ -34,6 +34,10 @@ type AlertProvider struct {
 	// default: 0
 	Priority int `yaml:"priority,omitempty"`
 
+	// Priority of resolved messages, ranging from -2 (very low) to 2 (Emergency)
+	// default: Priority
```
onedr0p, line 38: Should we default 0 here?

TwiN: What's the default if you don't specify it in pushover? If it's 0, then yes, we should use 0

TwiN: Isn't 0 already the default?

onedr0p: Yes but my worry is that this is kind of a breaking change. Meaning previously if someone set the priority to 2, the resolved priority would be the same. With this PR the priority would still be 2 but the resolved priority would be 0 unless the person updated the resolved priority to be 2 as well.

onedr0p: @TwiN are we fine with this being a "breaking" change and just assigning `resolved-priority` to a default of `0` instead of trying to hack in `priority == resolved-priority` if `resolved-priority` is not set?

TwiN: Ah, I see your point. I don't think the priority aspect of pushover is worth worrying about as far as breaking changes go, especially since this is just the priority for the resolve notification. I think it's fine if we have it default to 0, despite the fact that it technically changes the previous behavior.

onedr0p: Alrighty, everything should be good to go then

TwiN:
Looks like you forgot to update the tests

TwiN (changes requested):
Looks like you forgot to update the tests

onedr0p:
@TwiN Updated, I added a condition for resolved-priority too.

TwiN (approved):
@onedr0p Thank you for the contribution!

Issue #700, opened 2024-03-07
How many endpoint-defaults setting can be set ? only one setting ?


tarocjsu:
### Describe the feature request

I had been use one endpoint-defaults for default endpoints setting, but look like just can use one endpoint-defaults setting ? has any chance to use more than one endpoint-defaults settings ? Thanks.

endpoint-defaults: &defaults
  group: core
  interval: 30s
  client:
    timeout: 10s
  conditions:
    - "[STATUS] == 200"
    - "[CERTIFICATE_EXPIRATION] > 48h"
  alerts:
    - type: discord
      failure-threshold: 5
    - type: discord
      failure-threshold: 10
    - type: discord
      failure-threshold: 15
    - type: telegram
      failure-threshold: 5
    - type: telegram
      failure-threshold: 10
    - type: telegram
      failure-threshold: 15
    - type: email
      failure-threshold: 5
    - type: email
      failure-threshold: 10
    - type: email
      failure-threshold: 15

endpoints:
  - name: npm
    <<: *defaults
    url: "https://npm.xxxx.eu.org/"

### Why do you personally want this feature to be implemented?

More flexable for config.yaml setting.

TwiN:
You can use as many as you want, you just have to name them differently.

e.g. 
```yaml
default-graphql-endpoint: &default-graphql-endpoint
```

I suggest you look up "yaml anchors"

PR #899, opened 2024-11-16
docs: Improve YAML anchor documentation

pixitha:
Add example for using multiple anchors in the same config, spent a few minutes trying to get this right, seems like it would be nice to save some people time by adding it right into the main readme.

## Summary

#700 

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -2188,7 +2188,7 @@ If, for any reason, you need Gatus to wait for a given amount of time before mon
 
 ### Keeping your configuration small
 While not specific to Gatus, you can leverage YAML anchors to create a default configuration.
-If you have a large configuration file, this should help you keep things clean.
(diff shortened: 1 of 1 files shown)


TwiN:
While I appreciate the idea, I'm not sure this is really worth it as documentation about anchor is really just a Google search away 🤔

pixitha:
I figure, it would have saved me time and effort, might as well help the next person. Having never used yaml anchors before, it was harder than I expected to find the simple answer/example in the right context.

TwiN:
After some more thoughts, I decided that this wasn't really adding much to the documentation. 

I do sincerely appreciate the effort though - I'm sure others will find this PR if they search for more information on YAML anchors in Gatus - but I feel the update you suggested to the docs risk confusing more people than helping them, which would be counter-productive.

Either way, thank you for contributing!

PR #920, opened 2024-11-30
fix(alerting): Telegram overrides not working as expected

### LINKED ISSUE #919: alerting.telegram.overrides seems to be broken
### Describe the bug

When i add alerting.telegram.overrides, the provider configuration becomes invalid

```
    overrides:
      - group: "something" // redacted
        id: "1100" // redacted
```

Remove the entire override, and it becomes functional again

### What do you see?

2024/11/30 16:33:36 [config.validateAlertingConfig] Ignoring provider=telegram because configuration is invalid

working provider

1. add overrides with either token or id or both to telegram provider, it breaks.

### Version

latest tag on docker.

### Additional information

It would be also helpful if the logs said why it is invalid, instead of a blind mention of this config is invalid, thats neither helpful, and wastes a ton of time with trial and error.

baalajimaestro:
Overrides had unexported variables leading to forever invalid configuration

## Summary
Fixes #919 

## Checklist


alerting/provider/telegram/telegram.go
```diff
@@ -32,29 +32,29 @@ type AlertProvider struct {
...
+        if len(override.Group) == 0 {
+            return false
+        }
+        if _, ok := registerGroups[override.Group]; ok {
+            return false
+        }
+        registerGroups[override.Group] = true
+    }
+
+    return len(provider.Token) > 0 && len(provider.ID) > 0
```
TwiN, line 57: Can you undo this change?

TwiN (changes requested):

alerting/provider/telegram/telegram.go
baalajimaestro: The spaces?

TwiN: Yes, indentation in Go is with tabs, not spaces. It was previously indented with tabs, but your IDE must've automatically changed it.

TwiN:
FYI #929 should fix this issue

PR #933, opened 2024-12-14
docs: Change protocol backslash (\\) to forwardslash (//) in README.md

PhilipWhiteside:
Some of the text examples used `proto:\\` instead of `proto://`. The backslash is not correct and results in "panic: invalid endpoint testgroup-|-testendpoint: unknown endpoint type".

## Summary

NA

## Checklist


TwiN (approved):

TwiN:
Good catch

PR #941, opened 2024-12-20
fix(alerting): replace HTML entities with emojis in teams-workflows provider to prevent rendering issues

pieceofsoul:
## Summary

Emojis written as HTML entities are not working as expected. Especially on mobilde devices (android) the entity is not rendered. Desktop seems to work. The assumption is that there is no full support of html entities for adaptive cards: https://learn.microsoft.com/en-us/microsoftteams/platform/task-modules-and-cards/cards/cards-format?tabs=adaptive-md%2Cdesktop%2Cdesktop1%2Cdesktop2%2Cconnector-html

However, replacing them with emojis itself, works on desktop and mobile devices.

## Checklist

Verified by running gatus locally with a local config. Desktop and mobile view are both showing the emojis in an appropriate way.


TwiN (dismissed):

TwiN (changes requested):
You'll have to update the tests as well

pieceofsoul:
Apologies for the oversight. I've updated the tests, and they are now passing as expected.

TwiN (approved):

PR #945, opened 2024-12-24
feat(alerting): Add condition results to Pushover

### LINKED ISSUE #787: Add condition result to pushover alerts
#### Current

I currently use Pushover to receive alerts on my phone if something goes down. That alert contains the description I can provide with the Pushover config.

#### Desired

I just saw [here](https://github.com/TwiN/gatus/blob/master/alerting/provider/discord/discord.go#L98) in the Discord alert implementation that it is possible to send the condition results inside the alert. I would love to have this for Pushover as well.

Some of the conditions are more important for me than others. I am for example notified about slow response times or soon to be expired certs. For me that does not require immediate action for me most of the time. On the other hand, if something is really down I want to know this so I can take immediate actions to bring the service back to life.

I think around a month

### Additional information

With some guidance I would offer to add this myself, but to be honest I never wrote a single line of GO yet.

mmed:
## Summary
- Includes conditions within messages
- Updates default title to `Gatus: <endpoint>`
- Enables [HTML](https://pushover.net/api#html) mode

Fixes #787

#### Notification
IMG_8084

#### In-App

IMG_8085

## Checklist


TwiN (approved):

TwiN:

Issue #803, opened 2024-06-16
MQTT Monitoring


zdzichu:
### Describe the feature request

Have you thought about adding MQTT monitoring? I'd like to be informed when specific topic stops receiving data (i.e. last publish was more than X minutes ago).

### Why do you personally want this feature to be implemented?

I have an IoT device which sometime hang and need to be manually reset.

### How long have you been using this project?

3 years

TwiN:
This is my first time hearing about MQTT, but sure! In the best of worlds, Gatus would support all protocols.

kadaan:
#596 seems like an idea way to add support for things like MQTT.  If the adoption of a custom check grows, then it could be an ideal candidate to fold into the core Gatus codebase.

eerison:
hey @zdzichu 

just to know, do you want to monitor zigbee devices?

zdzichu:
@eerison nope, I have no zigbee devices. Just couple ESPEasy controllers and scripts publishing to common MQTT broker.
I was looking for generic MQTT alerting in Gatus.

PR #950, opened 2024-12-31
Mqtt support

kadaan:
## Summary
#803 - Add support for MQTT testing

MQTT Endpoints are similar to DNS Endpoints in that they are recognized by the presence of an `mqtt` config block.  This enables the endpoint to be configured with the required topic and the optional username/password.  The check will subscribe to the topic and wait for a message to come in that contains a payload that matches the BODY.  While the check is waiting for incoming messages it will public a message containing the BODY as the payload.  If the subscription gets the message containing the body= before the timeout publish, CONNECTED will be set to true.  Otherwise, it will be false. The body can also be a text/template with support for a `uuidv4` function.  This enables the body to be specified like `gatus check - {{ uuidv4 }}` and ensures that the subscription is getting the message that was just published.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,7 @@ Have any feedback or questions? [Create a discussion](https://github.com/TwiN/ga
   - [Monitoring a WebSocket endpoint](#monitoring-a-websocket-endpoint)
   - [Monitoring an endpoint using ICMP](#monitoring-an-endpoint-using-icmp)
   - [Monitoring an endpoint using DNS queries](#monitoring-an-endpoint-using-dns-queries)
+  - [Monitoring an endpoint using MQTT](#monitoring-an-endpoint-using-mqtt)
   - [Monitoring an endpoint using SSH](#monitoring-an-endpoint-using-ssh)
   - [Monitoring an endpoint using STARTTLS](#monitoring-an-endpoint-using-starttls)
   - [Monitoring an endpoint using TLS](#monitoring-an-endpoint-using-tls)
@@ -2035,6 +2036,28 @@ There are two placeholders that can be used in the conditions for endpoints of t
 - The placeholder `[BODY]` resolves to the output of the query. For instance, a query of type `A` would return an IPv4.
 - The placeholder `[DNS_RCODE]` resolves to the name associated to the response code returned by the query, such as
 `NOERROR`, `FORMERR`, `SERVFAIL`, `NXDOMAIN`, etc.
+
+
+### Monitoring an endpoint using MQTT
+Defining an `mqtt` configuration in an endpoint will automatically mark said endpoint as an endpoint of type MQTT:
+```yaml
+endpoints:
+  - name: example-mqtt-query
+    url: "wss://example.com/mqtt"
+    mqtt:
+      topic: "my_topic"
+      username: "username" # Optional
+      password: "password" # Optional
+    body: "gatus check - {{ utcEpoch }}"
+    conditions:
+      - "[CONNECTED] == true"
+```
+
+The body can be plain text or a text/template.  If it is a text/template, the following functions are available:
+- `utcEpoch` returns the seconds since the unix epoch
+
+The following placeholders are supported for endpoints of type MQTT:
+- `[CONNECTED]` resolves to `true` if the MQTT message was published and the same message was consumed, `false` otherwise
 
 
 ### Monitoring an endpoint using SSH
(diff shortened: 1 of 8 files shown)


zdzichu:
Hey, thanks for implementing this! I gave it a spin, and while basic functionality seems to work, I have some questions.

First:
> `[CONNECTED]` resolves to `true` if the MQTT message was published and the same message was consumed, `false` otherwise`

It sounds like this is like "ping" service, checking if MQTT broker responds? Or possibly expecting a known value on configured topic? That would monitor the broker functionality itself, not the values published by other clients.

Second, I'm not sure how to configure for alerting for usecases mentioned in the original issue:
- alert when no message was published to given topic for 5 minutes
- alert when value published to given topic is larger than 10

Could you guide me?

kadaan:
Yeah, what I implemented is a ping to ensure that the basic MQTT infrastructure is up and running.  You are correct that it will not monitor the values published by other clients.  Given that, I don't think those kind of alerts can be created.  

The cases you mention seem to be quite different from how Gatus does other checks.  They would require a persistent checker that would be subscribing from MQTT and exposing a metric.  This is more in line with how a Prometheus exporter works.  I believe that external endpoints or #596 would be a better fit for those checks.

kadaan:
@TwiN Can you review?

client/client.go
```diff
@@ -32,6 +36,12 @@ var (
 
 	whoisClient              = whois.NewClient().WithReferralCache(true)
 	whoisExpirationDateCache = gocache.NewCache().WithMaxSize(10000).WithDefaultTTL(24 * time.Hour)
+
+	mqttTemplateEngine = template.New("base").Funcs(template.FuncMap{
```
TwiN, line 40: The only think I'm unsure about is the fact that this uses a templating engine, while everything else leverages placeholder and functions. Maybe this should be exposed as a placeholder instead? I feel supporting `[UUID4]` globally (and not just for MQTT) would be more beneficial, albeit more complicated.

The UUID4 placeholder should also be unique per endpoint, meaning that if it's in the url and the body, it should be the same. Let me know if that doesn't make sense, I can give you some examples

config/endpoint/endpoint_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"errors"
+	"github.com/TwiN/gatus/v5/config/endpoint/mqtt"
```
TwiN, line 8: This should be separate from the standard library dependencies, namely, at the bottom. See other go files for example.

TwiN (changes requested):

client/client.go
kadaan: Makes sense. I will look around and see if I can just figure it out.

config/endpoint/endpoint_test.go
kadaan: 👍

TwiN: Let me know/ping me when you resolved it 👀

README.md
```diff
@@ -2037,6 +2038,28 @@ There are two placeholders that can be used in the conditions for endpoints of t
 `NOERROR`, `FORMERR`, `SERVFAIL`, `NXDOMAIN`, etc.
 
 
+### Monitoring an endpoint using MQTT
+Defining an `mqtt` configuration in an endpoint will automatically mark said endpoint as an endpoint of type MQTT:
+```yaml
+endpoints:
+  - name: example-mqtt-query
+    url: "wss://example.com/mqtt"
```
TwiN, line 2046: why are you using a websocket endpoint for monitoring mqtt? 🤔 
shouldn't we be using mqtt?

TwiN: wouldn't this cause confusion, given that a websocket endpoint may be defined as a websocket endpoint internally? See the `TypeMQTT     Type = "MQTT"` change you made

client/client.go
kadaan: Haven’t had time to look into this yet

README.md
kadaan: Websocket is a valid transport for MQTT: https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718127

config/endpoint/endpoint_test.go
TwiN: Any news?

README.md
TwiN: Sounds good to me.

PR #956, opened 2025-01-07
feat(ssh): Support authless SSH health check

### LINKED ISSUE #723: fix(ssh): Support authless connection

A mode of the SSH check that would only check for a valid SSH connection banner. It's an upgrade from a plain TCP check but lighter weight than a full SSH check as done today.

An example of a borked OpenSSH server that passes a TCP check:
```
❯ nc borked.example.com 22
EHLO
```
... After which the connection terminates.

An example of a functioning OpenSSH server:
```
❯ nc fine.example.com 22
SSH-2.0-OpenSSH_9.2p1 Debian-2+deb12u2
EHLO
Invalid SSH identification string.
```
... After which the connection terminates.

While the current SSH check does allow checking an SSH endpoint it has limitations that I'd love to overcome:
- It can only validate with authentication -> Requires exposing credentials to monitoring, even if it can be done via env variables.
- It requires password auth -> No support for pubkey auth, requires keeping password auth enabled in sshd which is against security best practices.

1y

### Additional information

The banner check is just the first idea I was able to confirm in a real scenario. If someone knows a bit more of the protocol then perhaps there's more quick wins a bit deeper. But, then again a quick "does the server send *any* banner after TCP open would already be a big improvement.

ImTheCurse:
## Summary
- added support for no-auth SSH health check.
- added tests.

fixes #723

@TwiN mind checking this pr?

## Checklist


config/endpoint/ssh/ssh_test.go
```diff
@@ -7,11 +7,10 @@ import (
...
 func TestSSH_validate(t *testing.T) {
 	cfg := &Config{}
-	if err := cfg.Validate(); err == nil {
-		t.Error("expected an error")
-	} else if !errors.Is(err, ErrEndpointWithoutSSHUsername) {
-		t.Errorf("expected error to be '%v', got '%v'", ErrEndpointWithoutSSHUsername, err)
+	if err := cfg.Validate(); err != nil {
+		t.Error("didn't expect an error")
 	}
+
```
TwiN, line 13:
```suggestion
```

config/endpoint/ssh/ssh.go
```diff
@@ -19,6 +19,10 @@ type Config struct {
 
 // Validate the SSH configuration
 func (cfg *Config) Validate() error {
+	if len(cfg.Username) == 0 && len(cfg.Password) == 0 {
+		return nil
+	}
+
```
TwiN, line 25:
```suggestion
```

config/endpoint/endpoint.go
```diff
@@ -363,6 +363,19 @@ func (e *Endpoint) call(result *Result) {
 		}
 		result.Duration = time.Since(startTime)
 	} else if endpointType == TypeSSH {
+		if len(e.SSHConfig.Username) == 0 && len(e.SSHConfig.Password) == 0 {
+			result.Connected, result.HTTPStatus, err =
+				client.CheckSSHBanner(strings.TrimPrefix(e.URL, "ssh://"), e.ClientConfig)
+
```
TwiN, line 369:
```suggestion
```

config/endpoint/endpoint.go
```diff
@@ -363,6 +363,19 @@ func (e *Endpoint) call(result *Result) {
...
+			if err != nil {
+				result.AddError(err.Error())
+				return
+			}
+			result.Success = result.Connected
+			result.Duration = time.Since(startTime)
+			return
+		}
+
```
TwiN, line 378:
```suggestion
```

client/client.go
```diff
@@ -195,6 +196,42 @@ func CanCreateSSHConnection(address, username, password string, config *Config)
...
+	buf := make([]byte, 256)
+
+	_, err = io.ReadAtLeast(conn, buf, 1)
+
+	if err != nil {
+		return false, 1, err
+	}
+
+	return true, 0, err
+
```
TwiN, line 232:
```suggestion
	dialer := net.Dialer{}
	connStr := net.JoinHostPort(address, port)
	conn, err := dialer.Dial("tcp", connStr)
	if err != nil {
		return false, 1, err
	}
	defer conn.Close()
	conn.SetReadDeadline(time.Now().Add(time.Second))
	buf := make([]byte, 256)
	_, err = io.ReadAtLeast(conn, buf, 1)
	if err != nil {
		return false, 1, err
	}
	return true, 0, err
```

client/client.go
```diff
@@ -195,6 +196,42 @@ func CanCreateSSHConnection(address, username, password string, config *Config)
 	return true, cli, nil
 }
 
+func CheckSSHBanner(address string, cfg *Config) (bool, int, error) {
+	var port string
+	if strings.Contains(address, ":") {
+		addressAndPort := strings.Split(address, ":")
+		if len(addressAndPort) != 2 {
+			return false, 1, errors.New("invalid address for ssh, format must be host:port")
```
TwiN, line 204:
```suggestion
			return false, 1, errors.New("invalid address for ssh, format must be host:port")
```
I believe the format should be `ssh://host:port`, not just `host:port`, but please double check when you have a moment

TwiN (changes requested):

client/client.go
ImTheCurse: Yeah, originally I did that because it was already trimmed, but I'll change it.

ImTheCurse:
@TwiN if you have the time, would you mine checking on the PR? I've accidentally worked on the main branch and want to work on a different feature, and there is no way to change the source branch.

config/endpoint/endpoint.go
```diff
@@ -363,6 +363,17 @@ func (e *Endpoint) call(result *Result) {
 		}
 		result.Duration = time.Since(startTime)
 	} else if endpointType == TypeSSH {
+		if len(e.SSHConfig.Username) == 0 && len(e.SSHConfig.Password) == 0 {
```
TwiN, line 366: Can you add a comment below this line? 
Something like `// If there's no username/password specified, attempt to validate just the SSH banner`

(I'm on mobile, otherwise I'd have made the change myself)

config/endpoint/ssh/ssh.go
```diff
@@ -19,6 +19,9 @@ type Config struct {
 
 // Validate the SSH configuration
 func (cfg *Config) Validate() error {
+	if len(cfg.Username) == 0 && len(cfg.Password) == 0 {
+		return nil
```
TwiN, line 23: Can you add a comment above this line: `// If there's no username and password, this endpoint can still check the SSH banner, so the endpoint is still valid`

TwiN (changes requested):

ImTheCurse:
@TwiN I've added the changes

TwiN (approved):

TwiN:
Thank you for your contribution @ImTheCurse

PR #973, opened 2025-01-22
Trigger ondemand monitoring check for an endpoint

arvindiyengar:
## Summary

Adding a feature where in the user can trigger an adhoc monitoring check to Gatus. 

## Checklist


TwiN:
I can't allow this feature as-is, nor do I plan to allow a feature like this for the meantime, especially not without some form of authentication.

Main reason being, if I allowed this in, people could effectively DoS the underlying endpoints, even if they're configured to be private.

Issue #911, opened 2024-11-21
`maintenance.timezone` does not seem to work


simonhammes:
### Describe the bug

I'm using the offical gatus Docker image. I've set the `TZ` environment variable to `Europe/Berlin`.

This is my maintenance configuration:

```yml
maintenance:
  enabled: true
  start: 10:45
  duration: 1h
  timezone: "Europe/Berlin"
```

Is this a misunderstanding on my side?

### What do you see?

Example: Gatus thinks it is in the maintenance period if the local time (Europe/Berlin) is 12:03. There seems to be a 1 hour offset.

From the container logs (using `GATUS_LOG_LEVEL=DEBUG`): 

```
2024/11/21 12:03:02 [watchdog.execute] Not handling alerting because currently in the maintenance window
```

### What do you expect to see?

Gatus correctly detects if it should switch to maintenance mode.

### List the steps that must be taken to reproduce this issue

1. Start the application
2. Look at the container logs

### Version

`latest` from Docker Hub

simonhammes:
@alexmaras Thank you for fixing this :)

alexmaras:
You're welcome!

PR #981, opened 2025-01-28
fix(maintenance): timezone handling offset issue

alexmaras:
## Summary

Fixes an issue with the timezone handling in Maintenance blocks
https://github.com/TwiN/gatus/issues/911

The problem was with using Truncate, which always operates in UTC. Switching to constructing the date from component parts (`0, 0, 0, 0` for midnight) fixes this issue and makes the time always midnight for the timezone the maintenance window refers to.

I've added a test for this too - the previous test didn't catch the problem because it was only an hour out from UTC, and the window covered a 2 hour period. I've using UTC+8 (Australia/Perth) for a large enough gap to catch the issue. This test fails without the code changes, and passes now.

## Checklist

- [x] Updated documentation in `README.md`, if applicable. (Not applicable - bugfixing existing behavior)

### DIFF
--- a/config/maintenance/maintenance.go
+++ b/config/maintenance/maintenance.go
@@ -111,7 +111,6 @@ func (c *Config) IsUnderMaintenance() bool {
 		now = now.In(c.TimezoneLocation)
 	}
 	var dayWhereMaintenancePeriodWouldStart time.Time
-
 	adjustedDate := now.Day()
 	if now.Hour() < int(c.durationToStartFromMidnight.Hours()) {
 		// if time in maintenance window is later than now, treat it as yesterday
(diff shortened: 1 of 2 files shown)


alexmaras:
(just a heads-up - fairly new to golang, so let me know if there's any more go-typical ways of handling things)

config/maintenance/maintenance.go
```diff
@@ -111,11 +111,15 @@ func (c *Config) IsUnderMaintenance() bool {
...
-		dayWhereMaintenancePeriodWouldStart = now.Add(-c.Duration).Truncate(24 * time.Hour)
+
+	adjustedDate := now.Day()
+	if now.Hour() < int(c.durationToStartFromMidnight.Hours()) {
+		// if time in maintenance window is later than now, treat it as yesterday
+		adjustedDate--
 	}
+	// Set to midnight prior to adding duration
+	dayWhereMaintenancePeriodWouldStart = time.Date(now.Year(), now.Month(), adjustedDate, 0, 0, 0, 0, now.Location())
+
```
TwiN, line 122: Can you remove the blank lines?

TwiN (changes requested):
Excellent catch! Just a small nitpick with the newlines, but looks perfect otherwise!

config/maintenance/maintenance.go
alexmaras: Done :+1:

alexmaras:
I also just realised I can remove the separate variable definition for `dayWhereMaintenancePeriodWouldStart` as it's not being set in the `if/else` now.

TwiN (approved):

TwiN:
@alexmaras Thank you for the contribution!

Issue #696, opened 2024-03-04
Yaml parse error when volume are mounted via configmap or secret on kubernetes


jerome-karabenli:
### Describe the bug

When I add env GATUS_CONFIG_PATH=/config as described in doc, this env point on directory, gatus tries to recursively read all yaml or yml files but in kubernetes when we mount a volume inside a container and this volume is provided by configMap or Secret Kubernetes mount a kind of symlink on the mountPath, the real data is on the same folder but in hidden folder with date as name. This is probably to have latest date of change of configMap or Secret, because if we change/update the values from configMap or Secret theses data are updated after a few seconds inside a container in pod. 

### What do you see?

2024/03/04 14:58:22 [config][LoadConfiguration] **Reading configuration from /config/..2024_03_04_14_58_14.4259295514/config.yaml**
2024/03/04 14:58:22 [config][LoadConfiguration] **Reading configuration from /config/config.yaml**
panic: error reading configuration from directory /config: error due to parameter with value of primitive type: only maps and slices/arrays can be merged, which means you cannot have define the same key twice for parameters that are not maps or slices/arrays

goroutine 1 [running]:
main.main()
	/app/main.go:19 +0x1b1

### What do you expect to see?

2024/03/04 15:00:04 [config][LoadConfiguration] **Reading configuration from configFile=/config/config.yaml**
2024/03/04 15:00:04 [config][validateAlertingConfig] Alerting is not configured
2024/03/04 15:00:04 [config][validateEndpointsConfig] Validated 7 endpoints
2024/03/04 15:00:04 [controller][Handle] Listening on 0.0.0.0:8080

### List the steps that must be taken to reproduce this issue

add env in kubernetes deployment GATUS_CONFIG_PATH=/config

### Version

twinproduction/gatus:v5.7.0

### Additional information

if I could I would have proposed a PR but I don't know how to code in Go. 
The problem is probably because gatus tries to read all yaml files recursively in specified folder by env. 

Some improvement may be to not just read all yaml files inside specified folder but keep the recursive read because there is a security improvement to be achieved by splitting security configs, alerting configs and endpoints configs. 

For exemple in Kubernetes we can imagine to mount a secret as volume in /config/security/security.yaml and an alerting config which contains credentials from another secret on /config/alerting/alerting.yaml and developers can add monitoring endpoints from configMap on /config/endpoints/endpoints.yaml 

[...]

PR #984, opened 2025-01-29
fix(config): Ignore config files with double dot in it (#696)

vlaborie:
## Summary

Ignore config files with double dot in it, permit to mount Kubernetes ConfigMap as directory and ignore hidden files like `..date/filename.yaml`.

## Checklist

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -176,6 +176,10 @@ func LoadConfiguration(configPath string) (*Config, error) {
 			if err != nil {
 				return fmt.Errorf("error walking path %s: %w", path, err)
 			}
+			if strings.Contains(path, "..") {
+				logr.Warnf("[config.LoadConfiguration] Ignoring configuration from %s", path)
+				return nil
+			}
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -176,6 +176,10 @@ func LoadConfiguration(configPath string) (*Config, error) {
 			if err != nil {
 				return fmt.Errorf("error walking path %s: %w", path, err)
 			}
+			if strings.Contains(path, "..") {
+				logr.Warnf("[config.LoadConfiguration] Ignore configuration from %s", path)
```
TwiN, line 180: Can you update the log message to say "Ignoring" instead of ignore? (I'm on mobile and my laptop is currently out for repair, I'd have made the change myself otherwise)

TwiN (changes requested):

config/config.go
vlaborie: I have update the message

TwiN (approved):

PR #836, opened 2024-08-09
feat(storage): Support 30d badges

TwiN:
## Summary

Add support for 30d uptime badges, response time badges and response time chart.

Resolves #128
Resolves #714

## Checklist

- [X] Tested and/or added tests to validate that the changes work as intended, if applicable.
- [x] Updated documentation in `README.md`, if applicable.

PR #994, opened 2025-02-05
feat(ui): expose 30d badges and response chart selector for duration

toredash:
## Summary
This PR will expose the 30d badges in the endpoints details view, it may be confusing to not find it in the details view, as reported [here](https://github.com/TwiN/gatus/pull/836#issuecomment-2456534480).

Additionally, PR exposes a new dropdown menu in the endpoints details view to allow the user to select a different response time chart period. I'm a little unsure about the styling of it, but the end-result looks like this:
image

I noticed that there was no support for 1h response time chart. I added code for it with an additional test. If you want this present but with a dedicated PR please let me know.

## Checklist


web/app/src/views/Details.vue
```diff
@@ -35,10 +39,20 @@
...
-      <h1 class="text-xl xl:text-3xl font-mono text-gray-400">RESPONSE TIME</h1>
-      <hr/>
-      <img :src="generateResponseTimeChartImageURL()" alt="response time chart" class="mt-6"/>
+      <div class="flex items-center justify-between">
+        <h1 class="text-xl xl:text-3xl font-mono text-gray-400">RESPONSE TIME</h1>
+          <option value="1h">1 hour</option>
+          <option value="24h">24 hours</option>
+          <option value="7d">7 days</option>
+          <option value="30d">30 days</option>
+        </select>
```
TwiN, line 48: You're missing a `<select>` block here.

TwiN (changes requested):
I tried to build it locally and I got this:
 
```
VueCompilerError: Invalid end tag.
at <secret>/gatus/web/app/src/views/Details.vue:48:9
46 |            <option value="7d">7 days</option>
47 |            <option value="30d">30 days</option>
48 |          </select>
   |          ^
49 |        </div>
50 |        <img :src="generateResponseTimeChartImageURL(selectedChartDuration)" alt="response time chart" class="mt-6" />
```

README.md
```diff
@@ -2315,7 +2316,7 @@ The path to generate a badge is the following:
 /api/v1/endpoints/{key}/uptimes/{duration}/badge.svg
 ```
 Where:
-- `{duration}` is `30d` (alpha), `7d`, `24h` or `1h`
+- `{duration}` is `30d`, `7d`, `24h` or `1h`
```
TwiN, line 2319: Fair, the 30d badges has been around for a bit now and it's fair to stay it's no longer in alpha.

api/chart.go
```diff
@@ -42,8 +42,10 @@ func ResponseTimeChart(c *fiber.Ctx) error {
 		from = time.Now().Truncate(time.Hour).Add(-7 * 24 * time.Hour)
 	case "24h":
 		from = time.Now().Truncate(time.Hour).Add(-24 * time.Hour)
+	case "1h":
+		from = time.Now().Add(-2 * time.Hour) // Because response time metrics are stored by hour, we have to cheat a little
 	default:
-		return c.Status(400).SendString("Durations supported: 30d, 7d, 24h")
+		return c.Status(400).SendString("Durations supported: 30d, 7d, 24h, 1h")
```
TwiN, line 48: This doesn't really make sense, no? Because a chart with only 1h only has a single data point, which makes it, well, not really much of a chart.

I think it'd make more sense if we just supported 30d, 7d and 24h

TwiN (changes requested):

api/chart.go
toredash: Agree. Removed this in the latest commit.

TwiN (approved):
Excellent work! Thank you for contributing @toredash!

PR #995, opened 2025-02-06
feat(template): support templating

### LINKED ISSUE #593: Template manifest.json
I want to template the manifest.json, which contains the information needed for PWA, so I can customize some appereance of my gatus deployment.

Because the file is json we can not use javascript to template this like the index.html, or am I missing something? 
My current idea is to use a templating engine. 
Is there any good templating engine for go, like jinja for python?

I would provide a pull request, if we find a nice solution.

Could be helpful for #174.

### LINKED ISSUE #709: Support go templating in alerting

Would be nice instead of place holders if we could support full go templating in the alert providers if not all but at least the custom provider.

So we can do conditional logic and provide full templating.

https://pkg.go.dev/text/template

All I want from our slack alert is to be able to include the URL for instance if status code is 500. Or I could template out the response body into the alert message.

2 weeks

### Additional information

Would be willing to implement this myself.

FalcoSuessgott:
fixes #709 
fixes #593 

This PR adds support for Go templating including [sprout template-functions](https://sprout.atom.codes) the maintained fork of the popular [sprig project](http://masterminds.github.io/sprig/) (included by Helm, etc).

```yaml
endpoints:
{{ range $ep := (list "aws" "google" "azure") }}
  - name: {{ $ep | toUpper }}
    url: https://{{ $ep }}.org
    interval: 5s
    conditions:
      - "[STATUS] == 200"
{{end}}
```

would be rendered to 

```yaml
endpoints:
  - name: AWS
    url: https://aws.org
    interval: 5s
    conditions:
      - "[STATUS] == 200"
 - name: GOOGLE
    url: https://google.org
    interval: 5s
    conditions:
      - "[STATUS] == 200"
   - name: AZURE
    url: https://azure.org
    interval: 5s
    conditions:
      - "[STATUS] == 200"
```


FalcoSuessgott:
If you would like this feature, give me a hint and I can update the docs

TwiN:
I don't think this fixes either of the issues linked. One mentions manifest.json, which this doesn't impact, while the other mentions templating but in a way that would expose some informations via the alerts, which this doesn't do.

I'm very thankful for your contribution, but as I try to keep Gatus as minimal and readable as possible, I don't think this feature makes sense for now. I also think that Go templating is very difficult to read for those that aren't familiar with it (most common example of this is Helm templates)

Again though, thank you for attempting to contribute; it's very appreciated!

PR #1003, opened 2025-02-14
feat: add MySQL storage support

### LINKED ISSUE #283: Add support for MariaDB/MySQL

sqlite and postgres are both available as storage engines.  In some environments, where a MySQL or MariaDB server already exists, it would be convenient to use a DB on this server.

I have an environment with MySQL/MariaDB :)

Currently evaluating

### Additional information

If there are no known technical blockers why it would not work, I'm happy to take a pass at implementing the support and send a PR.  It looks like it should be fairly doable, and would be in the same style as the postgres/sqlite split.

devhaozi:
## Summary
Add MySQL storage support, close #283

## Checklist


README.md
```diff
@@ -402,11 +402,11 @@ storage:
...
 See [examples/docker-compose-sqlite-storage](.examples/docker-compose-sqlite-storage) for an example.
 
-- If `storage.type` is `postgres`, `storage.path` must be the connection URL:
+- If `storage.type` is `postgres` or `mysql`, `storage.path` must be the connection URL:
 ```yaml
 storage:
-  type: postgres
-  path: "postgres://user:password@127.0.0.1:5432/gatus?sslmode=disable"
+  type: postgres # or mysql
+  path: "postgres://user:password@127.0.0.1:5432/gatus?sslmode=disable" # or user:password@tcp(127.0.0.1:3306)/gatus?charset=utf8mb4&parseTime=True&loc=Local&multiStatements=true&interpolateParams=true
```
TwiN, line 409: Could you make a config separate from this specifically for mysql? Instead of mixing the two in the same config

go.sum
```diff
@@ -6,6 +6,8 @@ cloud.google.com/go/compute/metadata v0.6.0 h1:A6hENjEsCDtC1k8byVsgwvVcioamEHvZ4
 cloud.google.com/go/compute/metadata v0.6.0/go.mod h1:FjyFAW1MW0C203CEOMDTu3Dk1FlqW3Rga40jzHL4hfg=
 code.gitea.io/sdk/gitea v0.19.0 h1:8I6s1s4RHgzxiPHhOQdgim1RWIRcr0LVMbHBjBFXq4Y=
 code.gitea.io/sdk/gitea v0.19.0/go.mod h1:IG9xZJoltDNeDSW0qiF2Vqx5orMWa7OhVWrjvrd5NpI=
+filippo.io/edwards25519 v1.1.0 h1:FNf4tywRC1HmFuKW5xopWpigGjJKiJSV0Cqo0cJWDaA=
+filippo.io/edwards25519 v1.1.0/go.mod h1:BxyFTGdWcka3PhytdK4V28tE5sGfRvvvRV7EaN4VDT4=
```
TwiN, line 10: 🤨

storage/store/sql/sql.go
```diff
@@ -404,8 +416,14 @@ func (s *Store) DeleteAllEndpointStatusesNotInKeys(keys []string) int {
 // GetTriggeredEndpointAlert returns whether the triggered alert for the specified endpoint as well as the necessary information to resolve it
 func (s *Store) GetTriggeredEndpointAlert(ep *endpoint.Endpoint, alert *alert.Alert) (exists bool, resolveKey string, numberOfSuccessesInARow int, err error) {
 	//logr.Debugf("[sql.GetTriggeredEndpointAlert] Getting triggered alert with checksum=%s for endpoint with key=%s", alert.Checksum(), ep.Key())
+	var query string
+	if s.driver == "mysql" {
+		query = `SELECT resolve_key, number_of_successes_in_a_row FROM endpoint_alerts_triggered WHERE endpoint_id = (SELECT endpoint_id FROM endpoints WHERE endpoint_key = ? LIMIT 1) AND configuration_checksum = ?`
+	} else {
+		query = `SELECT resolve_key, number_of_successes_in_a_row FROM endpoint_alerts_triggered WHERE endpoint_id = (SELECT endpoint_id FROM endpoints WHERE endpoint_key = $1 LIMIT 1) AND configuration_checksum = $2`
+	}
```
TwiN, line 424: I'm not really a fan of having those  if/else blocks for each SQL query. With just postgres and sqlite it was easy because they both have the same syntax, but now with mysql, I think we might need a different solution.

Does `?` work with sqlite and postgres as well? Or does postgres/sqlite require `$1`/`$2`/`$...`? Because in most cases, the query is exactly the same, outside of how prepared parameters are passed

devhaozi: > I'm not really a fan of having those if/else blocks for each SQL query. With just postgres and sqlite it was easy because they both have the same syntax, but now with mysql, I think we might need a different solution.
> 
> Does `?` work with sqlite and postgres as well? Or does postgres/sqlite require `$1`/`$2`/`$...`? Because in most cases, the query is exactly the same, outside of how prepared parameters are passed

Unfortunately, `?` is MySQL-specific, and `$...` is Postgres-specific. If necessary, I recommend using an ORM (e.g. gorm, xorm, ent) to smooth over these differences.

go.sum
devhaozi: > 🤨

This is added by the MySQL driver to support the client_ed25519 authentication method.

devhaozi:
Tested in my local environment and ready for review.

.examples/docker-compose-mysql-storage/docker-compose.yml
```diff
@@ -0,0 +1,34 @@
...
+
+  gatus:
+    image: twinproduction/gatus:latest
+    restart: always
+    ports:
+      - "8080:8080"
+    environment:
+      - POSTGRES_USER=username
+      - POSTGRES_PASSWORD=password
+      - POSTGRES_DB=gatus
```
TwiN, line 25: Are those a typo?

go.sum
TwiN: Not a big fan of people using their personal blogs as host, whether it be for security reasons or for availability reasons, but looking at https://github.com/FiloSottile/edwards25519 (and the user's profile), they appear to be trustworthy

storage/store/sql/sql.go
```diff
@@ -532,68 +584,124 @@ func (s *Store) Close() {
...
 func (s *Store) insertEndpointResult(tx *sql.Tx, endpointID int64, result *endpoint.Result) error {
 	var endpointResultID int64
-	err := tx.QueryRow(
-		`
+	if s.driver == "mysql" {
+		res, err := tx.Exec(
+			`
+			INSERT INTO endpoint_results (endpoint_id, success, errors, connected, status, dns_rcode, certificate_expiration, domain_expiration, hostname, ip, duration, timestamp)
+			VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+		`,
```
TwiN, line 640: Instead of having 2 `tx.Exec(...)` spanning 20ish lines, could you make an if/else for just the string portion of the query?

TwiN (changes requested):
Just a heads up, I consider this a pretty large change, and it'll take a few months before I have the time to test this properly.

storage/store/sql/sql.go
devhaozi: > Instead of having 2 `tx.Exec(...)` spanning 20ish lines, could you make an if/else for just the string portion of the query?

done.

.examples/docker-compose-mysql-storage/docker-compose.yml
devhaozi: > Are those a typo?

Yes, fixed.

TwiN:
I've given this a lot of thoughts, and I decided that supporting multiple storage providers is just too much work and it makes implementing new features even more difficult. Given that I'm doing this solo and not full time, I decided that I don't have enough time to support more storage types.

I do appreciate your contribution, though!

PR #1015, opened 2025-03-02
feat(ui): Allow configuring default dark-mode value

Xetera:
## Summary

Update from maintainer: I modified this to also enable setting dark-mode to true/false by default depending on preference. Note that OS appearance (dark/light) settings override the default `ui.dark-mode` value set, but if the user (not the dev -- the person visiting the status page) sets the appearance to either dark or light, that preference supersedes any Gatus configuration

---

This PR fixes a flash of flash of inaccurate color theme

The current use of `localStorage` causes an initial flicker of light mode when the user has explicitly switched to dark mode and reloads the page. Switching to cookies as storage allows the frontend to tell the backend the preferred theme so the server can respond with the correct `theme="dark"` class on the html element.

I do need some code review here though. Since the `ui` config is passed directly into the template in `spa.go`, the only way I could figure out how to add a dark mode adjuster (without including a bigger, likely out-of-scope change to how template data gets passed in) was by including `DarkMode` in the config struct without adding a yaml annotation.

Things seem to work fine in my testing but this approach kind of rubs me the wrong way and I think there might even be a race condition with multiple people loading the homepage? I'm not experienced enough with go to tell but I have a hunch. Let me know if there's a better way to handle this or if there should be a new struct passed into `t.Execute` that groups `Config` and non-config variables together.

## Checklist


api/spa.go
```diff
@@ -12,6 +12,8 @@ import (
 
 func SinglePageApplication(ui *ui.Config) fiber.Handler {
 	return func(c *fiber.Ctx) error {
+		theme := string(c.Request().Header.Cookie("theme"))
+		ui.DarkMode = theme == "dark"
```
TwiN, line 16: Because this is the field of a pointer, it might change the default `ui.DarkMode` for other users. It might even panic if there's concurrent modifications to the field, though I imagine this would be a rare occurrence. 

I'll push some adjustments on your fork

web/app/src/components/Settings.vue
```diff
@@ -48,15 +57,15 @@ export default {
       this.setRefreshInterval(this.$refs.refreshInterval.value);
     },
     toggleDarkMode() {
-      if (localStorage.theme === 'dark') {
-        localStorage.theme = 'light';
+      if (getTheme() === 'dark') {
+        document.cookie = `theme=light; path=/; max-age=31536000; samesite=strict`;
       } else {
-        localStorage.theme = 'dark';
+        document.cookie = `theme=dark; path=/; max-age=31536000; samesite=strict`;
```
TwiN, line 63: This will likely cause the first toggle to do nothing if you don't already have a cookie:

1. User loads page (no `theme` cookie)
2. Defaults to dark theme because `wantsDarkMode()` returns `true` if there's no theme
3. User clicks on theme (moon/sun) button
4. Because `getTheme()` doesn't return `dark`, it falls on the `else` block and a cookie with `theme=dark` is created, which means that the first click doesn't do anything
5. Second click will change from dark -> light because the cookie now exists

I'll push some code to fix this as well.

TwiN: cb3ebb7 should address this, but I don't have time to test this in-depth right now. Would you mind giving it a shot and letting me know? 🙏

TwiN (approved):

TwiN:
@Xetera Thank you for the contribution!

CirnoT:
This PR breaks detection logic entirely. Should user agent request light mode (`prefers-color-scheme: light`), they will be denied and instead dark mode will be presented to them (as per default `ui.dark-mode`).

Flipping configuration of `ui.dark-mode` to `false` and having user agent present `prefers-color-scheme: dark` does work properly though.

TwiN:
@CirnoT Personally, I wouldn't qualify a change in default theme as a breaking change, given that it doesn't actually break anything. Bottom line is that before, there was no way to set the default theme through the configuration, but now, there is.

I plan on modernizing the UI in a couple of months either way.

FWIW, based on my tests, I was pretty confident the behavior didn't change, but I may have missed something.

Issue #1022, opened 2025-03-09
Give Love to JSONPath implementation


meerumschlungen:
I'd like to tackle the comment in https://github.com/TwiN/gatus/blob/master/jsonpath/jsonpath.go#L10
`Eval is a half-baked json path implementation that needs some love`

Rough Plan:
1. Max out test coverage for the current implementation
2. Properly re-implement the jsonpath package ensuring future extendability and maintainability, ensuring the tests from 1. pass for the new implementation
3. Max out test coverage for the new implementation
4. Run new tests from 3. against the old implementation to ensure backward compatibility
5. Check out benchmarks
6. Be happy (hopefully)

I do not intend to change or add functionality in the first place with the reimplementation but am eager to do so in the future since I could use it well myself.
This should provide a solid foundation for solving e.g. #335 or #225.

@TwiN Please confirm nobody's working on this already.
Thank you for your great work so far!

TwiN:
Nobody's working on this right now! Though I'm a little curious about your intent. The `needs some love` part of `Eval is a half-baked json path implementation that needs some love` was more about implementing features that were missing from the current implementation, such as asterisks

meerumschlungen:
I suggest rewriting the codebase, improving path parsing with a robust parser and tokenizer, improving type handling for consistent formatting of all JSON types, structuring the code for reusable parsed paths, and strengthening error handling with detailed messages. I find it difficult to achieve these improvements with the current codebase, especially regarding future extensibility with the missing features you mention. In addition, I'd like to add more tests and try to optimize performance, for example by reducing string operations.
I have already put some work into this and it looks quite promising so far. When I'm done, I'll submit a pull request.

ImTheCurse:
> I suggest rewriting the codebase, improving path parsing with a robust parser and tokenizer, improving type handling for consistent formatting of all JSON types, structuring the code for reusable parsed paths, and strengthening error handling with detailed messages. I find it difficult to achieve these improvements with the current codebase, especially regarding future extensibility with the missing features you mention. In addition, I'd like to add more tests and try to optimize performance, for example by reducing string operations. I have already put some work into this and it looks quite promising so far. When I'm done, I'll submit a pull request.

Couldn't agree more, the JSONPath is a bit of a nightmare to maintain.
@meerumschlungen let me know when you finish the rewrite I might take a look if I can find the time.

meerumschlungen:
[...]

Issue #335, opened 2022-10-05
JSON Path support issue


AlexeyPirogov:
### Describe the bug

I'm using next JSON Path ```"[BODY].apps.app[?(@.id==2)].errors == 0"```
to verify next JSON
```
{
  "apps": {
    "app": [
      {
        "id": 1,
        "error": 0
      },
      {
        "id": 2,
        "error": 0
      }
    ]
  }
}
```

### What do you see?

I'm getting next error on UI after evaluation ```'- couldn't walk through 'app[?(@' because type was '<nil>', but expected 'map[string]interface{}'```.

### What do you expect to see?

Validation should pass.

### List the steps that must be taken to reproduce this issue

Start gatus, point to endpoint that returns JSON from description.

### Version

image: twinproduction/gatus, latest, 8e9d9f67f545 (can't get labels as on-prem Artifactory is used :( )

TwiN:
The `[?(@.id==2)]` instruction is not currently supported; to get the second element in the array, you'd have to use `[1]` (`[BODY].apps.app[1].errors == 0`).

I guess you're using `[?(@.id==2)]` instead because the order of the elements in the array is not always the same?

AlexeyPirogov:
Thx, for quick reply. Real json is more complicated, I just provided minimal example to reproduce the issue.
We query Hadoop cluster and in response I need to look for my app by 'name' field and verify status:
```
{
  "apps": {
    "app": [
      {
        ...,
        "name": "appA",
        ...,
        "state": "RUNNING",
        ...
      },
      {
        ...,
        "name": "appB",
        ...,
        "state": "RUNNING",
        ...
      }
    ]
  }
}
```

If we can't do this with JSON Path, do you support regex?
I think it possible to write regex with similar functionality, just trickier.

TwiN:
There's currently no support for regex, but you may be able to do something using the `pat` function (see https://github.com/TwiN/gatus#functions).

That said, I do like the idea of adding support for a basic filter (the condition would be something like `[BODY].apps.app[?(@.id==1)].errors == 0`).

In fact, it shouldn't be _too_ hard to implement. Somewhere around these lines: https://github.com/TwiN/gatus/blob/be88af5d48d9e2bb2797df2ca9ea28f18c693feb/jsonpath/jsonpath.go#L43-L53

Issue #225, opened 2021-12-14
Parsing JSON array


kmickeletto:
I am attempting to parse an array.
`{
  "servers": [
    {
      "servername": "server1",
      "serveractiveconn": 0,
      "servertotalconn": 20,
      "serverconnections": [],
      "serverssl": 1,
      "serversslinfo": "",
      "serveractive": true,
      "servererror": "",
      "serverpriority": 0,
      "serveroptional": 0,
      "serverbps": "0 "
    },
    {
      "servername": "server2",
      "serveractiveconn": 0,
      "servertotalconn": 60,
      "serverconnections": [],
      "serverssl": 1,
      "serversslinfo": "",
      "serveractive": true,
      "servererror": "",
      "serverpriority": 0,
      "serveroptional": 0,
      "serverbps": "0 "
    },
    {
      "servername": "server2",
      "serveractiveconn": 0,
      "servertotalconn": 50,
      "serverconnections": [],
      "serverssl": 1,
      "serversslinfo": "",
      "serveractive": true,
      "servererror": "",
      "serverpriority": 0,
      "serveroptional": 1,
      "serverbps": "0 "
    }
  ]
}`

What I was hoping to get was if any of .servers[].servererror contained a value, that would be a bad condition.
`len([BODY].status.servers[].servererror) == 0`

This did not work and the only way I have found to make it work is by doing a test for each index of the array.  Is there a better way?

Also, is there a better way to check that servererror is empty?

TwiN:
Unfortunately, recursive evaluation isn't currently supported.
As you've mentioned, the best way to do this is by manually checking each index of the array:
```
conditions:
  - "len([BODY].status.servers[0].servererror) == 0"
  - "len([BODY].status.servers[1].servererror) == 0"
  - "len([BODY].status.servers[2].servererror) == 0"
```
This is definitely subpar, especially if you have a dynamic number of servers, in which case I would recommend having an endpoint that allows retrieving a single server by name instead.

The alternative would be to add support for checking all values, which I think would be a good addition to Gatus:
```
conditions:
  - "len([BODY].status.servers[*].servererror) == 0"
```

TwiN:
@aarif I wouldn't mind implementing a `distinct(...)` function, but I'd rather avoid using a library that has a different syntax than the current implementation, as to avoid breaking changes.

PartTimeDataScientist:
> [@TwiN](https://github.com/TwiN) is this a direction you're willing to consider?
> 
[...]

PR #1024, opened 2025-03-10
chore(jsonpath): Improve jsonpath package

meerumschlungen:
## Summary

#1022 
This PR is a complete rewrite of the jsonpath package. It optimizes several aspects of the current implementation such as:
- Maintainability
- Future extensibility (see #335, #225)
- Type handling
- Consistent formatting
- Error handling

### Test driven approach
To ensure this reimplementation does not break backwards compatibility, I focused on redesigning the current behavior exactly. To accomplish this I first maxed out test coverage for the current implementation to 100%. After the reimplementation I made sure the reimplementation passed all tests that existed up to that point. Then I maxed out test coverage for the reimplementation to 91.8%. 100% coverage can't be achieved with the new version because the uncovered lines handle types JSON can't represent (i.e. default branches in type switching), making lines effectively unreachable with valid JSON input or because `tokenize` rejects malformed paths before `walk` is called and the tests can't bypass this.
To ensure backwards compatibility, I ran the new test suite against the old implementation. Some tweaks were needed in the new implementation to ensure the same behavior but this also revealed some irregularities in the old implementation, namely:
```txt
go test -cover ./jsonpath
--- FAIL: TestEval (0.00s)
    --- FAIL: TestEval/no-path-non-array (0.00s)
        jsonpath_test.go:406: Expected output length to be 15, but was 16
        jsonpath_test.go:409: Expected output to be {"key":"value"}, but was {"key": "value"}
    --- FAIL: TestEval/malformed-path-missing-close-bracket (0.00s)
        jsonpath_test.go:400: Expected error, got '<nil>'
        jsonpath_test.go:406: Expected output length to be 0, but was 2
        jsonpath_test.go:409: Expected output to be , but was [1 2]
    --- FAIL: TestEval/array-negative-index (0.00s)
panic: runtime error: index out of range [-1] [recovered]
        panic: runtime error: index out of range [-1]
    --- FAIL: TestEval/key-on-array (0.00s)
        jsonpath_test.go:400: Expected error, got '<nil>'
        jsonpath_test.go:406: Expected output length to be 0, but was 3
        jsonpath_test.go:409: Expected output to be , but was [1 2 3]
FAIL    github.com/TwiN/gatus/v5/jsonpath       0.220s
FAIL
```
- **TestEval/no-path-non-array**: This is trivial and does not need treatment.
- **TestEval/malformed-path-missing-close-bracket**: Invalid paths like "[0" don't fail but return a result.
- **TestEval/array-negative-index**: Negative indexes like "data[-1]" cause panic.
- **TestEval/key-on-array**: Evaluating a path segment on an array didn't fail but returns a result.

So the new implementation also improves robustness and fixes some edge-case bugs.

### Benchmarks
Old implementation:
`BenchmarkEval-8   	  304356	      3809 ns/op	    4761 B/op	      89 allocs/op`
New implementation:
`BenchmarkEval-8   	  353466	      3224 ns/op	    5544 B/op	      77 allocs/op`

Performance improvement: The new implementation is faster (15.4% lower latency, 16.2% higher throughput).
Memory trade-off: It uses 16.4% more memory per operation but with 13.5% fewer allocations.
The new version prioritizes speed and allocation efficiency over memory footprint, which I think is a good tradeoff for this use case.

## Checklist

- [ ] ~Updated documentation in `README.md`, if applicable.~


jsonpath/jsonpath.go
```diff
@@ -2,130 +2,220 @@ package jsonpath
...
-			array, _ := value.([]interface{})
-			if len(array) > arrayIndex {
-				if isNestedArray {
-					return extractValue(currentKey[endOfBracket+1:], array[arrayIndex])
+	}
+
+	switch current.Type {
+	case tokenKey:
+		if obj, ok := value.(map[string]any); ok {
+			if nextVal, exists := obj[current.Value]; exists {
```
ImTheCurse, line 142: ```nextVal``` might be a bit confusing since where are getting the current value(even if it is a part of the tree that we are iterating over), I suggest something along the lines of currPathVal.
might be a nit pick but I think it could improve readability.

ImTheCurse (changes requested):
Overall LGTM, just a bit of a nit pick, perhaps make the code a bit more modular - separate the different cases into helper functions

dennisoderwald:
@meerumschlungen Are you still actively working on this? @TwiN Is it realistic to achieve a merge here in the near future? I think it would also help with my problem #1238.

meerumschlungen:
I was basically done with the implementation as far as I remember. I use the changed code in my personal fork. It got stale then.
@TwiN Are you interested in this contribution and are committed to perform the review? Then I would take the time to clean up, resolve conflicts, fix tests if necessary and implement review suggestions.
If you're not, please be honest, then let's close this. I gained some learnings in the design.

meerumschlungen:
> @meerumschlungen Are you still actively working on this? @TwiN Is it realistic to achieve a merge here in the near future? I think it would also help with my problem #1238.

Without having tested this, I think that the reimplementation might solve your problem. I could create a unit test for this in this branch.

TwiN:
Not against reviewing this, but I'm currently working on a large change, and I want to merge that first.

dennisoderwald:
@TwiN Thanks so far. Is there an ETA for this large change? Just so @meerumschlungen can plan ahead, it would be helpful for him to know when to schedule it.

joaojacome:
random idea: would it make sense to use yqlib to parse both json and yaml files?

https://pkg.go.dev/github.com/mikefarah/yq/v4@v4.47.2/pkg/yqlib

PR #1056, opened 2025-04-01
fix(client): Use libcap to fix icmp not working in containers as non-root

### LINKED ISSUE #697: ICMP not working on Kubernetes even if sysctl -w net.ipv4.ping_group_range="0 2147483647"

thequailman:
## Summary

Fixes #697

This PR modifies the Gatus binary in the container to have the CAP_NET_RAW capability.  Gatus can now perform ICMP monitoring without running as a privileged or root container.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,11 @@
 # Build the go application into a binary
 FROM golang:alpine AS builder
-RUN apk --update add ca-certificates
+RUN apk --update add ca-certificates libcap-setcap
 WORKDIR /app
 COPY . ./
 RUN go mod tidy
 RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o gatus .
+RUN setcap CAP_NET_RAW+ep gatus
 
 # Run Tests inside docker image if you don't have a configured go environment
 #RUN apk update && apk add --virtual build-dependencies build-base gcc


TwiN (approved):

thequailman:
I think there are three options here: 

- Require CAP_NET_RAW since gatus doesn't work without it (Go ping requires it).  Folks may not be using ping, but the capability doesn't really decrease security that much.
- Have two gatus container images, one with the cap added, one without
- Make the setcap step conditional by adding a shell and cap tooling to the container and have an entrypoint script that conditionally performs setcap.

thequailman:
FWIW, [Docker enables CAP_NET_RAW by default)[https://docs.docker.com/engine/containers/run/#:~:text=The%20following%20table%20lists%20the%20Linux%20capability%20options%20which%20are%20allowed%20by%20default%20and%20can%20be%20dropped.]

[Telegraf had a similar situation and required CAP_NET_RAW for podman folks](https://williamlieurance.com/telegraf-podman-cap-net-raw/)

TwiN:
The chaotic inside of me just makes me feel like having those using podman deal with it, mostly because podman users are a minority.

TwiN:
Based on what I'm seeing on https://github.com/containerd/containerd/issues/10434, it seems like CAP_NET_RAW being enabled by default may be changed in the future.

I wonder if this is why Gatus worked just fine for me on Kubernetes even before this change?

dkowis:
I don't mind that it would be documented for podman users. As a podman user, I'm fine with that :) It seems reasonable to me.

PR #1059, opened 2025-04-02
feat: Add refresh button functionality to Settings component

mouradev1:
## Summary
Added functionality to the refresh button in the `Settings` component. Clicking the button now triggers the `refreshData` method, which emits a `refreshData` event to update the data.

## Checklist
- [X] Tested and validated that the refresh button triggers the `refreshData` method as intended.
- [ ] No updates to the `README.md` were necessary for this change.


TwiN:
No longer necessary as there's a refresh button now

PR #1077, opened 2025-04-15
Update Dependabot configuration to stop automatic PRs

paskal:
This PR updates the Dependabot configuration to:
- Remove GitHub Actions updates
- Set open-pull-requests-limit to 0 for Go modules and NPM
- Remove groups configuration where present

These changes prevent Dependabot from creating automatic PRs for version updates while still allowing it to scan for vulnerabilities and create security updates when needed.


TwiN:
Uh, I'm not sure why you think I don't want automatic dependabot PRs, but I specifically configured dependabot so I would get PRs from it

paskal:
My bad, I was doing changes somewhere else systemically and then by accident grabbed your fork and claude code followed my instructions and sent the PR to the main repo (yours) instead of the fork it was supposed to send the change to.
Hence the second commit: it's unmerged change #442 from the fork.

PR #1092, opened 2025-05-05
Feature/external endpoint data

bcross:
## Summary

External endpoints should be able to report and track health check duration similar to other endpoints.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,13 +319,14 @@ external-endpoints:
 
 To push the status of an external endpoint, the request would have to look like this:
 ```
-POST /api/v1/endpoints/{key}/external?success={success}&error={error}
+POST /api/v1/endpoints/{key}/external?success={success}&error={error}&duration={duration}
 ```
 Where:
 - `{key}` has the pattern `<GROUP_NAME>_<ENDPOINT_NAME>` in which both variables have ` `, `/`, `_`, `,` and `.` replaced by `-`.
   - Using the example configuration above, the key would be `core_ext-ep-test`.
 - `{success}` is a boolean (`true` or `false`) value indicating whether the health check was successful or not.
 - `{error}`: a string describing the reason for a failed health check. If {success} is false, this should contain the error message; if the check is successful, it can be omitted or left empty.
+- `{duration}`: the time that the request took as a duration string. It can be omitted.
 
 You must also pass the token as a `Bearer` token in the `Authorization` header.
 
--- a/api/external_endpoint.go
+++ b/api/external_endpoint.go
@@ -46,6 +46,14 @@ func CreateExternalEndpointResult(cfg *config.Config) fiber.Handler {
 			Success:   c.QueryBool("success"),
 			Errors:    []string{},
 		}
+		if result.Success && c.Query("duration") != "" {
+			parsedDuration, err := time.ParseDuration(c.Query("duration"))
+			if err != nil {
+				logr.Errorf("[api.CreateExternalEndpointResult] Invalid duration from string=%s", c.Query("duration"))
+				return c.Status(400).SendString("invalid duration")
+			}
+			result.Duration = parsedDuration
+		}
 		if !result.Success && c.Query("error") != "" {
 			result.Errors = append(result.Errors, c.Query("error"))
 		}
--- a/api/external_endpoint_test.go
+++ b/api/external_endpoint_test.go
@@ -71,6 +71,12 @@ func TestCreateExternalEndpointResult(t *testing.T) {
 			ExpectedCode:                   400,
 		},
 		{
+			Name:                           "bad-duration-value",
+			Path:                           "/api/v1/endpoints/g_n/external?success=true&duration=invalid",
+			AuthorizationHeaderBearerToken: "Bearer token",
+			ExpectedCode:                   400,
+		},
+		{
 			Name:                           "good-token-success-true",
 			Path:                           "/api/v1/endpoints/g_n/external?success=true",
 			AuthorizationHeaderBearerToken: "Bearer token",
@@ -97,6 +103,12 @@ func TestCreateExternalEndpointResult(t *testing.T) {
 		{
 			Name:                           "good-token-success-false-with-error",
 			Path:                           "/api/v1/endpoints/g_n/external?success=false&error=failed",
+			AuthorizationHeaderBearerToken: "Bearer token",
+			ExpectedCode:                   200,
+		},
+		{
+			Name:                           "good-duration-success-true",
+			Path:                           "/api/v1/endpoints/g_n/external?success=true&duration=10s",
 			AuthorizationHeaderBearerToken: "Bearer token",
 			ExpectedCode:                   200,
 		},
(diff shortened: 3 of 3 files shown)


api/external_endpoint.go
```diff
@@ -46,6 +46,14 @@ func CreateExternalEndpointResult(cfg *config.Config) fiber.Handler {
...
 			Errors:    []string{},
 		}
+		if result.Success && c.Query("duration") != "" {
+			parsedDuration, err := time.ParseDuration(c.Query("duration"))
+			if err != nil {
+				logr.Errorf("[api.CreateExternalEndpointResult] Invalid duration from string=%s", c.Query("duration"))
+				return c.Status(400).SendString("invalid duration")
+			}
+			result.Duration = parsedDuration
+		}
```
TwiN, line 56: Is there a reason why only successful results should parse the duration? I feel like it should always be parsed. If a user doesn't want to take the duration into account on failure, they should handle that on the client side.

TwiN (changes requested):

TwiN:
@bcross Sorry about the delay, just a small comment. Looks good otherwise.

api/external_endpoint.go
bcross: Thank you for the feedback! I went back and forth. Just because I can't think of a situation I would record a failure duration doesn't mean it doesn't exist. I've made that change.

api/external_endpoint.go
```diff
@@ -46,7 +46,7 @@ func CreateExternalEndpointResult(cfg *config.Config) fiber.Handler {
 			Success:   c.QueryBool("success"),
 			Errors:    []string{},
 		}
-		if result.Success && c.Query("duration") != "" {
+		if c.Query("duration") != "" {
```
TwiN, line 49: can you use `len(c.Query("duration")) > 0` instead? Nitpicking, but it's more efficient.

TwiN (changes requested):

api/external_endpoint.go
bcross: Of course! Thanks for the insight!

README.md
```diff
@@ -319,13 +319,14 @@ external-endpoints:
...
 ```
-POST /api/v1/endpoints/{key}/external?success={success}&error={error}
+POST /api/v1/endpoints/{key}/external?success={success}&error={error}&duration={duration}
 ```
 Where:
 - `{key}` has the pattern `<GROUP_NAME>_<ENDPOINT_NAME>` in which both variables have ` `, `/`, `_`, `,` and `.` replaced by `-`.
   - Using the example configuration above, the key would be `core_ext-ep-test`.
 - `{success}` is a boolean (`true` or `false`) value indicating whether the health check was successful or not.
 - `{error}`: a string describing the reason for a failed health check. If {success} is false, this should contain the error message; if the check is successful, it can be omitted or left empty.
+- `{duration}`: the time that the request took as a duration string. It can be omitted.
```
TwiN, line 329:
```suggestion
- `{duration}`: the time that the request took as a duration string (e.g. 10s). It can be omitted. 
```

TwiN (approved):

api/external_endpoint.go
```diff
@@ -46,6 +46,14 @@ func CreateExternalEndpointResult(cfg *config.Config) fiber.Handler {
 			Success:   c.QueryBool("success"),
 			Errors:    []string{},
 		}
+		if len(c.Query("duration")) > 0 {
+			parsedDuration, err := time.ParseDuration(c.Query("duration"))
+			if err != nil {
+				logr.Errorf("[api.CreateExternalEndpointResult] Invalid duration from string=%s", c.Query("duration"))
+				return c.Status(400).SendString("invalid duration")
+			}
```
TwiN, line 54: Can you include the error message (`err.Error()`) in both the log message and the error string returned?

TwiN (changes requested):

api/external_endpoint.go
bcross: I've included the error text in the log and the return string. Let me know if it's not how you were thinking about it.

bcross:
@TwiN I don't want to be a pest. Do you need anything else from me on this PR?

TwiN:
@bcross you are absolutely not being a pest 😆

I have a newborn home, and I've had to prioritize real life for the past couple months, but I'm slowly ramping back up online shenanigans. I'll have a look now :)

api/external_endpoint_test.go
```diff
@@ -138,16 +150,19 @@ func TestCreateExternalEndpointResult(t *testing.T) {
 		if len(endpointStatus.Results[1].Errors) > 0 {
 			t.Errorf("expected second result to have no errors")
 		}
-		if endpointStatus.Results[2].Success {
-			t.Errorf("expected third result to be unsuccessful")
```
TwiN, line 142: I don't think the numbering makes much sense here.  The `[2]` element is still the third, not the sixth

api/external_endpoint_test.go
```diff
@@ -138,16 +150,19 @@ func TestCreateExternalEndpointResult(t *testing.T) {
...
+		if endpointStatus.Results[2].Duration == 0 || endpointStatus.Results[2].Duration.Seconds() != 10 {
+			t.Errorf("expected sixth result to have a duration of 10 seconds")
 		}
 		if endpointStatus.Results[3].Success {
-			t.Errorf("expected fourth result to be unsuccessful")
+			t.Errorf("expected third result to be unsuccessful")
```
TwiN: Same here, this change doesn't make sense.

TwiN (changes requested):

bcross:
> I have a newborn home

I have four kids. You're preaching to the choir.

TwiN:
Four???? Where do you even find the time to make pull requests?! 🤣

api/external_endpoint_test.go
bcross: You're right. I fixed the numbering altogether.

api/external_endpoint_test.go
bcross: You're right. I fixed the numbering altogether.

TwiN (approved):
Looks great!

api/external_endpoint.go
```diff
@@ -46,6 +46,14 @@ func CreateExternalEndpointResult(cfg *config.Config) fiber.Handler {
 			Success:   c.QueryBool("success"),
 			Errors:    []string{},
 		}
+		if len(c.Query("duration")) > 0 {
+			parsedDuration, err := time.ParseDuration(c.Query("duration"))
+			if err != nil {
+				logr.Errorf("[api.CreateExternalEndpointResult] Invalid duration from string=%s with error: %s", c.Query("duration"), err.Error())
+				return c.Status(400).SendString(fmt.Sprintf("invalid duration with error: %s", err.Error()))
```
TwiN:
```suggestion
				return c.Status(400).SendString("invalid duration: " + err.Error())
```

TwiN (approved):

TwiN:

PR #1099, opened 2025-05-12
docs: Update list of characters sanitised in endpoint key

mloskot:
The hash was missing.

This aligns the docs with the implementation

https://github.com/TwiN/gatus/blob/b388cc87aa30fcfb83601a32c40a49286cc7374d/config/endpoint/key.go#L10-L19


TwiN (approved):

TwiN (approved):
Good catch! Thank you

Issue #850, opened 2024-08-29
Add support for sending response body in the alert message


arnasbr:
### Describe the feature request

It would be nice to have a flag that would make Gatus send the response body together with the alert. A bonus would be setting a condition for it. For example, I only want to see the response body if `response.status != 200`.

This issue https://github.com/TwiN/gatus/issues/709 sort of overlaps, but maybe implementing this flag would be quicker / easier than full on templating.

### Why do you personally want this feature to be implemented?

It is very inconvenient seeing a non-200 response code, but not being able to see what error message came with it.

### How long have you been using this project?

Over a month

darkspine:
I would also love to have this feature. i have just created and api that does health check on our systems and return the error message if there is something wrong. made gatus preodicly call this api and send opsgenie alert whenever there is a failure. I would love to for gatus to also send the response of this api so we can know what is the issue directly

imrelaszlo:
Related idea: it would be great if text could be placed in the notification based on a JSON path from the HTTP response body, like in the conditions. Example:

```
    alerts:
      - type: email
        description: "[BODY][0].response_text"
```

PR #1105, opened 2025-05-15
(feat) Add support for sending response body in alert message (with JSONPath support)

imrelaszlo:
This modification allows sending HTTP(S) endpoint response (BODY) in notification messages. To do this, the alert provider's description value must be set to [BODY]. For universal usability, it is also possible to use JSONPath, for this I used this library:

https://github.com/ohler55/ojg

It is also possible to concatenate values returned in an array with an enumeration, for example:

Endpoint response HTTP BODY:

```
[
    {
        "category": "0",
        "device": "5",
        "label": "Field value 1",
        "id": "17589"
    },
    {
        "category": "0",
        "device": "3",
        "label": "Field value 2",
        "id": "17378"
    },
    {
        "category": "0",
        "device": "3",
        "label": "Field value 3",
        "id": "17379"
    }
]
```

Relevant config section:

```
    alerts:
      - type: email
        description: "[BODY][*].label"
```

The description in the alert will be: _Field value 1, Field value 2, Field value 3_

## Summary

#850 - Add support for sending response body in alert message

## Checklist

### DIFF
--- a/alerting/alert/alert.go
+++ b/alerting/alert/alert.go
@@ -142,9 +142,8 @@ func (alert *Alert) GetDescription(body []byte) string {
 			str, err := ExtractStringFromResponseBodyByJsonPath(strings.TrimPrefix(strings.TrimPrefix(*alert.Description, BodyPlaceholder), "."), body)
 			if err != nil {
 				logr.Debug(err.Error())
-			} else {
-				return str
 			}
+			return str
 		}
 	}
 	return *alert.Description


TwiN:
There's a reason why I didn't implement this. We don't persist the body because it's too big.

This PR adds too much confusion to the code (e.g. why does alert.GetDescription() now take in the body as a parameter? The name of the method no longer makes sense, we're not just retrieving the description, we're parsing a variable in it):which is generally why I ask people to create an issue first so we can discuss it.

Once upon a time, the body was persisted, but the issue is that the body is sometimes extremely large, which causes memory issues (and occasionally breaks some alerting providers' max character limits). Sure, truncating it might help, but then it would break json parsing. At the moment, however, the result body is set to nil after the endpoint's health is evaluated, which I believe means your implementation wouldn't work.

Another issue is that you're now introducing a new JSONPath implementation with a different syntax. I'm not against improving the current JSONPath implementation (or even replacing it with an existing library, so long as we can keep it backward compatible), but having two different syntaxes in one application is not something I'm ok with because it confuses users. If you want a new JSONPath implementation, I'd suggest starting with this, and once that's done, we can look into passing values from the body to the description. While I don't think we should persist the entire body, I think we might be able to come up with a way to pass values from conditions that use jsonpath into the description. In essence, we wouldn't save the whole body, just the parts we use.

imrelaszlo:
Hello @TwiN !

Of course I understand and accept the rejection. This would be a very important feature for us. Unfortunately, I am a beginner GO programmer, this is my first attempt in this language. We would pay for the development of the feature. I know that this is not the norm in the open source world, but if there is a way, we would actually have money for it. Let's talk about it, please, if it can be solved.

imrelaszlo:
> @imrelaszlo First off, amazing work for your first interaction with Go! If you hadn't told me this was your first time, I'd have never guessed.

Thank you. :) I have about 20 years of programming experience, but in completely different languages.

> I'd suggest you publish your modified version of Gatus into its own Docker container image for now if this feature is that important to you. While I intend to introduce quite a few changes to Gatus in the coming months, I'm not sure it's possible to persist the body of every request sent without potentially causing performance issues, so it's not something I want to do lightly as it would impact many people.

I know this implementation is not pretty, but it was really necessary. As you suggested, I created my own image, and it works.

> Can you create an issue that outlines what feature(s) you need? Also make sure to link the new issue to this PR, since there is some valuable information here already.

Yes, of course. Here: https://github.com/TwiN/gatus/discussions/1107

PR #1106, opened 2025-05-15
perf(sqlite): Create indices to fix performance issue

g-hodgson-tup:
## Summary
Adds indices to sqlite tables. see https://github.com/TwiN/gatus/discussions/1102

### DIFF
--- a/storage/store/sql/specific_sqlite.go
+++ b/storage/store/sql/specific_sqlite.go
@@ -81,5 +81,27 @@ func (s *Store) createSQLiteSchema() error {
 	`)
 	// Silent table modifications TODO: Remove this in v6.0.0
 	_, _ = s.db.Exec(`ALTER TABLE endpoint_results ADD domain_expiration INTEGER NOT NULL DEFAULT 0`)
+
+	if err != nil {
+		return err
+	}
+	_, err = s.db.Exec(`
+        CREATE INDEX IF NOT EXISTS endpoint_results_endpoint_id_idx ON endpoint_results (endpoint_id);
+	`)
+
+	if err != nil {
+		return err
+	}
+	_, err = s.db.Exec(`
+		CREATE INDEX IF NOT EXISTS endpoint_uptimes_endpoint_id_idx ON endpoint_uptimes (endpoint_id);
+	`)
+
+	if err != nil {
+		return err
+	}
+	_, err = s.db.Exec(`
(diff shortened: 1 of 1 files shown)


storage/store/sql/specific_sqlite.go
```diff
@@ -81,5 +81,27 @@ func (s *Store) createSQLiteSchema() error {
 	`)
 	// Silent table modifications TODO: Remove this in v6.0.0
 	_, _ = s.db.Exec(`ALTER TABLE endpoint_results ADD domain_expiration INTEGER NOT NULL DEFAULT 0`)
+
```
TwiN, line 84: Can you move the code you added above the `// Silent table modification` comment? Maybe add in a comment about how this improves the perf too above the block of code you're adding 👀

TwiN (changes requested):

storage/store/sql/specific_sqlite.go
```diff
@@ -81,5 +81,27 @@ func (s *Store) createSQLiteSchema() error {
...
+		CREATE INDEX IF NOT EXISTS endpoint_uptimes_endpoint_id_idx ON endpoint_uptimes (endpoint_id);
+	`)
+
+	if err != nil {
+		return err
+	}
+	_, err = s.db.Exec(`
+		CREATE INDEX IF NOT EXISTS endpoint_result_conditions_endpoint_result_id_idx ON endpoint_result_conditions (endpoint_result_id);
+	`)
+
```
TwiN, line 105:
```suggestion
	if err != nil {
		return err
	}
	// Create indices for performance reasons
	_, err = s.db.Exec(`
		CREATE INDEX IF NOT EXISTS endpoint_results_endpoint_id_idx ON endpoint_results (endpoint_id);
	`)
	if err != nil {
		return err
	}
	_, err = s.db.Exec(`
		CREATE INDEX IF NOT EXISTS endpoint_uptimes_endpoint_id_idx ON endpoint_uptimes (endpoint_id);
	`)
	if err != nil {
		return err
	}
	_, err = s.db.Exec(`
		CREATE INDEX IF NOT EXISTS endpoint_result_conditions_endpoint_result_id_idx ON endpoint_result_conditions (endpoint_result_id);
	`)
	// Silent table modifications TODO: Remove this in v6.0.0
	_, _ = s.db.Exec(`ALTER TABLE endpoint_results ADD domain_expiration INTEGER NOT NULL DEFAULT 0`)
```

TwiN (approved):

PR #1113, opened 2025-05-20
Allow configuration of config reload duration

g-hodgson-tup:
## Summary

Adds a configuration option to set the duration between checks for config changes, when reloading configuration on the fly.

Example config: `configReloadCheckDuration: 10s`

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,6 +228,7 @@ If you want to test it locally, see [Docker](#docker).
 | `security`                   | [Security configuration](#security).                                                                                                 | `{}`                       |
 | `disable-monitoring-lock`    | Whether to [disable the monitoring lock](#disable-monitoring-lock).                                                                  | `false`                    |
 | `skip-invalid-config-update` | Whether to ignore invalid configuration update. <br />See [Reloading configuration on the fly](#reloading-configuration-on-the-fly). | `false`                    |
+| `configReloadCheckDuration`  | Duration between checking for config changes, e.g. 10s. <br />See [Reloading configuration on the fly](#reloading-configuration-on-the-fly). | `30s`                    |
 | `web`                        | Web configuration.                                                                                                                   | `{}`                       |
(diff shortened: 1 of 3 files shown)


TwiN:
What's the use case for this? I'm a big fan of keeping the configuration as simple as possible, and personally, I find this isn't particularly useful, unless there's a good reason why you need this should be configurable

g-hodgson-tup:
Hi - in some cases we would rather reload the config in gatus faster - rather than waiting 30s for the config polling to occur.

README.md
```diff
@@ -228,6 +228,7 @@ If you want to test it locally, see [Docker](#docker).
 | `security`                   | [Security configuration](#security).                                                                                                 | `{}`                       |
 | `disable-monitoring-lock`    | Whether to [disable the monitoring lock](#disable-monitoring-lock).                                                                  | `false`                    |
 | `skip-invalid-config-update` | Whether to ignore invalid configuration update. <br />See [Reloading configuration on the fly](#reloading-configuration-on-the-fly). | `false`                    |
+| `configReloadCheckDuration`  | Duration between checking for config changes, e.g. 10s. <br />See [Reloading configuration on the fly](#reloading-configuration-on-the-fly). | `30s`                    |
```
TwiN, line 231: The configuration uses kebab-case, not camelCase

config/config.go
```diff
@@ -101,6 +101,9 @@ type Config struct {
 
 	configPath      string    // path to the file or directory from which config was loaded
 	lastFileModTime time.Time // last modification time
+
+	// how often to check for config changes
+	ConfigReloadCheckDuration time.Duration `yaml:"configReloadCheckDuration,omitempty"`
```
TwiN, line 106: `ConfigReloadCheckInterval` would make more sense than *Duration

TwiN (changes requested):

config/config.go
g-hodgson-tup: ah yes, a much better name :)

Issue #1100, opened 2025-05-13
Pipe character in endpoint key breaks detailed chart in Chrome


mloskot:
### Describe the bug

Using `|` character in endpoint name seems to break access to the endpoint details due to malformed URL of API request. Interestingly, it works in Firefox.

### What do you see?

Endpoint responses time chart is missing

### What do you expect to see?

Endpoint responses time chart and other details displayed.

### List the steps that must be taken to reproduce this issue

1. Replace `-` with `|` in this endpoint https://github.com/TwiN/gatus/blob/f3a346d91c151b599db22300665083d9e285e436/config.yaml#L2
2. Launch the UI in Chrome
3. Click on the endpoint to access page with its details

Here is screencast showing the issue in action, first in Firefox (works) then in Chrome (broken) on Ubuntu 25.04:

https://github.com/user-attachments/assets/8af82de8-90f2-4bec-80de-e61649f3e654

### Version

Current master branch

### Additional information

Perhaps `|` should be added to this set:
https://github.com/TwiN/gatus/blob/f3a346d91c151b599db22300665083d9e285e436/config/endpoint/key.go#L10-L17

dpkrane:
All characters except Latin letters in group or endpoint names break the detailed chart display.

![Image](https://github.com/user-attachments/assets/125e066d-42c9-46e2-aad1-0a5d9b2b7231)

mloskot:
@dpkrane 
> All characters except Latin letters (...) break

I found `@`, `:` and `/` are benign:
![Image](https://github.com/user-attachments/assets/e31e744f-cbea-41e9-980f-ea70e5c67487)

dpkrane:
> [@dpkrane](https://github.com/dpkrane)
> 
> > All characters except Latin letters (...) break
> 
> I found `@`, `:` and `/` are benign: ![Image](https://github.com/user-attachments/assets/e31e744f-cbea-41e9-980f-ea70e5c67487)
> 
> ![Image](https://github.com/user-attachments/assets/7373633f-139f-4df9-8dbf-1e7a39c27cf3)

Try cyrillic, please.

mloskot:
@dpkrane 
> Try cyrillic, please.

Yes, I get that may non-English letters may not work as you showed and I'm not arguing with that.
I just reacted to your _"All characters except Latin letters"_ refining the **All** with several exceptions of non-letters that do not break the UI.

dpkrane:
[...]

Issue #1088, opened 2025-05-04
Umlaut in endpoint name breaks detail page


twiesing:
### Describe the bug

If the name of an endpoint contains an umlaut (e.g. “ä”) or "ß", no detail page of the check is generated. The API error is "endpoint not found".

### What do you see?

<img width="1309" alt="Image" src="https://github.com/user-attachments/assets/7362307e-e4a0-414a-b94e-8f54bae978a3" />

TwiN:
Fixed in #1114

PR #1114, opened 2025-05-22
ISSUE #1100 | Fix bug with URL encoding

dpkrane:
The issue occurred because the configuration file stores an endpoint with a Cyrillic name as-is, while the API request sends the key URL-encoded.
The solution was to apply **_url.QueryUnescape_** in the relevant places.

### DIFF
--- a/api/badge.go
+++ b/api/badge.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/url"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


twiesing:
Maybe related https://github.com/TwiN/gatus/issues/1088?

dpkrane:
> Maybe related #1088?

Yes, I think so.
You can check it.
Patched code in my forked repo.
Docker container builded from original Dockerfile.

TwiN (approved):
Makes sense!

Issue #1117, opened 2025-05-25
update certificate expiration check to check full chain


adamaze:
### Describe the feature request

I like the cert expiration check, but it only checks the initial cert. Can it check all the certs in the chain?

### Why do you personally want this feature to be implemented?

We just had an internal issuer cert expire, and this feature would have saved me, as I was already monitoring cert expiration

### How long have you been using this project?

6 months

TwiN:
That's a good point!

volodimyr:
@TwiN Let me know if it looks good. I've tested the changes it works for me.

parkr:
Hey, I have a Tailscale Let's Encrypt cert that expires in 8 days, but Gatus thinks it's >30d expiration. My hunch is that the certificate is a chain and it's just checking the intermediate / root certs because this box uses TrueNAS and maybe they serve the certificates in a non-standard order where the leaf cert is last or something.

PR #1121, opened 2025-05-27
Feature/cert chain

volodimyr:
## Summary
https://github.com/TwiN/gatus/issues/1117
Enhanced certificate monitoring to check the entire certificate chain (leaf, intermediate, and root certificates) instead of just the leaf certificate.
Added new metrics `gatus_results_certificate_chain_expiration_seconds` with subject and issuer labels to track expiration times for each certificate in the chain.
Updated the endpoint evaluation logic in `evaluateSTARTTLS` and `evaluateTLS ` functions to store and expose the full certificate chain information, maintaining backward compatibility with existing [CERTIFICATE_EXPIRATION] placeholder.

## Checklist

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -122,53 +122,67 @@ func CanCreateSCTPConnection(address string, config *Config) bool {
 	}
 }
 
+// CertificateChainInfo contains information about a TLS certificate chain
+type CertificateChainInfo struct {
+	Connected bool
+	Chain     []*x509.Certificate
+	Error     error
+}
+
 // CanPerformStartTLS checks whether a connection can be established to an address using the STARTTLS protocol
-func CanPerformStartTLS(address string, config *Config) (connected bool, certificate *x509.Certificate, err error) {
+func CanPerformStartTLS(address string, config *Config) CertificateChainInfo {
 	hostAndPort := strings.Split(address, ":")
 	if len(hostAndPort) != 2 {
-		return false, nil, errors.New("invalid address for starttls, format must be host:port")
+		return CertificateChainInfo{Error: errors.New("invalid address for starttls, format must be host:port")}
 	}
 	connection, err := net.DialTimeout("tcp", address, config.Timeout)
 	if err != nil {
-		return
+		return CertificateChainInfo{Error: err}
 	}
 	smtpClient, err := smtp.NewClient(connection, hostAndPort[0])
 	if err != nil {
-		return
+		return CertificateChainInfo{Error: err}
 	}
 	err = smtpClient.StartTLS(&tls.Config{
 		InsecureSkipVerify: config.Insecure,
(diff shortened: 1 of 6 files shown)


TwiN (commented):
If the root certificate in the chain expires in, for instance, 5 days, but the child certificate expires in 30 days, but the root certificate authority isn't owned by the user using Gatus to monitor his environment, does it make sense to alert based on the expiration of the root certificate? Maybe they automatically renew a day before it expires.

(This isn't a rhetorical question, feel free to let me know what you think. I'm just trying to figure out if this change in behavior may upset some users)

client/client.go
```diff
@@ -122,53 +122,67 @@ func CanCreateSCTPConnection(address string, config *Config) bool {
...
+	Error     error
+}
+
 // CanPerformStartTLS checks whether a connection can be established to an address using the STARTTLS protocol
-func CanPerformStartTLS(address string, config *Config) (connected bool, certificate *x509.Certificate, err error) {
+func CanPerformStartTLS(address string, config *Config) CertificateChainInfo {
 	hostAndPort := strings.Split(address, ":")
 	if len(hostAndPort) != 2 {
-		return false, nil, errors.New("invalid address for starttls, format must be host:port")
+		return CertificateChainInfo{Error: errors.New("invalid address for starttls, format must be host:port")}
```
TwiN, line 136: I don't know how I feel about not passing the error as a separate return value

TwiN:
In light of that, I think that what I'd want is a root-level configuration such as `disable-full-chain-certificate-expiration-check`.

By default, Gatus should check the full certificate chain, which means this variable would default to `false`, but if a user only wants to check the leaf certificate, then they can set that configuration to `true`.

P.S. If you can think of a better name for the configuration, please feel free to suggest it.

volodimyr:
> In light of that, I think that what I'd want is a root-level configuration such as `disable-full-chain-certificate-expiration-check`.
> 
> By default, Gatus should check the full certificate chain, which means this variable would default to `false`, but if a user only wants to check the leaf certificate, then they can set that configuration to `true`.
> 
> P.S. If you can think of a better name for the configuration, please feel free to suggest it.

Finalised the work and tested it. Looks good.
Hope I didn't mess up any previous logic. Though I did find few mistakes today.

Exagone313:
My two cents about this: would it be possible to keep the current behavior as default and only enable full-chain certificate check on demand?

TwiN:
I've got another idea instead. We could implement a new placeholder, `[CERTIFICATE_CHAIN_EXPIRATION]`.
That would allow people to check for whatever they wish. Open to other names - perhaps `[CERTIFICATE_FULL_CHAIN_EXPIRATION]` would make more sense, though it's quite lengthy

PR #1122, opened 2025-05-27
Fix Gitea issues not being closed

miggland:
## Summary
Gitea issues created by Gatus were not closed when resolved, for two reasons fixed in this PR:
- do not ask for page 100, but pagesize (limit) 100
- URL/edit function must use Index, not ID.

## Checklist

-[x] Tested and/or added tests to validate that the changes work as intended, if applicable.


TwiN (approved):
Good catch!

PR #1124, opened 2025-05-29
feat: support new tls.ca-file field

sdil:
## Summary

We have a use case where we want to monitor self-signed cert served from MTLS endpoint. Currently, Gatus will show error `tls: failed to verify certificate: x509: certificate signed by unknown authority`

With this change, I'm able to specify my own CA and the check is passing.

## Checklist


README.md
```diff
@@ -525,6 +526,7 @@ endpoints:
       tls:
         certificate-file: /path/to/user_cert.pem
         private-key-file: /path/to/user_key.pem
+        ca-file: /path/to/root-ca.pem
```
TwiN: Could you specify in the table that  ca-file is optional for non-self-signed certificate authorities, and a similar comment in the yaml example  next to `ca-file: /path/to/root-ca.pem`?

For verbosity, I want devs to keep their configuration as simple as possible, and by adding these comments, I hope it'll make it more clear that this isn't required unless it's an unrecognized/self-signed CA

TwiN (changes requested):

README.md
sdil: I've updated the readme with the comment next to the yaml example & the table. does this looks good to you?

define42:
:+1:  @TwiN , please accept this PR—this feature is urgently needed.

define42 (approved):

PR #1129, opened 2025-06-03
feat(alerting): Make email alert provider e-mail customizable and templateable

imrelaszlo:
Please see this pull request: https://github.com/TwiN/gatus/pull/1119

In addition to what is described there, it is possible to send HTML (multipart) emails. To do this, an environment variable must be set for the template folder (GATUS_EMAIL_TEMPLATE_DIR), and for each notification you can configure which template file to use. If there is no template, then use the custom texts, if there is none, then it works according to the original operation (everything is backwards compatible). If template found, plain text body is auto generated (HTML tags stripped out).

```
        provider-override:
          text-email-subject-triggered: "Figyelem - I.T. - Cold security service - riasztás!"
          text-email-subject-resolved: "I.T. - Cold security service - a hiba elhárításra került!"
          file-email-body-triggered: "email-template-alert-triggered.html"
          file-email-body-resolved: "email-template-alert-resolved.html"
```

## Checklist

### DIFF
--- a/alerting/provider/email/email.go
+++ b/alerting/provider/email/email.go
@@ -29,16 +29,10 @@ type Config struct {
 	Port     int    `yaml:"port"`
 	To       string `yaml:"to"`
 
-	// Strings used in the email body and subject
-	// These fields are optional and will be replaced with default values if not set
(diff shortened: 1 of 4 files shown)


TwiN (changes requested):
Please include some documentation in the README

PR #1134, opened 2025-06-11
feat: Add body to TCP, UDP, and TLS endpoints and templating

### LINKED ISSUE #659: UDP endpoint always connected
### Describe the bug

I set up a udp endpoint to check ETS2 Dedicated Server. If I stop ETS2 the endpoint still shows the service as connected.

### What do you see?

_No response_

I'd expect it to fail when the service is not running.

1. Start ETS2
2. Add this config to config.yaml

```
  - name: euro truck sim 2
    group: game
    url: "udp://IPADDRESS:27016"
    interval: 60s
    conditions:
      - "[CONNECTED] == true"
    alerts:
      - type: email
        description: "healthcheck failed"
        send-on-resolved: true     
```

3. Monitor it, is up
4. stop ETS2
5. monitor it, it's still up for gatus

### Version

GO

### Additional information

_No response_

jasonshugart:
## Summary

Fixes #659 

## Checklist


README.md
```diff
@@ -2025,7 +2027,10 @@ endpoints:
       - "[CONNECTED] == true"
 ```
 
-Placeholders `[STATUS]` and `[BODY]` as well as the fields `endpoints[].body`, `endpoints[].headers`,
+If `endpoints[].body` is set it is sent and the first 1024 bytes will be in `[BODY]`. You can use Go template 
+syntax. The functions LocalAddr and RandomString with a length can be used.
```
TwiN: Could you explain what the advantage of adding Go template syntax support add here? 
Given how the number of bytes is restricted to 1024, I can't really think of much use for adding Go templating

client/client.go
```diff
@@ -76,24 +78,62 @@ func GetDomainExpiration(hostname string) (domainExpiration time.Duration, err e
...
+			return localAddr.String()
+		},
+		"RandomString": func(n int) string {
+			const availableCharacterBytes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+			b := make([]byte, n)
+			rand.Seed(time.Now().UnixNano())
+			for i := range b {
+				b[i] = availableCharacterBytes[rand.Intn(len(availableCharacterBytes))]
+			}
+			return string(b)
```
TwiN: I see, this is why you want to use Go templating. 

Honestly, I'm a little perplexed, because we're using placeholders (e.g. `[ENDPOINT_NAME]` and functions (e.g. `len([BODY])`) everywhere else, so it'd be weird if we only used Go templating here.

How would you feel about dropping Go templating and instead adding a function for `RandomString` (e.g. `rand(6)` where `6` is the number of character) and a placeholder for `[LOCAL_ADDRESS]`? fwiw, I don't mind if you add support for those two only for TCP/UDP/TLS endpoints and ignore the rest - though it'd be nice if it was added everywhere - just for the sake of consistency.

README.md
jasonshugart: The response is restricted to 1024 bytes, not the body that is sent. I need it because I am sending a SIP OPTIONS message which requires the local IP address and several random strings. It seems useful for other protocols, but I can certainly remove that part of the code if it would make more sense.

client/client.go
jasonshugart: I can look at changing the templating. Using the Go standard stuff was just easiest. And yes, its seems best to have it everywhere. Let me work on those changes.

jasonshugart: I updated the templating to be more consistent. I applied it to all endpoints that use Body. I used [RANDOM_STRING_6] because the functions (rand(6)) are only in conditions. This seemed easier to understand.

config/endpoint/endpoint.go
```diff
@@ -326,6 +329,26 @@ func (e *Endpoint) EvaluateHealth() *Result {
 	return result
 }
 
+func (e *Endpoint) getModifiedBody() string {
```
TwiN: Can you rename this to getParsedBody()?

Sorry about the delay btw

client/client.go
```diff
@@ -166,9 +182,26 @@ func CanPerformTLS(address string, config *Config) (connected bool, certificate
 	// Reference: https://pkg.go.dev/crypto/tls#PeerCertificates
 	if len(verifiedChains) == 0 || len(verifiedChains[0]) == 0 {
 		peerCertificates := connection.ConnectionState().PeerCertificates
-		return true, peerCertificates[0], nil
```
TwiN, line 169: You're not setting connected to true here

client/client.go
```diff
@@ -166,9 +182,26 @@ func CanPerformTLS(address string, config *Config) (connected bool, certificate
...
+		}
+		buf := make([]byte, MaximumMessageSize)
+		var n int
+		n, err = connection.Read(buf)
+		if err != nil {
+			return
+		}
+		response = buf[:n]
 	}
-	return true, verifiedChains[0][0], nil
```
TwiN, line 171: Same here.

TwiN (changes requested):

client/client.go
```diff
@@ -76,24 +76,37 @@ func GetDomainExpiration(hostname string) (domainExpiration time.Duration, err e
...
+func GetStringReplacement(item string, localAddr net.Addr) string {
+	item = strings.ReplaceAll(item, "[LOCAL_ADDRESS]", localAddr.String())
+	return item
 }
 
-// CanCreateUDPConnection checks whether a connection can be established with a UDP endpoint
-func CanCreateUDPConnection(address string, config *Config) bool {
-	conn, err := net.DialTimeout("udp", address, config.Timeout)
+// CanCreateNetConnection checks whether a connection can be established with a TCP or UDP endpoint
+func CanCreateNetConnection(netType string, address string, body string, config *Config) (bool, []byte) {
```
TwiN: Could you rename `CanCreateNetConnection` to `CanCreateNetworkConnection`?

client/client.go
```diff
@@ -76,24 +76,37 @@ func GetDomainExpiration(hostname string) (domainExpiration time.Duration, err e
...
-	conn, err := net.DialTimeout("tcp", address, config.Timeout)
-	if err != nil {
-		return false
-	}
-	_ = conn.Close()
-	return true
+// GetStringReplacement returns a string with the local address replaced
+func GetStringReplacement(item string, localAddr net.Addr) string {
+	item = strings.ReplaceAll(item, "[LOCAL_ADDRESS]", localAddr.String())
+	return item
```
TwiN: Can you rename the function to `parseLocalAddressPlaceholder`?

client/client.go
```diff
@@ -76,24 +76,37 @@ func GetDomainExpiration(hostname string) (domainExpiration time.Duration, err e
...
-// CanCreateTCPConnection checks whether a connection can be established with a TCP endpoint
-func CanCreateTCPConnection(address string, config *Config) bool {
-	conn, err := net.DialTimeout("tcp", address, config.Timeout)
-	if err != nil {
-		return false
-	}
-	_ = conn.Close()
-	return true
+// GetStringReplacement returns a string with the local address replaced
+func GetStringReplacement(item string, localAddr net.Addr) string {
```
```suggestion
func parseLocalAddressPlaceholder(item string, localAddr net.Addr) string {
```

client/client.go
```diff
@@ -76,24 +76,37 @@ func GetDomainExpiration(hostname string) (domainExpiration time.Duration, err e
...
+		MaximumMessageSize = 1024 // in bytes
+	)
+	connection, err := net.DialTimeout(netType, address, config.Timeout)
 	if err != nil {
-		return false
+		return false, nil
+	}
+	defer connection.Close()
+	if body != "" {
+		body = GetStringReplacement(body, connection.LocalAddr())
```
```suggestion
		body = parseLocalAddressPlaceholder(body, connection.LocalAddr())
```

client/client.go
```diff
@@ -76,24 +76,37 @@ func GetDomainExpiration(hostname string) (domainExpiration time.Duration, err e
...
 
-// CanCreateTCPConnection checks whether a connection can be established with a TCP endpoint
-func CanCreateTCPConnection(address string, config *Config) bool {
-	conn, err := net.DialTimeout("tcp", address, config.Timeout)
-	if err != nil {
-		return false
-	}
-	_ = conn.Close()
-	return true
+// GetStringReplacement returns a string with the local address replaced
```
```suggestion
// parseLocalAddressPlaceholder returns a string with the local address replaced
```

client/client.go
```diff
@@ -166,9 +182,27 @@ func CanPerformTLS(address string, config *Config) (connected bool, certificate
...
 	if len(verifiedChains) == 0 || len(verifiedChains[0]) == 0 {
 		peerCertificates := connection.ConnectionState().PeerCertificates
-		return true, peerCertificates[0], nil
+		certificate = peerCertificates[0]
+	} else {
+		certificate = verifiedChains[0][0]
+	}
+	connected = true
+	if body != "" {
+		body = GetStringReplacement(body, connection.LocalAddr())
```
```suggestion
		body = parseLocalAddressPlaceholder(body, connection.LocalAddr())
```

client/client.go
```diff
@@ -234,6 +268,7 @@ func ExecuteSSHCommand(sshClient *ssh.Client, body string, config *Config) (bool
 	}
 	defer sshClient.Close()
 	var b Body
+	body = GetStringReplacement(body, sshClient.Conn.LocalAddr())
```
```suggestion
	body = parseLocalAddressPlaceholder(body, sshClient.Conn.LocalAddr())
```

client/client.go
```diff
@@ -304,6 +339,7 @@ func QueryWebSocket(address, body string, config *Config) (bool, []byte, error)
 		return false, nil, fmt.Errorf("error dialing websocket: %w", err)
 	}
 	defer ws.Close()
+	body = GetStringReplacement(body, ws.LocalAddr())
```
```suggestion
	body = parseLocalAddressPlaceholder(body, ws.LocalAddr())
```

client/client.go
```suggestion
func CanCreateNetworkConnection(netType string, address string, body string, config *Config) (bool, []byte) {
```

client/client_test.go
```diff
@@ -235,11 +235,13 @@ func TestCanPerformTLS(t *testing.T) {
 	}
 }
 
-func TestCanCreateTCPConnection(t *testing.T) {
-	if CanCreateTCPConnection("127.0.0.1", &Config{Timeout: 5 * time.Second}) {
+func TestCanCreateConnection(t *testing.T) {
+	connected, _ := CanCreateNetConnection("tcp", "127.0.0.1", "", &Config{Timeout: 5 * time.Second})
```
TwiN:
```suggestion
	connected, _ := CanCreateNetworkConnection("tcp", "127.0.0.1", "", &Config{Timeout: 5 * time.Second})
```

client/client_test.go
```diff
@@ -235,11 +235,13 @@ func TestCanPerformTLS(t *testing.T) {
...
+	if connected {
 		t.Error("should've failed, because there's no port in the address")
 	}
-	if !CanCreateTCPConnection("1.1.1.1:53", &Config{Timeout: 5 * time.Second}) {
+	connected, _ = CanCreateNetConnection("tcp", "1.1.1.1:53", "", &Config{Timeout: 5 * time.Second})
```
TwiN:
```suggestion
	connected, _ = CanCreateNetworkConnection("tcp", "1.1.1.1:53", "", &Config{Timeout: 5 * time.Second})
```

config/connectivity/connectivity.go
```diff
@@ -42,7 +42,8 @@ type Checker struct {
 }
 
 func (c *Checker) Check() bool {
-	return client.CanCreateTCPConnection(c.Target, &client.Config{Timeout: 5 * time.Second})
+	connected, _ := client.CanCreateNetConnection("tcp", c.Target, "", &client.Config{Timeout: 5 * time.Second})
```
TwiN:
```suggestion
	connected, _ := client.CanCreateNetworkConnection("tcp", c.Target, "", &client.Config{Timeout: 5 * time.Second})
```

config/endpoint/endpoint.go
```diff
@@ -365,18 +388,18 @@ func (e *Endpoint) call(result *Result) {
 		result.Duration = time.Since(startTime)
 		result.CertificateExpiration = time.Until(certificate.NotAfter)
 	} else if endpointType == TypeTCP {
-		result.Connected = client.CanCreateTCPConnection(strings.TrimPrefix(e.URL, "tcp://"), e.ClientConfig)
+		result.Connected, result.Body = client.CanCreateNetConnection("tcp", strings.TrimPrefix(e.URL, "tcp://"), e.getParsedBody(), e.ClientConfig)
```
TwiN:
```suggestion
		result.Connected, result.Body = client.CanCreateNetworkConnection("tcp", strings.TrimPrefix(e.URL, "tcp://"), e.getParsedBody(), e.ClientConfig)
```

config/endpoint/endpoint.go
```diff
@@ -365,18 +388,18 @@ func (e *Endpoint) call(result *Result) {
...
 		result.Duration = time.Since(startTime)
 	} else if endpointType == TypeUDP {
-		result.Connected = client.CanCreateUDPConnection(strings.TrimPrefix(e.URL, "udp://"), e.ClientConfig)
+		result.Connected, result.Body = client.CanCreateNetConnection("udp", strings.TrimPrefix(e.URL, "udp://"), e.getParsedBody(), e.ClientConfig)
```
TwiN:
```suggestion
		result.Connected, result.Body = client.CanCreateNetworkConnection("udp", strings.TrimPrefix(e.URL, "udp://"), e.getParsedBody(), e.ClientConfig)
```

TwiN (approved):

client/client.go
```diff
@@ -76,24 +76,37 @@ func GetDomainExpiration(hostname string) (domainExpiration time.Duration, err e
...
+// parseLocalAddressPlaceholder returns a string with the local address replaced
+func parseLocalAddressPlaceholder(item string, localAddr net.Addr) string {
+	item = strings.ReplaceAll(item, "[LOCAL_ADDRESS]", localAddr.String())
+	return item
 }
 
-// CanCreateUDPConnection checks whether a connection can be established with a UDP endpoint
-func CanCreateUDPConnection(address string, config *Config) bool {
-	conn, err := net.DialTimeout("udp", address, config.Timeout)
+// CanCreateNetConnection checks whether a connection can be established with a TCP or UDP endpoint
```
```suggestion
// CanCreateNetworkConnection checks whether a connection can be established with a TCP or UDP endpoint
```

TwiN (approved):

TwiN:
@jasonshugart Thank you for your contribution!

PR #1157, opened 2025-07-12
Adding best practice pagination, filtering for success and errors and…

stefanhall2704:
… search bar

## Summary
This PR introduces server-side pagination and filtering to the endpoint statuses view. Previously, all checks were loaded at once, which caused long load times for thousands of endpoints. With these changes, only a single page of results is fetched from the backend at a time, improving performance and responsiveness. The UI now supports paging, searching, and filtering without loading the entire dataset upfront.

## Checklist


TwiN:
Excellent work! I'll have a closer look at it when I have some free time, but at a glance, I'm very thankful you went through the work of implementing such a large change on your own!

stefanhall2704:
Of course, I need it for work is all. And it is on a different endpoint, which you can see in Home.vue. And I only tested in memory, no testing on SQLite or Postgres. But I don't think it'd be a problem.
/endpoints-list.
I don't want it to break Home. So for now I think it may be good there until further testing is done. Tests pass, had to alter tests a bit.

Also generated a config file to test large quantity of endpoints for pagination testing too.

TwiN:
New UI thread: https://github.com/TwiN/gatus/discussions/1195

Those changes will bring filtering/sorting, but only on the frontend. I presume that once the UI changes are merged, you'll have a lot of conflicts, unfortunately 😬

stefanhall2704:
So will all these changes I made basically be covered on the new UI?? If so, I can close this PR

TwiN:
@stefanhall2704 Only the frontend portion. In other words, we'd still retrieve all statuses from the backend and filter through those instead of just the ones we need, so the new UI won't solve your issue wrt large number of endpoints slowing things down when they're fetched.

stefanhall2704:
It's not just front end pagination. We are also fetching via api to paginate by page. Page = 10-20 or something similar depending on the page. Doing it through querying. I've done over 1k endpoints, as shown in the test config file. And I had a way better response time rather than fetching all endpoints at once.

TwiN:
Just to clarify, I'm aware that the new UI won't fix all the change your PR suggests. That's why I didn't close the PR. I'm just giving you a heads up since the UI changes will cause big conflicts with your PR once they're merged

stefanhall2704:
Okay, sounds good. Will keep this on hold until that merges and the new UI is up. Then after that, If some of this code will help us out and solve issues, I'll reference this PR with anything I can help with, and link the two and close this.
Thanks

TwiN:
New UI has been merged - I'm not sure if this is relevant anymore, but have a look and let me know when you have time

TwiN:
Closing this as I'm not sure this is still relevant.

PR #1180, opened 2025-08-02
feat(websocket): add support for custom headers in WS check

olexandrd:
## Summary
Add support for custom HTTP headers (including `User-Agent`) to WebSocket (`ws://` and `wss://`) checks. Headers defined in the endpoint config are now injected into the WebSocket handshake, just like in HTTP checks.

### Motivation
Out of the box, WebSocket checks did not include any HTTP headers in their upgrade handshake. This caused real-world issues such as requests being blocked by WAFs with rules like `NoUserAgent`, making monitoring unreliable for protected endpoints.
Additionally, allowing user-defined headers (e.g., `Authorization`, custom tracing headers, etc.) significantly expands observability and flexibility for WebSocket-based services by enabling authentication, routing, or identification during the handshake.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -2183,7 +2183,7 @@ This works for SCTP based application.
 
 
 ### Monitoring a WebSocket endpoint
-By prefixing `endpoints[].url` with `ws://` or `wss://`, you can monitor WebSocket endpoints at a very basic level:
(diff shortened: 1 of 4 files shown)


TwiN:

Issue #1201, opened 2025-08-14
Add an [EXTERNAL_IP] placeholder for DNS records against dynamic-DNS


gabay:
### Describe the feature request

I think that validating dynamic-DNS records correctness could be useful.
For example "[BODY] == [EXTERNAL_IP]" on a DNS endpoint

Technically - the EXTERNAL_IP should be cached since it rarely changes, and I guess it should also be saved in storage.

I also created a proof-of-concept for that here - https://github.com/gabay/gatus/commit/f17aa50e214d9b34c18c8a3de1b31de4a985794e

I'll gladly open a PR for it if it's useful to more than just me :)

Issue #270, opened 2022-03-24
use response body from another endpoint


emrahcetiner:
### Describe the feature request

I read all docs completely but I couldn't see any information about this. Is there any way for this? I want to add a sesion-token to endpoints.header from another request's response.

emrahcetiner:
It may be possible like this:

```yaml
endpoints:
  - name: website                 
    url: "https://twin.sh/health"
    interval: 5m                 
    conditions:
      - "[STATUS] == 200"        
      - "[BODY].status == UP"     
      - "[RESPONSE_TIME] < 300"   
    env_export:                                  # <--
      - WEBSITE_TOKEN: "[BODY].data.token"       # <--
  - name: example
    url: "https://example.org/"
    headers:
      token: "[${WEBSITE_TOKEN}]"                # <--
    interval: 60s
    conditions:
      - "[STATUS] == 200"
```

emrahcetiner:
@TwiN what do you think about this feature and solution?
Should extracted variables be stored in storage or not?
I want to contribute, but first, is the design right for the project?

TwiN:
I tried looking for the relevant issue to no avail, but in the past, I had started considering the implementation of "suites".

Basically, a suite would contain multiple endpoints that would be run in order. If one of them fails, then the suite would be considered as failed. 

To start, the configuration could look something like this:
```yaml
suites:
  - name: "suite-name"
    group: "group-name"
    endpoints:
      - ...
```

The suite would have a shared context (basically a map) that each endpoint would be able to write to/read from.

It would allow for much more advanced tests, the only problem is that it's quite a large feature and I haven't had the time to spare for a task of this size. That being said, you're free to give it a shot! If you have any questions, don't hesitate to ask :)

TwiN:
Closing in favor of #1230

PR #1225, opened 2025-08-25
feat(endpoint): add "[EXTERNAL_IP]" placeholder.

gabay:
The external IP is fetched on demand from "ifconfig.me" and cached for an hour.

## Summary

#1201
I opened this issue, and would like to have this in a non-forked version.

A user can use [EXTERNAL_IP] as follows:

1. check that a global DNS query points to itself: `[BODY] == [EXTERNAL_IP]`
2. check that its external IP remains constant: `[EXTERNAL_IP] == 1.2.3.4`

## Checklist

### DIFF
--- a/config/endpoint/condition.go
+++ b/config/endpoint/condition.go
@@ -28,6 +28,9 @@ const (
 	// Values that could replace the placeholder: NOERROR, FORMERR, SERVFAIL, NXDOMAIN, NOTIMP, REFUSED
 	DNSRCodePlaceholder = "[DNS_RCODE]"
 
+	// ExternalIPPlaceHolder is a placeholder for dynamically-fetched external IP.
+	ExternalIPPlaceholder = "[EXTERNAL_IP]"
+
 	// ResponseTimePlaceholder is a placeholder for the request response time, in milliseconds.
 	//
 	// Values that could replace the placeholder: 1, 500, 1000, ...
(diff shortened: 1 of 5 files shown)


TwiN:
Not sure how I feel about this. If a user needs to check what IP is assigned to an endpoint, I'd rather they make a DNS call.

Not just that, but I've never heard of that domain before, and I feel it's not appropriate for me to use their API for an open source project of this size without their approval. This is also tricky, because I'd be essentially exposing information to a third party, which would certainly be a problem for some Gatus users (even though they wouldn't so long as they don't use the `[EXTERNAL_IP]` placeholder, not all employees in a company would know about that) and if outbound traffic is detected coming from Gatus to an unknown host, this may become a security concern and reflect negatively on Gatus.

If we decided to implement this, I'd rather use a reputable API (with permission), but again, I'm not sure this is really necessary for most scenarios except for the one you included in the issue (you want to check if your machine's external ip matches the DNS record).

Ultimately, you could just create an endpoint that makes a call to an API that lets you resolve a host's external IP.

gabay:
Fair enough, thanks for your reply

Changing the endpoint would be easy, but if i'm the only consumer of this then it doesn't make sense going forward with it.

Do you want to close this PR and keep the issue to see how many people want it maybe? or just scrap that for now?
I'm fine either way - i'm really enjoying gatus so thanks :D

TwiN:
@gabay I really appreciate your understanding.

If it can be of any consolation, I am considering implementing what I mentioned in this comment: https://github.com/TwiN/gatus/issues/270#issuecomment-1086973356

_maybe_.

It would allow you to define a "suite" of endpoints, the first one of which could be calling ifconfig.me to retrieve the IP, and the second one could be making a DNS call, and then you'd be able to compare the result of endpoint 1 with the result of endpoint 2, which would essentially allow you to do what you're trying to do here.

No promises on timeline, but it's on my radar.

PR #1236, opened 2025-08-31
feat(results): allow for data points in checks to be "clicked"

perfectra1n:
## Summary

Make the data points clicky, so they stay open and you can select the text!

chrome_ky38YqspZt

## Checklist

- [ x ] Tested and/or added tests to validate that the changes work as intended, if applicable.

### DIFF
--- a/web/app/src/App.vue
+++ b/web/app/src/App.vue
@@ -148,7 +148,7 @@
     </div>
 
     <!-- Tooltip -->
-    <Tooltip :result="tooltip.result" :event="tooltip.event" />
+    <Tooltip :result="tooltip.result" :event="tooltip.event" :isPersistent="tooltipIsPersistent" />
   </div>
 </template>
 
@@ -173,6 +173,7 @@ const announcements = ref([])
 const tooltip = ref({})
 const mobileMenuOpen = ref(false)
 const isOidcLoading = ref(false)
+const tooltipIsPersistent = ref(false)
 let configInterval = null
 
 // Computed properties
@@ -209,8 +210,39 @@ const fetchConfig = async () => {
   }
 }
 
-const showTooltip = (result, event) => {
-  tooltip.value = { result, event }
+const showTooltip = (result, event, action = 'hover') => {
+  if (action === 'click') {
+    if (!result) {
+      // Deselecting
+      tooltip.value = {}
+      tooltipIsPersistent.value = false
+    } else {
+      // Selecting new data point
+      tooltip.value = { result, event }
+      tooltipIsPersistent.value = true
+    }
+  } else if (action === 'hover') {
+    // Only update tooltip on hover if not in persistent mode
+    if (!tooltipIsPersistent.value) {
+      tooltip.value = { result, event }
+    }
+  }
+}
+
+const handleDocumentClick = (event) => {
+  // Close persistent tooltip when clicking outside
+  if (tooltipIsPersistent.value) {
+    const tooltipEl = document.getElementById('tooltip')
+    // Check if click is on a data point bar or inside tooltip
+    const clickedDataPoint = event.target.closest('.flex-1.h-6, .flex-1.h-8')
+    
+    if (tooltipEl && !tooltipEl.contains(event.target) && !clickedDataPoint) {
+      tooltip.value = {}
+      tooltipIsPersistent.value = false
+      // Emit event to clear selections in child components
+      window.dispatchEvent(new CustomEvent('clear-data-point-selection'))
+    }
+  }
 }
 
 // Fetch config on mount and set up interval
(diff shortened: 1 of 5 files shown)


perfectra1n:
I guess just let me know if you're interested in this @TwiN and I can resolve the merge conflicts.

TwiN:
Gatus' old, old UI used to have this a long time ago, but it caused issues when you clicked on one, then navigated to another page without closing it first. Can you check that this isn't an issue?

Otherwise, sounds like a good idea.

aaldebs99:
> Gatus' old, old UI used to have this a long time ago, but it caused issues when you clicked on one, then navigated to another page without closing it first. Can you check that this isn't an issue?
> 
> Otherwise, sounds like a good idea.

I tested this, and it doesn't seem to cause issues when switching between pages. Even when jumping and clicking all over the place 😂

perfectra1n:
Fixed :)

chrome_18mEnWCLPN

TwiN:
@perfectra1n Could you fix the merge conflicts when you have a moment?

perfectra1n:
Should be good to go now @TwiN :)

web/app/src/App.vue
```diff
@@ -209,15 +210,48 @@ const fetchConfig = async () => {
...
+    if (!tooltipIsPersistent.value) {
+      tooltip.value = { result, event }
+    }
+  }
+}
+
+const handleDocumentClick = (event) => {
+  // Close persistent tooltip when clicking outside
+  if (tooltipIsPersistent.value) {
+    const tooltipEl = document.getElementById('tooltip')
```
TwiN: Can you rename the variable to tooltipElement?

Nitpicking, I know, but I prefer avoiding shortened words when possible

web/app/src/components/Tooltip.vue
```diff
@@ -1,12 +1,12 @@
 <template>
-  <div 
-    id="tooltip" 
-    ref="tooltip" 
+  <div
+    id="tooltip"
+    ref="tooltip"
     :class="[
-      'fixed z-50 px-3 py-2 text-sm rounded-md shadow-lg border transition-all duration-200',
+      'absolute z-50 px-3 py-2 text-sm rounded-md shadow-lg border transition-all duration-200',
```
TwiN, line 6: Does this work if you scroll down? (e.g. if you have enough groups/endpoints where you can scroll down, when you scroll up/down, is the tooltip properly positioned?)

TwiN (changes requested):

web/app/src/components/Tooltip.vue
perfectra1n: Sorry I'm not sure if I'm understanding - were you hoping to see/validate functionality separate from the gif shared [here](https://github.com/TwiN/gatus/pull/1236#issuecomment-3412770929)?

perfectra1n: Is this what you were looking to see? 👀

chrome_NwoPwzYC0e

web/app/src/App.vue
perfectra1n: Done :)

web/app/src/components/Tooltip.vue
TwiN: Yes, perfect!

TwiN:
@perfectra1n Quick question for you, if the user doesn't click a status (just hovers it), does it go away if cursor hovers elsewhere?

In other words, is the tooltip persistent only when the user clicks a status? I'll test it on my end when I have a minute, but just want to confirm.

perfectra1n:
That is correct:
chrome_x8bYMaJ98c

TwiN:
@perfectra1n Thank you for your contribution!

TwiN:
@perfectra1n If you have some time, when you click multiple results from different EndpointCard, it looks like it leaves the result color from the other EndpointCards as if they're still selected: 


Not sure if you have time to look into this or not - I'll look when I have some time if you don't

TwiN:
@perfectra1n Nevermind, I fixed it in [91931e4](https://github.com/TwiN/gatus/commit/91931e48b46cddffeeeef934118ece177409814b)

perfectra1n:
Ahhhh interesting, good catch. Thanks for doing that!

amai2012:
@TwiN  I think this has some problem still:
In case the tooltip window doesn't fit on the screen (too long w/o scrolling) and I haven't clicked yet it is flickering and barely readable...

I couldn't create a screenshot because then it was rendered fine...

TwiN:
@amai2012 any chance you could record it instead of trying to screenshot it? I tried replicating it on my end, but wasn't able to.

@perfectra1n I believe this may have to do with the change you made to the position algorithm, either that or the fact that we now use absolute instead of fixed for the tooltip container.

perfectra1n:
Hmm, I can't get it to reproduce either. I've tried in viewports that are:
- Very narrow
- Very short
- Resizing from very wide to very skinny
- Resizing from very skinny to very wide
- Resizing from very short to very tall
- Resizing from very tall to very short

and I can't get it to reproduce. I'm wondering if this has to do with some form of caching?

PR #1261, opened 2025-09-15
fix(client): add forward ip support in PTR query-name

denis-ev:
## Summary

Added forward IP support to PTR record, with backwards compatibility
if no in-addr.arpa. is found, the ip will be treated as plain ip and the suffix is added
reduces the risk of typos when adding IP and makes it a lot more user friendly 

## Checklist

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -416,7 +416,6 @@ func QueryDNS(queryType, queryName, url string) (connected bool, dnsRcode string
 	if queryTypeAsUint16 == dns.TypePTR &&
 		!strings.HasSuffix(queryName, ".in-addr.arpa.") &&
 		!strings.HasSuffix(queryName, ".ip6.arpa.") {
-
 		if rev, convErr := reverseNameForIP(queryName); convErr == nil {
 			queryName = rev
 		} else {


TwiN (changes requested):
Looks like you replaced all alignment from tabs to spaces. Can you revert that change?

denis-ev:
> Looks like you replaced all alignment from tabs to spaces. Can you revert that change?

Apologies, converted them back to tabs.

client/client.go
```diff
@@ -410,6 +411,19 @@ func QueryDNS(queryType, queryName, url string) (connected bool, dnsRcode string
 		url = fmt.Sprintf("%s:%d", url, dnsPort)
 	}
 	queryTypeAsUint16 := dns.StringToType[queryType]
+
+	// 🔑 Special handling: if this is a PTR query and queryName looks like a plain IP,
+	// convert it to the proper reverse lookup domain automatically.
+	if queryTypeAsUint16 == dns.TypePTR &&
+		!strings.HasSuffix(queryName, ".in-addr.arpa.") &&
+		!strings.HasSuffix(queryName, ".ip6.arpa.") {
+
```
```suggestion
```

client/client.go
```diff
@@ -410,6 +411,19 @@ func QueryDNS(queryType, queryName, url string) (connected bool, dnsRcode string
 		url = fmt.Sprintf("%s:%d", url, dnsPort)
 	}
 	queryTypeAsUint16 := dns.StringToType[queryType]
+
+	// 🔑 Special handling: if this is a PTR query and queryName looks like a plain IP,
```
```suggestion
	// Special handling: if this is a PTR query and queryName looks like a plain IP,
```

denis-ev:
Thanks, yeah had to look at the tests to figure out how to configure PTR and thought this would make the most sense.
Kept the old way in place so it's not a breaking change.

PR #1286, opened 2025-09-23
feat(client): Support body placeholder for SSH endpoints 

### LINKED ISSUE #1285: Add a placeholder such as "BODY" or "RESPONSE_TEXT" to the SSH type endpoint to get the output of the command

Get the output text of the command (plain text or JSON) for SSH type endpoints.

I write a bash command to monitor memory or disk usage and output it in the form of plain text or JSON. If Gatus can obtain the output of the command, it will become a lightweight server performance monitoring and alarm system.

3 months ago

### Additional information

@TwiN Do you support this feature? If supported, I will work on preparing a PR!

yansh97:
## Summary

Fixes #1285 
- Add BODY placeholder support for SSH endpoints to capture stdout output from remote commands
- Enable use of existing Gatus condition features (JSONPath, pattern matching, functions) with SSH command outputs
- Maintain backward compatibility with existing SSH endpoint configurations

## Checklist


README.md
```diff
@@ -2951,6 +2951,57 @@ endpoints:
       - "[STATUS] == 0"
 ```
 
+You can also use the `[BODY]` placeholder to validate the command output:
```
TwiN: Please remove the two configuration examples you added, it's a bit excessive

yansh97: Two examples have been removed

yansh97:
@TwiN I've removed two configuration examples. Is there anything else that needs to be modified?

yansh97:
Are there any plans to merge?

yansh97:
If you don't have plans to merge this PR, please let me know. I'll use other monitoring tools instead of waiting for you to allow the merge PR. Thank you. @TwiN

TwiN:
I'll merge it in a couple minutes.

TwiN:
@yansh97 The tests are failing.

yansh97:
> @yansh97 The tests are failing.

This is a bit strange. I can pass all the tests locally. I see that the previous ssh test just used the uptime command. Is it possible that CI only provides output for "uptime" and not support for "echo"? @TwiN

TwiN:
@yansh97 I'd be surprised if that was the case, but you could try something other than `echo` just to test. Perhaps `ls`, `cat` or `ping`? 

Might be interesting to try and output any errors

yansh97:
@TwiN I think the previous SSH test was invalid. I found that the "scenario" user is not added to the Github Action test in the configuration. So the previous test case returned the correct status even when the user was not present, causing the test to succeed unexpectedly.

But my newly added test relies on the return of the command result, which causes the test to fail. I tested locally with a real SSH user and got the correct results.

I'm not familiar with Github Action test configuration and I'm not the maintainer for making relevant changes. Please let the maintainers to check the configuration.

TwiN:
I'm not sure what you mean by the scenario user.

According to the test, right now, there's no tests actually connecting to a SSH server. Excluding your tests, there are two tests, one with a condition that says the exit code is 0, which is successful because the connection fails and since integers default to 0, the exit code defaults to 0, which means the condition passes, and another one that says the exit code is 1, which fails because the connection fails which means the exit code again defaults to 0.

Spelling it out like this, I do realize the tests should probably be modified to use a condition with [CONNECTED] instead of [STATUS] (status = exit code) to make this clearer, but ultimately, there's no real SSH server running, so to actually test what - you're trying to test, you'd have to start a SSH server locally prior to running the test.

yansh97:
I don't know how to start an SSH server during CI testing, so I can't modify the test configuration to get the new SSH test to pass. But if you agree, I can remove the newly added test, since my local test has passed.

yansh97:
> I don't know how to start an SSH server during CI testing, so I can't modify the test configuration to get the new SSH test to pass. But if you agree, I can remove the newly added test, since my local test has passed.

I know this is more dangerous and can easily break related features in the future. What do you think?

TwiN:
Go ahead and remove the tests for now. Worst case scenario, the new feature doesn't work, which is fine because it's a new feature. My main concern is that nothing people are currently relying on breaks, and given the size of the feature, it should be fine to take the risk.

yansh97:
I've removed the relevant test cases. @TwiN

client/client.go
```diff
@@ -301,34 +302,43 @@ func CheckSSHBanner(address string, cfg *Config) (bool, int, error) {
...
 	sess, err := sshClient.NewSession()
 	if err != nil {
-		return false, 0, err
+		return false, 0, nil, err
 	}
+
+	// Capture stdout
+	var stdout bytes.Buffer
+	sess.Stdout = &stdout
+
```
```suggestion
	// Capture stdout
	var stdout bytes.Buffer
	sess.Stdout = &stdout
```

client/client.go
```diff
@@ -301,34 +302,43 @@ func CheckSSHBanner(address string, cfg *Config) (bool, int, error) {
...
 	if err != nil {
-		return false, 0, err
+		return false, 0, nil, err
 	}
 	defer sess.Close()
 	err = sess.Wait()
+
+	// Get the output
+	output := stdout.Bytes()
+
```
```suggestion
	output := stdout.Bytes()
```

README.md
TwiN, line 3039: Could you update this example to include an example of how to use `[BODY]`? You can change the command to something you've tried locally if you prefer

TwiN, line 3053:
```suggestion
```

TwiN: ignore this one

yansh97: I've added a minimal example containing "[BODY]". @TwiN

TwiN (approved):

TwiN:
@yansh97 Thank you for the contribution!

PR #1287, opened 2025-09-23
feat: adds multiple groups to the config

avendretter:
## Summary
As some endpoints can belong to multiple groups I have added the functionality to define groups as a list.
Endpoints will then be shown in the UI per Group. 

## Checklist

Screenshot 2025-09-23 at 12 45 20

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,7 +270,7 @@ If you want to test it locally, see [Docker](#docker).
 | `ui.buttons[].link`          | Link to open when the button is clicked.                                                                                                 | Required `""`              |
 | `ui.custom-css`              | Custom CSS                                                                                                                               | `""`                       |
 | `ui.dark-mode`               | Whether to enable dark mode by default. Note that this is superseded by the user's operating system theme preferences.                   | `true`                     |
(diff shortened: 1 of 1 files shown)


TwiN:
I wish you'd have created an issue first so we could discuss it prior to you spending the time to implement this.

This is not a change I'm open to making, as this would break _hundreds_ of scripts for me and would take me _weeks_ of work to support, not to mention this is a massive breaking change. I noticed you just grab the first group in the list to generate the key, but what if somebody calls the API with a key generated from the 2nd group? What about people who previously had, for instance, an endpoint named "website" in "group1" and an endpoint "website" in "group2" - these were previously considered separate unique keys, but now, you'd have to check if they're the same endpoint.

Beyond that, I also don't agree that an endpoint can belong to multiple groups. Adding a new field `tags` which users could filter by and endpoints could have more than 1 tag assigned to them is something, but making a breaking configuration change like this is not something I would do without significant preparation, notices, announcements, logs to warn people about the breaking change first, etc.

I'm really sorry you did all of this work for nothing, I wish you'd discussed it with me first, but do know that I appreciate the effort despite having to decline this feature.

Issue #1322, opened 2025-10-12
Add support for grace-period in heartbeats


onedr0p:
### Describe the feature request

Hi 👋 

I've been using the heartbeat feature for some time now and I was curious if you'd be open to adding a grace-period option?

```yaml
external-endpoints:
  - name: Heartbeat
    token: xyz
    heartbeat:
      interval: 5m
      grace-period: 30s
```

### Why do you personally want this feature to be implemented?

Right now I need to have my webhook heartbeat gatus every `2m30s`, for a `5m` interval. It would be much better if I could tell gatus to have a `30s` (or whatever) wait on the interval. That means I can keep my webhooks interval to `5m` but gatus won't report it as down if the heartbeat came in a few seconds too late.

### How long have you been using this project?

3y

### Additional information

One could argue why not set the interval to `5m30s` and have the webhook configured to send every `5m`  and that would effectively mean the same thing. However this is more about lining up the intervals in gatus and on the webhook for UX, while having a specific option for a grace period.

TwiN:
> One could argue why not set the interval to 5m30s and have the webhook configured to send every 5m and that would effectively mean the same thing. However this is more about lining up the intervals in gatus and on the webhook for UX, while having a specific option for a grace period.

@onedr0p if I understand correctly, you're saying that basically, if something is pushed at 5m15s, don't show a 5m failed status because it's within the `heartbeat interval + grace period` range, whereas if nothing is pushed after 5m30s, the failed result would be created as if it happened at the interval rather than at the `heartbeat interval+grace period`?

onedr0p:
@TwiN If i understand the second part of your message right. That's correct. 

Ideally I'd love to see this expanded to the way [healthchecks.io](https://healthchecks.io/) does this with the different types of statuses but a simple grace period is good enough for now IMO 

<img width="1080" height="1592" alt="Image" src="https://github.com/user-attachments/assets/1795cdf5-89d7-470d-aed1-69ecbcad6ce3" />

TwiN:
This would be great, _if_ Gatus' UI supported yellow status rather than just green or red. It could better convey the "late" aspect of exceeding the `heartbeat interval`, but not past `heartbeat interval + grace period`.

I feel like this doesn't really serve a real purpose right now, given that setting the interval to `5m30s` would have the same outcome (despite your thoughts under the `Additional information` section).

The way I see it, `heartbeat.interval` is _already_ a grace period, so it feels odd to have a grace period on a grace period.

[...]

PR #1323, opened 2025-10-12
feat(heartbeat): add support for grace-period configuration

onedr0p:
## Summary

This is a very basic implementation that only adds the interval and grace period and uses that as the effective interval. I could see this being expanded on in the future if needed 
 
Fixes: #1322 

## Checklist


onedr0p:
@TwiN what do you think?

TwiN:
@onedr0p Let's chat about the feature on the issue, to avoid having two separate discussions

onedr0p:
Closing until we can move forward with a better way.

PR #1328, opened 2025-10-13
Public private view

### LINKED ISSUE #638: Public/Private view

Hi, after discovering that Uptime Kuma doesn't support config files (WTF?), i luckily came across Gatus and knew immediately that this is what i actually want to run! <3 

The one feature i'd really like to see would be being able to keep the status page publicly accessible, while also enabling authentication and providing an extended list of checks that are not publicly visible - e.g. anonymous users see only website endpoints, authenticated users see services, private websites, DNS, etc.
This could be done by adding another field to the config dict for an endpoint, e.g. `public: False`.

I'm (obviously) self-hosting a metric shit ton of stuff and want to monitor everything, but i also don't want everyone to see everything i host. I could just run 2 instances of Gatus with separate config files, but that's not the "nice" solution. ;-)

1 hour

### Additional information

_No response_

loispostula:
## Summary

Implements public/private status visibility for endpoints and endpoint groups.

This allows a single Gatus instance to expose a public status page while keeping internal checks private.

The implementation follows the approach discussed in this PR:
```yaml
visibility:
  public: true
```
When endpoint-level security is enabled, unauthenticated users only see endpoints and groups marked as public. Authenticated users keep access to the full status page.

### Why

Today, Gatus is mostly all-or-nothing from a visibility point of view.

This is useful for teams that want to:

- expose public-facing service health
- keep internal/private checks hidden
- avoid running multiple Gatus instances just to separate public and private status pages

### Example

```yaml
security:
  basic:
    username: admin
    password-bcrypt-base64: ...
  level: endpoint

endpoints:
  - name: Public website
    url: https://example.com
    visibility:
      public: true

  - name: Internal service
    url: https://internal.example.com

```

In this example:

- unauthenticated users can see Public website
- authenticated users can see both Public website and Internal service

### Behavior
- Default behavior remains unchanged.
- `security.level`: global keeps the current all-or-nothing authentication behavior.
- `security.level`: endpoint enables public/private filtering.
- Endpoints are private by default unless explicitly marked with `visibility.public: true`.
- Endpoint groups are filtered based on the visibility of their endpoints.
- API responses are filtered consistently with the UI.

### Tests

This PR adds or updates coverage for:

- visibility filtering in memory storage
- visibility filtering in SQL storage
- endpoint-level security behavior
- config validation
- API behavior for public/private resources
- Basic authentication parsing edge cases

### Related issue

Closes #638


TwiN:
I appreciate your enthusiasm, but you should've really created an issue first so we could discuss the feature before you jump straight to the implementation and risk wasting your time 😅

I also want to add support for public and private endpoints in the future, but I don't have the time to spend on this right now and it's not currently at the top of my list of priorities. To keep it short, I'd like to have RBAC support for OIDC, with support for a public role. I don't want to just add a `public` parameter, because I would have to make a breaking change later.

loispostula:
@TwiN I thought I had open my pr against my own fork, oops.

Alright that make sense, I went this way after reading https://github.com/TwiN/gatus/issues/638 where a config by endpoint was what you added.

I'll go back to the issue and explain my `proposed` changes and let's take it from here.

This feature is what's missing for my org to onboard this tool

TwiN:
@loispostula My bad, I see there is already #1328 opened by somebody else, and I did already sort of give my thumbs up for it; I had completely forgotten about it 🤦

Feel free to keep working on it, and since I already gave the green lights, I'll make some time to look at it. 

Keep going with your approach of using the `public` parameter.

loispostula:
@TwiN I'm quite open to implement this in a different way, could you maybe check in #638 for explaining you vision?

TwiN:
Let's go with your approach for now. The only change I'd make at a glance is perhaps put `public` under `visibility`, e.g.

```yaml
endpoints:
  - name: ...
    visibility:
      public: true
```

This would allow us to add support for RBAC later without making breaking changes, as we could put it all under the `visibility` config.

loispostula:
Hey @TwiN, I've revived this PR and reworked it based on your feedback.

What changed since last time:

- Config now uses `visibility.public: true` as you suggested, so RBAC can be added later without breaking changes:
```
endpoints:
  - name: website
    url: https://example.com
    visibility:
      public: true
```
- security.level controls auth behavior: global (current default, backwards compatible) or endpoint (per-resource filtering)
- When level: endpoint, unauthenticated users see only public endpoints/suites. All API routes are accessible but responses are filtered.
- Badge, uptime, and response-time endpoints also respect visibility — no data leaks for private endpoints
- Frontend shows a login form when level: endpoint and user is not authenticated
- OIDC session cookie hardened with Secure, HttpOnly, SameSite=None

Test coverage added for:
- Visibility filtering (onlyPublic parameter across all store methods)
- IsAuthenticated Basic Auth parsing (8 edge cases)
- Config validation (level + auth provider requirements)

Happy to adjust the approach if you have feedback

loispostula:
Hello @TwiN before i rebase and fix the conflict, can you give me pointer on how to facilitate your review of this pr?

loispostula:
@TwiN I updated the PR description with a clearer summary, example config, behavior notes, and a suggested review order.

The main design question is:

Are you OK with `security.level: global|endpoint` as the switch for this behavior?

If yes, I’ll keep polishing this PR as-is. If you would prefer a different config shape, I can adjust that before spending more time on the implementation.

Issue #1091, opened 2025-05-05
feat(alerting): Allow configuring Discord message contents


object-Object:
### Describe the feature request

I'd like to be able to configure the message content (not the embed description/contents) of the Discord alert message. This is necessary to allow pinging users and/or roles in the message, as mentions in embeds do not send notifications.

Here's the current behaviour, achieved by setting `alerting.discord.default-alert.description: "<@150374394005422080>"` (note that this message did not send a ping/notification or highlight the message):

![Image](https://github.com/user-attachments/assets/7da2bf7a-6c6e-427d-a36d-5be932a33765)

Here's an example of how this would look if implemented (sent manually using an embed generator tool):

![Image](https://github.com/user-attachments/assets/a05f0175-c57f-4ea5-91a1-0a41d087dc9e)

### Why do you personally want this feature to be implemented?

The main reason is to be able to ping users/roles in alerts. This could potentially be useful for pinging specific users/teams for issues with their services. I mostly just want it because a notification badge is more noticeable than a normal channel message notification.

### How long have you been using this project?

A couple of days.

### Additional information

I think the relevant line of code is here:

https://github.com/TwiN/gatus/blob/fe214e9e251ddbcbc70e7b3e07837b7a9510bc59/alerting/provider/discord/discord.go#L145

PR #1335, opened 2025-10-15
feat(alerting): Add message-content parameter for Discord pings

aaldebs99:
## Summary

Relevant issue: #1091 
This PR allows the addition of a "prefix-message" to discord alerts. Discord doesn't allow pinging users/roles within embeds so the prefix-message aims to bring that functionality by sending a `Content` message alongside the embed that allows user/roles ping.

## Checklist


aaldebs99:
Example configuration:
```
endpoints:
  - name: "Google"
    group: "Infrastructure"
    url: "https://www.google.com"
    interval: 60s
    conditions:
      - "[STATUS] == 200"
      - "[RESPONSE_TIME] < 3000"
    alerts:
      - type: discord
        enabled: true
        send-on-resolved: true
        failure-threshold: 1
        provider-override:
          prefix-message: "<@12345678901>"
```

image

TwiN:
I've been thinking about this for a couple minutes, does calling this `prefix-message` really make sense? Terminology aside, perhaps the right thing here would be to extract the pings from the alert description, and automatically include them in the `Content` field so that the intended effect (pinging the users mentioned in the alert description) happens?

My concern here is that for most people, they may run into this issue (discord's embeds not pinging users), which means that when it does actually matter (i.e. there's an outage and the alert went out), they won't get the notification.

Sure, adding `prefix-message` would allow users to get around Discord's ping behavior inside of embeds, but realistically speaking, they'll only look into why the ping didn't work _after_ they ran into the issue, which means that Gatus would've effectively failed to ping/notify the user(s) when it mattered most.

For that reason, I think perhaps the right approach is to abstract the complexity of needing to specify `prefix-message` to ping users, and do that implicitly.

What do you think? Does it make sense?

aaldebs99:
> I've been thinking about this for a couple minutes, does calling this `prefix-message` really make sense? Terminology aside, perhaps the right thing here would be to extract the pings from the alert description, and automatically include them in the `Content` field so that the intended effect (pinging the users mentioned in the alert description) happens?
> 
> My concern here is that for most people, they may run into this issue (discord's embeds not pinging users), which means that when it does actually matter (i.e. there's an outage and the alert went out), they won't get the notification.
> 
> Sure, adding `prefix-message` would allow users to get around Discord's ping behavior inside of embeds, but realistically speaking, they'll only look into why the ping didn't work _after_ they ran into the issue, which means that Gatus would've effectively failed to ping/notify the user(s) when it mattered most.
> 
> For that reason, I think perhaps the right approach is to abstract the complexity of needing to specify `prefix-message` to ping users, and do that implicitly.
> 
> What do you think? Does it make sense?

I've seen it called prefix-message before in other projects, but I'm by no means married to that term. No one accused me of being creative with names LOL.

I definitely see your point though, it's a little annoying having to add a separate field just for pings and not being pinged in the description. For me personally, I would rather just add the ping message than have the default description just copy-pasted multiple times over with the small modification of adding the ping that I want. It might also not be the most elegant approach to extract pings. I'm personally more partial to keeping this approach and documenting it very well to make sure the user knows that pings don't work in the description section and they're expected to use prefix-message or whatever we end up calling it. 

Also just having the option to add an out-of-embed message, whatever it may be, might be useful for parsing or bot integrations.

I'm fine with having the pings extracted out of the embeds but it wouldn't be my preference.

TwiN:
I'm not super against it either, but I'm not a fan of prefix-message, however nitpicky that might sound. Maybe we should just name it `mentions`? Or is that misleading because it sounds like it'd be a list? Do you have any idea? I'm creatively spent, it's been a long week 😂

I thought just extracting the pings/mentions would make sense, because there's probably many alerting providers (e.g. slack, teams, etc) that have a similar behavior (no pings in embed), so taking the ping extraction approach might save us having to add new fields to multiple providers

aaldebs99:
> I'm not super against it either, but I'm not a fan of prefix-message, however nitpicky that might sound. Maybe we should just name it `mentions`? Or is that misleading because it sounds like it'd be a list? Do you have any idea? I'm creatively spent, it's been a long week 😂
> 
> I thought just extracting the pings/mentions would make sense, because there's probably many alerting providers (e.g. slack, teams, etc) that have a similar behavior (no pings in embed), so taking the ping extraction approach might save us having to add new fields to multiple providers

Lol we can just call it `content` or `message-content`, that would be the most "descriptive" term for it in terms of the discord API at least.

I'm not sure how other providers like Teams would work, but if we were to extract pings, that logic would probably differ between providers so we'd likely have to modify all the providers anyways to extract pings. For example for discord we can use `<@id>` but Team's pings looks something like `<at>username</at>` with their APIs. Slack seems to be the same as Discord and uses `<@id>`.

Again not totally against it but not my preference, if you'd prefer I try something like that let me know and I can give it a shot
If you're okay with `message-content` I can go ahead and rename it too

TwiN:
message-content sounds good to me! Just make sure to update the README to mention why this is needed (because pings work elsewhere)

aaldebs99:
> message-content sounds good to me! Just make sure to update the README to mention why this is needed (because pings work elsewhere)

Done! Glad we didn't have to get creative 😂

TwiN (approved):

TwiN:
@aaldebs99 Thank you for the contribution!

aaldebs99:
> @aaldebs99 Thank you for the contribution!

Thank you for the great app!

Issue #1339, opened 2025-10-16
Endpoint page just shows wait animation


amai2012:
### Describe the bug

I've added a couple of endpoints which have never been green (service just being created).
For one of them the details page just show the wait animation. And I cannot find entry in the log corresponding to that behavior.

### What do you see?

<img width="651" height="346" alt="Image" src="https://github.com/user-attachments/assets/00251c58-e703-44ef-8232-50e026549d4a" />

### What do you expect to see?

Regular endpoint page reflecting something like

<img width="275" height="185" alt="Image" src="https://github.com/user-attachments/assets/7c4e214d-0a8e-416e-9a14-aefec4934a39" />

TwiN:
Can you share the configuration of that buggy endpoint? You're really not giving me a lot to look at 😅

amai2012:
Here's a config (manually reduced from the complete config, not verified in real life, and some names changed...):
~~~ yaml
---
metrics: true

alerting:
  teams-workflows:
    webhook-url: "${WEBHOOK_MSTEAMS}"
    title: "&#x26D1; our-gatus-${ENV_NAME}"
    default-alert:
      description: "health check failed"
      failure-threshold: 2
      send-on-resolved: true
      success-threshold: 1

storage:
  type: sqlite
  path: /data/data.db

default-endpoint_https_secure: &defaults_https_secure
  interval: 5m
  client:
    insecure: false
    timeout: 500ms
  conditions:
    - "[STATUS] == 200"
    - "[CERTIFICATE_EXPIRATION] > 72h"

endpoints:
  - name: "Prometheus (Services+Federation)"
    url: "https://prometheus.some.where/-/healthy"
    group: "XXX Services (YYY)"
    <<: *defaults_https_secure
    conditions:
      - "[STATUS] == 200"
      - "[CERTIFICATE_EXPIRATION] > 72h"
      - "[BODY] == pat(Prometheus Server is Healthy.)"
~~~

TwiN:
Have you tried removing the + and ( ) from the name and group?

amai2012:
I've added a copy of the given endpoint with a difference name "Prometheus Services and Federation" - and the details page for that is working fine!

TwiN:
It's likely because of the special characters you had in the name/group

amai2012:
@TwiN  I can accept that the code isn't fixed/changed, but I'd ask for a doc fix at least:
[...]

PR #1340, opened 2025-10-16
fix(key): Support `(`, `)`, `+` and `&` as name/group

TwiN:
## Summary

Adds support for `(`, `)`, `+` and `&` as name/group

Relevant: #1339

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -366,7 +366,7 @@ or send an HTTP request:
 POST /api/v1/endpoints/{key}/external?success={success}&error={error}&duration={duration}
 ```
 Where:
-- `{key}` has the pattern `<GROUP_NAME>_<ENDPOINT_NAME>` in which both variables have ` `, `/`, `_`, `,`, `.` and `#` replaced by `-`.
+- `{key}` has the pattern `<GROUP_NAME>_<ENDPOINT_NAME>` in which both variables have ` `, `/`, `_`, `,`, `.`, `#`, `(`, `)`, `+` and `&` replaced by `-`.
   - Using the example configuration above, the key would be `core_ext-ep-test`.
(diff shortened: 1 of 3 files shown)


amai2012:
@TwiN  That indeed fixed https://github.com/TwiN/gatus/issues/1339 - thank you.

Side-effect: the history for endpoints with affected names has been reset.

TwiN:
The side effect is acceptable, thank you for confirming.

markormesher:
Out of interest this broke some working checks that contained parentheses but no plus sign. I have checks called `foo bar (host name)` which had the history dropped and external URLs that no longer work, causing some alerts. Easily fixed, just needed to replace the parens with dashes, and the history is no great loss - mostly just posing this in case anyone else comes across the same issue!

TwiN:
@markormesher Hm, did the detail page with parentheses previously work?

That's a bit of an oversight on my part, my bad

markormesher:
Yep, the dashboard view and detail views both worked fine with parens in the names. I just rolled back to 5.26.0 to confirm, and other than dropping the history each time I rolled back/forwards everything worked the same. I don't have any checks with `+` in the name so I didn't test that, but I'd guess it's because `()` are handled normally in URLs but `+` is interpreted differently.

Without external checks with heartbeat intervals I probably wouldn't have even noticed, I only saw it because their URL format changed which caused their heartbeats to start expiring.

TwiN:
I decided to revert the support for `(` and `)` since it already worked. It's a shame that this will break some that were staying closely up-to-date with the latest Gatus version like you were, but for the greater good of the community, I'll revert it to minimize those impacted.

Thank you for bringing this to my attention @markormesher, this kind of situation where you lose your history is definitely not the kind of experience I want people to have with Gatus :/

Issue #1255, opened 2025-09-11
Option: Expand newly collapsed-by-default groups


tmeuze:
### Describe the feature request

Hello, 

With the introduction of #1212 , groups are collapsed by default. 

It would be nice to have a configurable `collapse-by-default` key to actually undo this behavior. In certain instances such as mine, having all groups expanded by default is ideal. 

Thank you!

### Why do you personally want this feature to be implemented?

For use by less technically-inclined users that may not take the time to expand collapsed groups. Also, I think it just looks better, and to have an option fits everyone's desires!

### How long have you been using this project?

<1 mo.

Exagone313:
I just updated to the new version and it annoyed me too, so I wrote a very hacky patch to remove the collapse feature in the mean time.

<details>

<summary>Patch</summary>

```diff
diff --git a/web/app/src/views/Home.vue b/web/app/src/views/Home.vue
index 28182eae..359eb202 100644
--- a/web/app/src/views/Home.vue
+++ b/web/app/src/views/Home.vue
@@ -55,12 +55,9 @@
           <div v-for="(endpoints, group) in paginatedEndpoints" :key="group" class="endpoint-group border rounded-lg overflow-hidden">
             <!-- Group Header -->
             <div 
-              @click="toggleGroupCollapse(group)"
-              class="endpoint-group-header flex items-center justify-between p-4 bg-card border-b cursor-pointer hover:bg-accent/50 transition-colors"
+              class="endpoint-group-header flex items-center justify-between p-4 bg-card border-b"
             >
               <div class="flex items-center gap-3">
-                <ChevronDown v-if="uncollapsedGroups.has(group)" class="h-5 w-5 text-muted-foreground" />
-                <ChevronUp v-else class="h-5 w-5 text-muted-foreground" />
                 <h2 class="text-xl font-semibold text-foreground">{{ group }}</h2>
               </div>
               <div class="flex items-center gap-2">
@@ -73,7 +70,7 @@
             </div>
             
             <!-- Group Content -->
-            <div v-if="uncollapsedGroups.has(group)" class="endpoint-group-content p-4">
+            <div class="endpoint-group-content p-4">
               <div class="grid gap-3 grid-cols-1 sm:grid-cols-2 lg:grid-cols-3">
                 <EndpointCard
                   v-for="endpoint in endpoints"
@@ -141,7 +138,7 @@
 <script setup>
 /* eslint-disable no-undef */
 import { ref, computed, onMounted } from 'vue'
[...]

PR #1344, opened 2025-10-17
feat: add configureable default for collapsing statuses

rickklaasboer:
## Summary

Fixes https://github.com/TwiN/gatus/issues/1255. I think this should be all to achieve this setting, if not, please tell! 

## Checklist

### DIFF
--- a/web/app/src/views/Home.vue
+++ b/web/app/src/views/Home.vue
@@ -536,29 +536,27 @@ const toggleGroupCollapse = (groupName) => {
   localStorage.removeItem('gatus:collapsed-groups') // Remove old key if it exists
 }
 
-// Initialize group collapse state from localStorage or config defaults.
-// Called when user enables group view via SearchBar.
 const initializeCollapsedGroups = () => {
-  // First, try to restore user's previous collapse preferences from localStorage
+  // Get saved uncollapsed groups from localStorage
   try {
     const saved = localStorage.getItem('gatus:uncollapsed-groups')
     if (saved) {
       uncollapsedGroups.value = new Set(JSON.parse(saved))
       return
     }
+    // If no saved state, fall through to apply config defaults
   } catch (e) {
     console.warn('Failed to parse saved uncollapsed groups:', e)
     localStorage.removeItem('gatus:uncollapsed-groups')
+    // On error, fall through to apply config defaults
   }
 
-  // No saved state found - apply config default (collapseByDefault)
+  // Apply config default (collapseByDefault)
   if (!collapseByDefault.value) {
-    // Expand all groups by default
     const groups = Object.keys(combinedGroups.value || {})
-    uncollapsedGroups.value = new Set(groups)
+    uncollapsedGroups.value = new Set(groups) // expanded by default
   } else {
-    // Collapse all groups by default (empty set = nothing uncollapsed)
-    uncollapsedGroups.value = new Set()
+    uncollapsedGroups.value = new Set() // collapsed by default
   }
 }
 


web/app/src/views/Home.vue
```diff
@@ -17,13 +17,26 @@
...
             </Button>
+
             <Button variant="ghost" size="icon" @click="refreshData" title="Refresh data">
               <RefreshCw class="h-5 w-5" />
             </Button>
+
+            <Button
+              variant="ghost"
+              size="icon"
+              @click="toggleDefaultCollapse"
```
TwiN: This is just for groups, right? Meaning it only applies if the user sorts by group.

Isn't this button kind of useless? I think I would rather have this as a configuration just in the configuration, and not allow people to toggle it in the UI

config/ui/ui.go
```diff
@@ -40,6 +42,7 @@ type Config struct {
 	DarkMode        *bool    `yaml:"dark-mode,omitempty"`         // DarkMode is a flag to enable dark mode by default
 	DefaultSortBy   string   `yaml:"default-sort-by,omitempty"`   // DefaultSortBy is the default sort option ('name', 'group', 'health')
 	DefaultFilterBy string   `yaml:"default-filter-by,omitempty"` // DefaultFilterBy is the default filter option ('none', 'failing', 'unstable')
+	DefaultCollapse *bool     `yaml:"default-collapse,omitempty"`  // DefaultCollapse is a flag to enable/disable collapsing of groups by default
```
TwiN: Since this only applies to groups, I think `DefaultGroupCollapsed` would make more sense

web/app/src/views/Home.vue
rickklaasboer: Fair! Didn't think about that one, since I've always got it set to groups. Will remove the button :)

TwiN: It was a good initiative though, I appreciate you taking the time to attempt an implementation that doesn't break consistency!

web/app/package-lock.json
```diff
@@ -8,14 +8,10 @@
       "name": "gatus",
       "version": "4.0.0",
       "dependencies": {
```
TwiN, line 10: What's going on there? 🧐

rickklaasboer: Mhm, don't know actually. Maybe I installed dependencies wrong?

TwiN: I'm not sure 🧐 did you make any changes to the package.json locally?

rickklaasboer: Nope, no changes on my end. I could just reset the changes to HEAD and recommit? 🤷‍♂️

TwiN: Yes please. I'll look into whether I'm due for updating the frontend deps separately from this PR

config/ui/ui.go
```diff
@@ -80,6 +83,7 @@ func GetDefaultConfig() *Config {
 		DarkMode:               &defaultDarkMode,
 		DefaultSortBy:          defaultSortBy,
 		DefaultFilterBy:        defaultFilterBy,
+		DefaultGroupCollapse:        &defaultGroupCollapse,
 		MaximumNumberOfResults: storage.DefaultMaximumNumberOfResults,
```
TwiN, line 103: Please fix the alignment here

config/ui/ui.go
```diff
@@ -21,11 +21,13 @@ const (
...
 
 var (
-	defaultDarkMode = true
+	defaultDarkMode  = true
+	defaultGroupCollapse  = true
 
 	ErrButtonValidationFailed = errors.New("invalid button configuration: missing required name or link")
 	ErrInvalidDefaultSortBy   = errors.New("invalid default-sort-by value: must be 'name', 'group', or 'health'")
 	ErrInvalidDefaultFilterBy = errors.New("invalid default-filter-by value: must be 'none', 'failing', or 'unstable'")
+	ErrInvalidDefaultGroupCollapse = errors.New("invalid default-group-collapse value: must be a boolean")
```
TwiN, line 35: Please fix the alignment here

config/ui/ui.go
```diff
@@ -40,6 +42,7 @@ type Config struct {
 	DarkMode        *bool    `yaml:"dark-mode,omitempty"`         // DarkMode is a flag to enable dark mode by default
 	DefaultSortBy   string   `yaml:"default-sort-by,omitempty"`   // DefaultSortBy is the default sort option ('name', 'group', 'health')
 	DefaultFilterBy string   `yaml:"default-filter-by,omitempty"` // DefaultFilterBy is the default filter option ('none', 'failing', 'unstable')
+	DefaultGroupCollapse *bool     `yaml:"default-group-collapse,omitempty"`  // DefaultGroupCollapse is a flag to enable/disable collapsing of groups by default
```
TwiN: Please fix the alignment here

web/app/src/views/Home.vue
```diff
@@ -503,21 +523,32 @@ const toggleGroupCollapse = (groupName) => {
...
+    const groups = Object.keys(combinedGroups.value || {})
+    uncollapsedGroups.value = new Set(groups) // expanded by default
+  } else {
+    uncollapsedGroups.value = new Set() // collapsed by default
   }
 }
 
+watch(() => combinedGroups.value, () => {
+  if (!localStorage.getItem('gatus:uncollapsed-groups')) {
+    initializeCollapsedGroups()
```
TwiN: Was this (`initializeCollapsedGroups`) not called before? Or am I missing something?
Looks like `initializeCollapsedGroups` was previously dead code

web/app/src/views/Home.vue
```diff
@@ -503,21 +523,32 @@ const toggleGroupCollapse = (groupName) => {
...
     localStorage.removeItem('gatus:uncollapsed-groups')
-    // On error, uncollapsedGroups stays empty (all collapsed by default)
+  }
+
+  if (!collapseByDefault.value) {
+    const groups = Object.keys(combinedGroups.value || {})
+    uncollapsedGroups.value = new Set(groups) // expanded by default
+  } else {
+    uncollapsedGroups.value = new Set() // collapsed by default
   }
```
TwiN, line 560: Can you add some comments here? I see you removed all comments, it's a little hard to follow what the code is doing.

TwiN (changes requested):

ps610:
Is there any chance of getting this PR through?

PR #1346, opened 2025-10-18
fix(client): update icmp/ping logic to determine pinger privileged mode

h3mmy:
## Summary

https://github.com/TwiN/gatus/issues/697#issuecomment-2081700989
Based around fix attempted in https://github.com/TwiN/gatus/pull/748 since the experimental tagged image worked for pings in non-privileged containers. Apparently it broke for users in root containers, so this PR includes a check for root privilege, albeit a naive one.

As of v5.26.0, ICMP checks still do not work for my (non-root) deployment regardless of whether I add `CAP_NET_RAW`

> Checking individual capabilities will require adding "kernel.org/pub/linux/libs/security/libcap/cap" as a dependency. I'm not sure if you want to add more dependencies for a small check, so I can use some naive logic to check if the app is running as root like checking the EUID == 0
>
> Since SetPrivileged needs to be set to false for non-privileged processes running on linux or darwin, I figured this is a reasonable check unless you are wanting a more precise check with the extra dependency. 

## Checklist

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -373,7 +373,6 @@ func ShouldRunPingerAsPrivileged() bool {
 	if runtime.GOOS == "windows" {
 		return true
 	}
-
 	// To actually check for cap_net_raw capabilities, we would need to add "kernel.org/pub/linux/libs/security/libcap/cap" to gatus.
 	// Or use a syscall and check for permission errors, but this requires platform specific compilation
 	// As a backstop we can simply check the effective user id and run as privileged when running as root


h3mmy:
I can work on adding unit tests once an approach is finalized. I always feel weird making reviews without tests.

h3mmy:
I played around with precisely checking for `CAP_NET_RAW` and `CAP_NET_ADMIN` but a consequence of relying on `"kernel.org/pub/linux/libs/security/libcap/cap"` is tht `CGO_ENABLED` would need to be set to 1 instead of 0. I hesitate to bring in a new dependency that will change the build process that way. 

I would expect folks running as root without having the uid set to 0 is not super common. In order to be thorough, I decided to check for cap_net_raw by simply testing if opening a raw socket throws a permission error. Let me know if you like or dislike it and I can re-adjust

h3mmy:
After some testing of the syscall method, it came to light that it won't compile on windows due to the different flags, so I discarded that route and fell back to the initial implementation.

h3mmy:
I built my own image for testing purposes. Confirmed it works. 

 repository: `ghcr.io/h3mmy/gatus`
 tag: `testing@sha256:87f8e8ea0073d69dd6eecacecf93c55699f2ee9e2e062e547ce4b10b536743d1`

client/client.go
```diff
@@ -364,6 +361,26 @@ func Ping(address string, config *Config) (bool, time.Duration) {
 	return true, 0
 }
 
+// ShouldUsePrivilegedPinger will determine whether or not to run pinger in privileged mode.
+// It should be set to privileged when running as root, and always on windows. See https://pkg.go.dev/github.com/macrat/go-parallel-pinger#Pinger.SetPrivileged
+func ShouldRunPingerAsPrivileged() bool {
+	// Set the pinger's privileged mode to false for darwin
```
heathcliff26, line 366: I'm no expert in Darwin, is it guaranteed that
```
os.Geteuid() == 0
```
will always return `false` on Darwin? Otherwise this should be added as a special case here.

client/client.go
```diff
@@ -364,6 +361,26 @@ func Ping(address string, config *Config) (bool, time.Duration) {
...
+	// See https://github.com/TwiN/gatus/pull/748 and https://github.com/TwiN/gatus/issues/697#issuecomment-2081700989
+	//
+	// Note that for this to work on Linux, Gatus must run with sudo privileges. (in certain cases)
+	// See https://github.com/prometheus-community/pro-bing#linux
+	if runtime.GOOS == "windows" {
+		return true
+	}
+
+	// To actually check for cap_net_raw capabilities, we would need to add "kernel.org/pub/linux/libs/security/libcap/cap" to gatus.
+	// Or use a syscall and check for permission errors, but this requires platform specific compilation
```
heathcliff26, line 377: Is it architecture or os specific implementations?
Because OS specific would be fairly easy in this case.

client/client.go
```diff
@@ -343,12 +344,8 @@ func Ping(address string, config *Config) (bool, time.Duration) {
...
 	pinger.Count = 1
 	pinger.Timeout = config.Timeout
-	// Set the pinger's privileged mode to true for every GOOS except darwin
-	// See https://github.com/TwiN/gatus/issues/132
-	//
-	// Note that for this to work on Linux, Gatus must run with sudo privileges.
-	// See https://github.com/prometheus-community/pro-bing#linux
-	pinger.SetPrivileged(runtime.GOOS != "darwin")
+
+	pinger.SetPrivileged(ShouldRunPingerAsPrivileged())
```
```suggestion
	pinger.SetPrivileged(ShouldRunPingerAsPrivileged())
```
Please remove unnecessary newlines

client/client.go
```diff
@@ -364,6 +361,26 @@ func Ping(address string, config *Config) (bool, time.Duration) {
...
+	// See https://github.com/TwiN/gatus/issues/132
+	// linux should also be set to false, but there are potential complications
+	// See https://github.com/TwiN/gatus/pull/748 and https://github.com/TwiN/gatus/issues/697#issuecomment-2081700989
+	//
+	// Note that for this to work on Linux, Gatus must run with sudo privileges. (in certain cases)
+	// See https://github.com/prometheus-community/pro-bing#linux
+	if runtime.GOOS == "windows" {
+		return true
+	}
+
```
```suggestion
```

client/client.go
```diff
@@ -364,6 +361,26 @@ func Ping(address string, config *Config) (bool, time.Duration) {
 	return true, 0
 }
 
+// ShouldUsePrivilegedPinger will determine whether or not to run pinger in privileged mode.
+// It should be set to privileged when running as root, and always on windows. See https://pkg.go.dev/github.com/macrat/go-parallel-pinger#Pinger.SetPrivileged
+func ShouldRunPingerAsPrivileged() bool {
```
```suggestion
// ShouldRunPingerAsPrivileged will determine whether or not to run pinger in privileged mode.
// It should be set to privileged when running as root, and always on windows. See https://pkg.go.dev/github.com/macrat/go-parallel-pinger#Pinger.SetPrivileged
func ShouldRunPingerAsPrivileged() bool {
```
First word in the comment should match the name of the function

client/client.go
```diff
@@ -364,6 +361,26 @@ func Ping(address string, config *Config) (bool, time.Duration) {
...
+	// Note that for this to work on Linux, Gatus must run with sudo privileges. (in certain cases)
+	// See https://github.com/prometheus-community/pro-bing#linux
+	if runtime.GOOS == "windows" {
+		return true
+	}
+
+	// To actually check for cap_net_raw capabilities, we would need to add "kernel.org/pub/linux/libs/security/libcap/cap" to gatus.
+	// Or use a syscall and check for permission errors, but this requires platform specific compilation
+	// As a backstop we can simply check the effective user id and run as privileged when running as root
+	return os.Geteuid() == 0
```
TwiN, line 379: I'm ok with giving this a try in `latest` and see what happens, but this may break people whose container is configured to use a custom user id.

TwiN (changes requested):

TwiN:
Couple comments, but I've tested it on my end and it seems to work fine on my Kubernetes cluster.

client/client.go
h3mmy: If using as a custom user ID with cap net raw, it should be running as privileged to work. I'd be curious if anyone is running it as unprivileged with CAP_NET_RAW and have it be working. I'm happy to work on something more thorough, but it will likely involve some sort of libc linking requirement unless I figure out a different approach.

client/client.go
h3mmy: If someone is running this as root on darwin, then they'd want the Privileged mode as per my understanding.

client/client.go
h3mmy: The low level libraries needed for creating a raw socket on POSIX systems require C-linking. This project currently builds with CGO not enabled, and I am reluctant to make changes that would impact the build process as this will introduce some additional complexity around building for multiple systems and thus increase the scope. 

If you have suggestions that don't require CGO, I'm open to trying them out.

TwiN:
@h3mmy Thank you for the contribution!

enorasec:
Can confirm this is working for me as well. ARM deployment on an Oracle Kubernetes Engine Cluster. Thank you so much @h3mmy!

TwiN:
Thank you for confirming, I appreciate it! 

Having to validate changes alone isn't as scalable as it once used to be 🤣

roughnecks:
Ping is broken for me now. I'm compiling gatus myself and run it with systemd:

```
systemctl cat gatus.service
# /etc/systemd/system/gatus.service
[Unit]
Description=Gatus daemon
After=network-online.target

[Service]
Type=simple
ExecStart=/home/user/go/bin/gatus
Environment=GATUS_CONFIG_PATH="/home/user/config/gatus/"
Environment=GATUS_LOG_LEVEL="WARN"
CapabilityBoundingSet=CAP_NET_RAW
AmbientCapabilities=CAP_NET_RAW
NoNewPrivileges=yes
Restart=always
RestartSec=20s
User=user
WorkingDirectory=/home/user/

[Install]
WantedBy=multi-user.target
```

h3mmy:
> Ping is broken for me now. I'm compiling gatus myself and run it with systemd:
> 
> ```
> systemctl cat gatus.service
> # /etc/systemd/system/gatus.service
> [Unit]
> Description=Gatus daemon
> After=network-online.target
> 
> [Service]
> Type=simple
> ExecStart=/home/user/go/bin/gatus
> Environment=GATUS_CONFIG_PATH="/home/user/config/gatus/"
> Environment=GATUS_LOG_LEVEL="WARN"
> CapabilityBoundingSet=CAP_NET_RAW
> AmbientCapabilities=CAP_NET_RAW
> NoNewPrivileges=yes
> Restart=always
> RestartSec=20s
> User=user
> WorkingDirectory=/home/user/
> 
> [Install]
> WantedBy=multi-user.target
> ```

@roughnecks Would you try running it without the `CAP_NET_RAW` capabilities? You would only need to remove it from the `AmbientCapabilities`

This seems to be exactly the edge case not covered in the PR. I can work on updating it to include this, although it will be tricky without CGO. But I'd like to confirm that removing that ambient capability resolves this for you

roughnecks:
Hey, I just commented out the AmbientCapabilities, still no dice.

h3mmy:
@roughnecks To confirm, you commented it out and restarted the systemd service to check, right? 

What is your output for `sysctl net.ipv4.ping_group_range`? Some distros make this range too narrow

You can allow your uid to use unprivileged ping by using something like 
`echo 'net.ipv4.ping_group_range = 0 2147483647' | sudo tee -a /etc/sysctl.conf`

The main thing here is just that the guid for the Linux user needs to be less than the right hand number.

The other thing you can try is changing the `User=user` to `User=root` (this ensures EUID is 0 and uses privileged mode

I probably should have added  debug logging to check the EUID.

It's super late for me at the moment, but I'll work on this in the morning. I might be able to use the deprecated syscall method to test capabilities to make config changes not a required thing.

roughnecks:
yeah, "daemon-reloaded", restarted service.

I changed this: 'net.ipv4.ping_group_range = 0 2147483647' and it's fixed now.

I even had that line already in my sysctl.conf with a comment about gatus, but I believe at the time I thought it was better like I did with the systemd unit.

Oh well. Thanks

h3mmy:
@TwiN I want to understand how you want this prioritized. It seems like this scenario will primarily impact people who needed to add a CAP_NET_RAW to get pinger to work in the first place (technically as a workaround). Should I prioritize trying to find a way to put together a raw packet without CGO in order to make the logic more precise, or focus on updating documentation with an explanation of config?

TwiN:
@h3mmy If the fix you made impacts people who previously had to do some shenanigans to get it to work before, then it's fine to me, there's no need to fix anything. That said, a note in the documentation saying something along the lines of "Prior to v5.31.0, some environments required setting CAP_NET_RAW to [...]. As of v5.31.0, this is no longer necessary. See #1346." Would be greatly appreciated

h3mmy:
Thanks for the clarification! Note added. https://github.com/TwiN/gatus/pull/1384

PR #1349, opened 2025-10-18
feat(ui): show reason for failure in endpoint events

perfectra1n:
## Summary

This PR shows the reason as to why an Endpoint failed, instead of just "it failed". E.g. right now you look at something like this:
image

and you ask yourself "But why did it fail? What criteria failed?"

So this is what I've designed at this point:
image

## Checklist

- [ x ] Tested and/or added tests to validate that the changes work as intended, if applicable.

### DIFF
--- a/config/endpoint/event.go
+++ b/config/endpoint/event.go
@@ -1,8 +1,18 @@
 package endpoint
 
 import (
+	"strings"
 	"time"
 )
+
+// FailureReason represents a failure with its timestamp
+type FailureReason struct {
+	// Description is the error message or failed condition
+	Description string `json:"description"`
+
+	// Timestamp is when this failure was first observed
+	Timestamp time.Time `json:"timestamp"`
+}
 
 // Event is something that happens at a specific time
 type Event struct {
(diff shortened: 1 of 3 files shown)


perfectra1n:
I'm not exactly sold on the look just yet, but I was curious to see what your opinion was, or what your appetite for such a thing would be. As a user I would like to know "why" and "what" failed with such a program, but you may have different intentions.

TwiN:
This isn't a bad idea, but what if there are multiple reasons for the endpoint being unhealthy? For instance, what if, for the first 20 minutes, it's unhealthy because of the status, but from 20 to 40 minutes, it's unhealthy because of the response time?

perfectra1n:
I'm sure that Health criteria can become quite complex - so if we have an example EndpointA with health criteria: criteria1, criteria2, and criteria3...

If it fails at first due to criteria1, then after 10 minutes (or some other time-based criteria requirement) fail due to criteria2, I've seen some platforms have the UI reflect that e.g.:

"failed due to criteria1"
"10 minutes later"
"failed due to criteria2"

What do you think of something like that? Or instead, we could show just all criteria that failed initially (or after X amount of time) and leave it to the user's discretion :)

TwiN:
Perhaps having it as a set (unique values) would be sufficient for now.

Just a heads up, the current implementation is missing a lot of bits and pieces, the biggest one being persistence.

perfectra1n:
image

Perhaps something like this then, where we aggregate the unique failures and show the timestamp of failure?

config/endpoint/event.go
```diff
@@ -1,16 +1,40 @@
...
+	FailedConditions []string `json:"failedConditions,omitempty"`
+
+	// ErrorReasons is a list of errors with their timestamps (for UNHEALTHY events)
+	ErrorReasons []FailureReason `json:"errorReasons,omitempty"`
+
+	// FailedConditionReasons is a list of failed conditions with their timestamps (for UNHEALTHY events)
+	FailedConditionReasons []FailureReason `json:"failedConditionReasons,omitempty"`
 }
 
 // EventType is, uh, the types of events?
```
TwiN, line 40: lmao, I forgot about this

TwiN:
> Perhaps something like this then, where we aggregate the unique failures and show the timestamp of failure?

This is a great idea, but there's a few things we'll have to take into consideration, such as everything under `endpoints[].ui.*` (e.g. `hide-port`, `dont-resolve-failed-conditions`, `hide-url`, etc.)

If a user configures it so certain things are hidden, they also shouldn't be reflected in the event, as it may be considered sensitive to the user.

Also, I want to point out that I'm working on replacing the chart as we speak, and I'm thinking of perhaps incorporating events (specifically failed events) in the chart to show when the service is unhealthy, so whatever we add in this PR could also be included as information in the chart (i.e. show the failed reasons on hover)

PR #1350, opened 2025-10-18
feat(api-tokens): add optional api-tokens

aaldebs99:
## Summary

Attempt 2!

This PR aims to implement optional API tokens used to access the API when basic or OIDC auth is enabled.

Example configuration:

```yaml
security:
  api:
    tokens:
      - "my-secret-token-123"
      - "${API_TOKEN_FROM_ENV}"
      - "my-secret-token-2
```

This allows you to provide one of those tokens to API calls with the `Authorization` header:
```
curl http://localhost:8080/api/v1/config
{"announcements":[],"authenticated":false,"oidc":false}⏎

curl http://localhost:8080/api/v1/config -H "Authorization: Bearer my-secret-token-123"
{"announcements":[],"authenticated":true,"oidc":false}⏎

curl http://localhost:8080/api/v1/config -H "Authorization: Bearer my-secret-token-2"
{"announcements":[],"authenticated":true,"oidc":false}
```

## Checklist

### DIFF
--- a/security/api-tokens.go
+++ b/security/api-tokens.go
@@ -1,6 +1,9 @@
 package security
 
-import "errors"
+import (
+	"errors"
+	"slices"
+)
 
 var (
 	ErrAPITokensEmpty = errors.New("security.api.tokens must not contain empty tokens")
@@ -31,10 +34,5 @@ func (c *APIConfig) IsValid(token string) bool {
 	if c == nil || len(c.Tokens) == 0 || len(token) == 0 {
 		return false
 	}
-	for _, validToken := range c.Tokens {
-		if validToken == token {
-			return true
-		}
-	}
-	return false
+	return slices.Contains(c.Tokens, token)
 }
--- a/security/api-tokens.go
+++ b/security/api-tokens.go
@@ -1,0 +1,40 @@
+package security
+
+import "errors"
+
+var (
+	ErrAPITokensEmpty = errors.New("security.api.tokens must not contain empty tokens")
+)
+
+// APIConfig is the configuration for API token authentication
+type APIConfig struct {
+	// Tokens is a list of valid API tokens for authentication
+	Tokens []string `yaml:"tokens"`
+}
+
+// Validate validates the APIConfig
+func (c *APIConfig) Validate() error {
+	if c == nil {
+		return nil
+	}
+	// Ensure no empty tokens are configured
+	for _, token := range c.Tokens {
+		if len(token) == 0 {
+			return ErrAPITokensEmpty
+		}
+	}
+	return nil
+}
+
+// IsValid checks if a given token is valid
+func (c *APIConfig) IsValid(token string) bool {
+	if c == nil || len(c.Tokens) == 0 || len(token) == 0 {
+		return false
+	}
+	for _, validToken := range c.Tokens {
+		if validToken == token {
+			return true
+		}
+	}
+	return false
+}
--- a/security/config.go
+++ b/security/config.go
@@ -58,8 +58,8 @@ func (c *Config) ApplySecurityMiddleware(router fiber.Router) error {
 		router.Use(func(ctx *fiber.Ctx) error {
 			// Check for Bearer token
 			authHeader := ctx.Get("Authorization")
-			if strings.HasPrefix(authHeader, "Bearer ") {
-				token := strings.TrimSpace(strings.TrimPrefix(authHeader, "Bearer "))
+			if after, ok := strings.CutPrefix(authHeader, "Bearer "); ok {
+				token := strings.TrimSpace(after)
 				if c.API.IsValid(token) {
 					// Valid API token - mark as authenticated and skip other auth middleware
 					ctx.Locals("api_token_authenticated", true)
@@ -137,8 +137,8 @@ func (c *Config) IsAuthenticated(ctx *fiber.Ctx) bool {
 	// Check for API token authentication first
 	if c.API != nil && len(c.API.Tokens) > 0 {
 		authHeader := ctx.Get("Authorization")
-		if strings.HasPrefix(authHeader, "Bearer ") {
-			token := strings.TrimSpace(strings.TrimPrefix(authHeader, "Bearer "))
+		if after, ok := strings.CutPrefix(authHeader, "Bearer "); ok {
+			token := strings.TrimSpace(after)
 			if c.API.IsValid(token) {
 				return true
 			}
--- a/security/config.go
+++ b/security/config.go
@@ -3,6 +3,7 @@ package security
 import (
 	"encoding/base64"
 	"net/http"
+	"strings"
 
 	g8 "github.com/TwiN/g8/v2"
 	"github.com/TwiN/logr"
@@ -22,13 +23,14 @@ const (
 type Config struct {
 	Basic *BasicConfig `yaml:"basic,omitempty"`
 	OIDC  *OIDCConfig  `yaml:"oidc,omitempty"`
+	API   *APIConfig   `yaml:"api,omitempty"`
 
 	gate *g8.Gate
 }
 
 // ValidateAndSetDefaults returns whether the security configuration is valid or not and sets default values.
 func (c *Config) ValidateAndSetDefaults() bool {
-	return (c.Basic != nil && c.Basic.isValid()) || (c.OIDC != nil && c.OIDC.ValidateAndSetDefaults())
+	return (c.Basic != nil && c.Basic.isValid()) || (c.OIDC != nil && c.OIDC.ValidateAndSetDefaults()) || (c.API != nil && c.API.Validate() == nil && len(c.API.Tokens) > 0)
 }
 
 // RegisterHandlers registers all handlers required based on the security configuration
@@ -46,7 +48,30 @@ func (c *Config) RegisterHandlers(router fiber.Router) error {
 // ApplySecurityMiddleware applies an authentication middleware to the router passed.
 // The router passed should be a sub-router in charge of handlers that require authentication.
 func (c *Config) ApplySecurityMiddleware(router fiber.Router) error {
-	if c.OIDC != nil {
+	// Use a wrapper middleware that checks for API token first, then delegates to existing auth
+	hasAPITokens := c.API != nil && len(c.API.Tokens) > 0
+	hasOIDC := c.OIDC != nil
+	hasBasic := c.Basic != nil
+
+	// If we have API tokens, add a pre-check middleware
+	if hasAPITokens {
+		router.Use(func(ctx *fiber.Ctx) error {
+			// Check for Bearer token
+			authHeader := ctx.Get("Authorization")
+			if strings.HasPrefix(authHeader, "Bearer ") {
+				token := strings.TrimSpace(strings.TrimPrefix(authHeader, "Bearer "))
+				if c.API.IsValid(token) {
+					// Valid API token - mark as authenticated and skip other auth middleware
+					ctx.Locals("api_token_authenticated", true)
+					return ctx.Next()
+				}
+			}
+			// No valid Bearer token, continue to next middleware
+			return ctx.Next()
+		})
+	}
+
+	if hasOIDC {
 		// We're going to use g8 for session handling
 		clientProvider := g8.NewClientProvider(func(token string) *g8.Client {
 			if _, exists := sessions.Get(token); exists {
@@ -64,8 +89,15 @@ func (c *Config) ApplySecurityMiddleware(router fiber.Router) error {
 		// TODO: g8: Add a way to update cookie after? would need the writer
 		authorizationService := g8.NewAuthorizationService().WithClientProvider(clientProvider)
 		c.gate = g8.New().WithAuthorizationService(authorizationService).WithCustomTokenExtractor(customTokenExtractorFunc)
-		router.Use(adaptor.HTTPMiddleware(c.gate.Protect))
-	} else if c.Basic != nil {
+
+		// Wrap the g8 middleware to skip if already authenticated via API token
+		router.Use(func(ctx *fiber.Ctx) error {
+			if ctx.Locals("api_token_authenticated") == true {
+				return ctx.Next()
+			}
+			return adaptor.HTTPMiddleware(c.gate.Protect)(ctx)
+		})
+	} else if hasBasic {
 		var decodedBcryptHash []byte
 		if len(c.Basic.PasswordBcryptHashBase64Encoded) > 0 {
 			var err error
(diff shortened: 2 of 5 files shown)


aaldebs99:
@TwiN I've been running this version for 3 months now and it's been great. Thoughts about merging?

TwiN:
@aaldebs99 Sorry about the delay, I haven't had enough time to look through the code yet, but this looks super interesting.

@PythonGermany what do you think? (No pressure, feel free to ignore or look at it when/if you have time)

security/api-tokens.go
```diff
@@ -0,0 +1,40 @@
...
+// IsValid checks if a given token is valid
+func (c *APIConfig) IsValid(token string) bool {
+	if c == nil || len(c.Tokens) == 0 || len(token) == 0 {
+		return false
+	}
+	for _, validToken := range c.Tokens {
+		if validToken == token {
+			return true
+		}
+	}
```
PythonGermany:
```suggestion
	return slices.Contains(c.Tokens, token)
```
Suggested by gopls check

security/config.go
```diff
@@ -46,7 +48,30 @@ func (c *Config) RegisterHandlers(router fiber.Router) error {
...
+	hasOIDC := c.OIDC != nil
+	hasBasic := c.Basic != nil
+
+	// If we have API tokens, add a pre-check middleware
+	if hasAPITokens {
+		router.Use(func(ctx *fiber.Ctx) error {
+			// Check for Bearer token
+			authHeader := ctx.Get("Authorization")
+			if strings.HasPrefix(authHeader, "Bearer ") {
+				token := strings.TrimSpace(strings.TrimPrefix(authHeader, "Bearer "))
```
PythonGermany:
```suggestion
			if after, ok := strings.CutPrefix(authHeader, "Bearer "); ok {
				token := strings.TrimSpace(after)
```
Suggested by gopls check

security/config.go
```diff
@@ -74,27 +106,46 @@ func (c *Config) ApplySecurityMiddleware(router fiber.Router) error {
...
 }
 
 // IsAuthenticated checks whether the user is authenticated
 // If the Config does not warrant authentication, it will always return true.
 func (c *Config) IsAuthenticated(ctx *fiber.Ctx) bool {
+	// Check for API token authentication first
+	if c.API != nil && len(c.API.Tokens) > 0 {
+		authHeader := ctx.Get("Authorization")
+		if strings.HasPrefix(authHeader, "Bearer ") {
+			token := strings.TrimSpace(strings.TrimPrefix(authHeader, "Bearer "))
```
PythonGermany:
```suggestion
		if after, ok := strings.CutPrefix(authHeader, "Bearer "); ok {
			token := strings.TrimSpace(after)
```
Suggested by gopls check

security/config.go
```diff
@@ -22,13 +23,14 @@ const (
...
 	OIDC  *OIDCConfig  `yaml:"oidc,omitempty"`
+	API   *APIConfig   `yaml:"api,omitempty"`
 
 	gate *g8.Gate
 }
 
 // ValidateAndSetDefaults returns whether the security configuration is valid or not and sets default values.
 func (c *Config) ValidateAndSetDefaults() bool {
-	return (c.Basic != nil && c.Basic.isValid()) || (c.OIDC != nil && c.OIDC.ValidateAndSetDefaults())
+	return (c.Basic != nil && c.Basic.isValid()) || (c.OIDC != nil && c.OIDC.ValidateAndSetDefaults()) || (c.API != nil && c.API.Validate() == nil && len(c.API.Tokens) > 0)
```
PythonGermany, line 33: This validation method does not work as intendend:

If more than one auth method is configured and the config for at least one of them is valid the validation check result for all other config sections is disregarded.

In such cases Gatus will still start without any indication of an invalid config even though it should show an error and quit immediately.

security/api-tokens.go
```diff
@@ -0,0 +1,40 @@
+package security
+
+import "errors"
+
+var (
+	ErrAPITokensEmpty = errors.New("security.api.tokens must not contain empty tokens")
```
PythonGermany, line 9: When the security config validation fails only a generic message is displayed (should definitely be improved upon, since with the growing security config it becomes harder to understand what went wrong during validation).

Because of  this the error here is currently only used in the unit test.

security/config.go
PythonGermany, line 33: Not specifically related to this PR: I think this method should also validate that only one of Basic or OIDC is configured since basic auth is not even evaluated when OIDC auth is configured.

security/config.go
```diff
@@ -46,7 +48,30 @@ func (c *Config) RegisterHandlers(router fiber.Router) error {
...
+			if strings.HasPrefix(authHeader, "Bearer ") {
+				token := strings.TrimSpace(strings.TrimPrefix(authHeader, "Bearer "))
+				if c.API.IsValid(token) {
+					// Valid API token - mark as authenticated and skip other auth middleware
+					ctx.Locals("api_token_authenticated", true)
+					return ctx.Next()
+				}
+			}
+			// No valid Bearer token, continue to next middleware
+			return ctx.Next()
```
PythonGermany, line 70: This line here essentially leads to the configured API tokens doing nothing if no other auth method is configured.

I know the PR description states:
> optional API tokens used to access the API when basic or OIDC auth is enabled

So I am not entirely sure if it should also be possible to use API tokens independently of basic or OIDC auth. 

Certain I am about one thing though: From my perspective I think it is not a good idea to implement an authentication method like API tokens in a way where they are simply skipped if no other auth method is configured.
This opens up the possibility for bad situations where people configuring API tokens as their only auth method will learn one way or another that the auth config to protect the endpoints does nothing. 

Something like this would already improve it a lot (will break the UI though if API tokens are the only auth method configured since the API requests from the frontend will all  receive a 401 response):
```suggestion
			if hasBasic || hasOIDC {
				return ctx.Next()
			}
			return ctx.Status(401).SendString("Unauthorized")
```
Or as an alternative, if API tokens really are only wanted in combination with either basic or OIDC auth, enforce that API tokens cannot be configured if no other auth method is configured.

It's late in the day for me. I hope all of my points make sense and are understandable.

PythonGermany (changes requested):
@TwiN I went over the changes and found quite a few things. Haven't gotten around to test it locally with OIDC. I think I'll wait for this wave of comments to be resolved and go over it again also testing the OIDC compatibility of the changes (haven't used OIDC with gatus so far, so will have to look into that)

PythonGermany:
Hi @TwiN, may I ask why some API routers were not protected before even with basic auth or OIDC configured? Shouldn't they also be protected by an explicitly configured API token?
https://github.com/TwiN/gatus/blob/b65b4d83e17fad374911e4f0d42e1739146b4f7c/api/api.go#L75-L87

TwiN:
@PythonGermany which one specifically are you asking about? Some of them aren't protected because they're used as badges in public domain, others are needed for the frontend, etc.

security/config.go
aaldebs99: As I understand it, If there's no auth configured all API calls work without auth by default. If an API token is configured without any auth, the UI still works and all APIs still respond without a token. I believe that's just default behavior of Gatus, even outside of this PR

Example with API token and no auth configured:

```
curl -v http://localhost:8080/api/v1/endpoints/statuses
* Host localhost:8080 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8080...
* Immediate connect fail for ::1: Cannot assign requested address
*   Trying 127.0.0.1:8080...
* Established connection to localhost (127.0.0.1 port 8080) from 127.0.0.1 port 35886
* using HTTP/1.x
> GET /api/v1/endpoints/statuses HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.18.0
> Accept: */*
>
* Request completely sent off
< HTTP/1.1 200 OK
< Date: Tue, 03 Feb 2026 23:09:23 GMT
< Content-Type: application/json
< Content-Length: 6304
<
[{"name":"check-domain-expiration","key"[...]
```
`asdf` is actually valid in this case
```
curl -v http://localhost:8080/api/v1/endpoints/statuses -H "Authorization: Bearer asdf"
* Host localhost:8080 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8080...
* Immediate connect fail for ::1: Cannot assign requested address
*   Trying 127.0.0.1:8080...
* Established connection to localhost (127.0.0.1 port 8080) from 127.0.0.1 port 35562
* using HTTP/1.x
> GET /api/v1/endpoints/statuses HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.18.0
> Accept: */*
> Authorization: Bearer asdf
>
* Request completely sent off
< HTTP/1.1 200 OK
< Date: Tue, 03 Feb 2026 23:10:06 GMT
< Content-Type: application/json
< Content-Length: 6480
<
[{"name":"check-domain-expiration","key"[...]
```

aaldebs99:
I see the concern you have now that I think about it. In my opinion these API tokens are NOT a a source of auth like basic/OIDC auth. They're supposed to be supplemental to those for API access. 

The concern is that a user might expect the `api.tokens.[]` to enable auth for at least the APIs when it does not.

With that said it might be better to put the API token key UNDER those auths in the yaml (`security.oidc/basic.api.tokens.[]`)sort of like:
```
security:
  oidc:
    issuer-url: https://auth.example.com/application/o/gatus/
    client-id: "${OIDC_CLIENT_ID}"
    client-secret: "${OIDC_CLIENT_SECRET}"
    redirect-url: https://status.example.com/authorization-code/callback
    scopes: [openid]
    api:
      tokens:
        - ${API_TOKEN}
```

I'm not 100% sold on that approach thought so I'm open to suggestions

PythonGermany:
> which one specifically are you asking about? Some of them aren't protected because they're used as badges in public domain, others are needed for the frontend, etc.

@TwiN I see, so the goal was that "basic" endpoints providing things like current health and uptime for a single endpoint were never protected but endpoints providing more detailed insights can be protected through configuration.

I'm just asking because when I looked through the changes proposed in #1482 and also this PR I noticed that it might make sense to also be able to also optionally configure to protect the currently still unprotected endpoints. In a new PR of course in case that is something that's relevant/you are not against it.

Is there documentation somewhere informing about which endpoints are protected and which are not when an authentication method is configured? I was not able to find it in the readme.

TwiN:
@PythonGermany It's not documented, but basically everything that isn't protected isn't what I would qualify as confidential information. Badges with up/down/response times is nothing to worry about, nor is the config endpoint that exposes things like whether authentication is needed to access the status page.

The only gray area I'd say is the metrics endpoint, but metrics aren't enabled by default, nor is there any crazy information exposed there. For now, I'd say don't change the current behavior; if somebody ever has any problem with it, we'll discuss it then, but for now, this approach hasn't been a problem

security/api-tokens.go
aaldebs99: Done in 74861f1bf89cb51fcb81f70fe7c962f2f823b43c

security/config.go
aaldebs99: Done in 74861f1bf89cb51fcb81f70fe7c962f2f823b43c

aaldebs99:
So there's some design decisions to be made here I believe:
1. Validation OR logic allows invalid configs to pass silently. Should one auth type (Basic vs OIDC) take priority or should the config allow only one type of auth. 
  a. If both are allowed somehow, should having one invalid type error out during startup? This one is mostly unrelated to this PR.
3. Generic error messages don't indicate which field failed. Should those errors be more specific? 
4. API tokens should require Basic or OIDC auth. How should we handle that? We can leave the yaml key as is an throw an error if no auth is provided. Alternatively we can put the API tokens under those two auth
5. Should the metrics endpoint remain open or should that require auth too? Also mostly unrelated to this PR

Aside from those, I do think I need to separate the API tokens check from the other auths at the very least

PythonGermany:
> Validation OR logic allows invalid configs to pass silently.

Since this was a previously existing issue I created a separate PR to fix it so this one can build on it once merged: #1531

> Generic error messages don't indicate which field failed. Should those errors be more specific?

I think in for the scope of this PR it would be fine to leave it as is just removing the unused error message. The improvement for the error messages can then also be added for all existing auth types in a separate issue.

> Should the metrics endpoint remain open or should that require auth too?

Currently the metrics endpoint is never protected. #1483 will add the option to protect it via the already available auth methods.
The only question would be if API tokens should also apply to the metrics which are served under `/metrics` (not `/api/metrics`) which might be confusing since the token config includes `API` in its name.

aaldebs99:
> Currently the metrics endpoint is never protected. #1483 will add the option to protect it via the already available auth methods. The only question would be if API tokens should also apply to the metrics which are served under `/metrics` (not `/api/metrics`) which might be confusing since the token config includes `API` in its name.

Yea the naming might be confusing but I personally would want the metrics to be accessible by the API tokens. I use OIDC only and I would have no reliable way to access those metrics without the API tokens.
That's the whole reason I put this PR together lol. I wanted to set up Homepage widget with Gatus and couldn't with OIDC enabled.

PythonGermany:
> I use OIDC only and I would have no reliable way to access those metrics without the API tokens.

I understand and agree that there should be a way to access the metrics reliably even with OIDC configured.

Another suggestion would be to configure access to the metrics via a separate config token, e.g.
```yaml
security:
  metrics:
    protected: true # Will enable protection via basic or OIDC auth (since the default if false)
    tokens:
      - "supersecrettoken"
```
And if there's no need for separate tokens a user could use yaml anchors as mentioned in the README already.

Not sure if this idea introduces too much config complexity though it would certainly resolve the naming confusion issue.

aaldebs99:
> > I use OIDC only and I would have no reliable way to access those metrics without the API tokens.
> 
> I understand and agree that there should be a way to access the metrics reliably even with OIDC configured.
> 
> Another suggestion would be to configure access to the metrics via a separate config token, e.g.
> 
> ```yaml
> security:
>   metrics:
>     protected: true # Will enable protection via basic or OIDC auth (since the default if false)
>     tokens:
>       - "supersecrettoken"
> ```
> 
> And if there's no need for separate tokens a user could use yaml anchors as mentioned in the README already.
> 
> Not sure if this idea introduces too much config complexity though it would certainly resolve the naming confusion issue.

Yup that approach would be fine too in my opinion

PythonGermany:
So a complete config would look something like this (currently basic auth would be ignored since OIDC is configured):
```yaml
security:
  metrics:
    protected: true # Will enable protection via basic or OIDC auth (since the default if false, optional/implicitly true if tokens are configured to keep config small)
    tokens:
      - "supersecrettoken"
  oidc:
    issuer-url: https://auth.example.com/application/o/gatus/
    client-id: "${OIDC_CLIENT_ID}"
    client-secret: "${OIDC_CLIENT_SECRET}"
    redirect-url: https://status.example.com/authorization-code/callback
    scopes: [openid]
    api-tokens:
      - ${API_TOKEN}
  basic:
    username: "john.doe"
    password-bcrypt-base64: "JDJhJDEwJHRiMnRFakxWazZLdXBzRERQazB1TE8vckRLY05Yb1hSdnoxWU0yQ1FaYXZRSW1McmladDYu"
    api-tokens:
      - ${API_TOKEN}
```

I changed `security.oidc.api.tokens` to `security.oidc.api-tokens` in an attempt to make it shorter. Not sure if it is worth it. Depends entirely on whether there is any scenario where the api config for each auth method would grow.

@TwiN What do you think about the suggested config structure?

conradkun:
Interested in this too. Is this PR stale? I wouldn't mind giving it a go from a more recent gatus version to avoid merge conflicts.

PR #1360, opened 2025-10-22
feat(endpoints): add domain-expiration-only-check

### LINKED ISSUE #1159: Run Domain Expiration checks on the root domain

`DOMAIN_EXPIRATION` should automatically pick up the root domain in a HTTP(S) url and check the expiry for that. Currently, it does not and fails with a negative date:

`[DOMAIN_EXPIRATION] (-9223372036854) > 720h (2592000000)`

because the ExpiryDate resolves to an invalid date: `0001-01-01 00:00:00 +0000 UTC`.

I want a single status check that checks for 3 conditions: domain expiry, cert expiry, and website status. However, the website is often published on a subdomain only such as `www`.

```
- name: gully-tours
  interval: 6h
  client:
    timeout: 30s
  conditions:
    - "[STATUS] == 200"
    - "[DOMAIN_EXPIRATION] > 720h"
    - "[CERTIFICATE_EXPIRATION] > 240h"
    url: https://www.gully.tours/
```

In such cases, the `DOMAIN_EXPIRATION` check should be smart enough to use `gully.tours` as the domain name instead of `www.gully.tours`. 

_No response_

### Additional information

I checked `ExpirationDate` in the `twin/whois` library for both, and perhaps this could be resolved at the `whois` layer using the public suffix list?

- `gully.tours` - 2026-03-26 12:11:13 +0000 UTC
- `www.gully.tours` - 0001-01-01 00:00:00 +0000 UTC

qianlongzt:
## Summary

1. fix https://github.com/TwiN/gatus/issues/1159

image
image

2. add missing endpoint type test for grpc https://github.com/TwiN/gatus/pull/1376

## Checklist


qianlongzt:
@TwiN Is there anything I can do to help merge PR? Is there no progress?

config/endpoint/endpoint_test.go
```diff
@@ -390,6 +419,20 @@ func TestEndpoint_Type(t *testing.T) {
...
+			endpoint := Endpoint{
+				URL:       tt.args.URL,
+				DNSConfig: tt.args.DNS,
+			}
+			if got := endpoint.Type(); got != tt.want {
+				b.Errorf("Endpoint.Type() = %v, want %v", got, tt.want)
+			}
+		}
+	}
+}
```
PythonGermany: Should probably be moved to a new file just for benchmark tests `endpoint_bench_test.go` to be consistent with the other benchmark tests I've seen in this project so far.

qianlongzt: Makes sense. I'll move the benchmark to a separate `endpoint_bench_test.go` file and reuse the existing test data.

qianlongzt:
@PythonGermany Your PR https://github.com/qianlongzt/gatus/pull/1 looks good to me, but what should I do next?

1. Merge your PR first and then wait for @TwiN to review?
2. Or should I not merge it, and instead have you submit your PR to @TwiN after my PR is merged?

PythonGermany:
@qianlongzt Just merge my changes into your branch.

config/endpoint/endpoint.go
```diff
@@ -542,9 +545,15 @@ func (e *Endpoint) call(result *Result) {
...
 		if e.needsToReadBody() {
-			result.Body = []byte(fmt.Sprintf("{\"status\":\"%s\"}", status))
+			result.Body = fmt.Appendf(nil, "{\"status\":\"%s\"}", status)
 		}
-	} else {
+	case TypeDomain:
+		// domain expiration checked before call `call`
+		return
+	case TypeHTTP:
+		request = e.buildHTTPRequest()
```
PythonGermany, line 551: I noticed my previous refactoring introduces one unintended (and maybe unwanted change). The startTime is now set before starting to build the http request.

To restore the original behavior we should update the startTime after building the request:
```suggestion
		request = e.buildHTTPRequest()
		startTime = time.Now()
```

qianlongzt: I don’t think it’s necessary, since buildHTTPRequest is already very fast (around 5,500 ns).

image

```
goos: windows
goarch: amd64
pkg: github.com/TwiN/gatus/v5/config/endpoint
cpu: AMD Ryzen 7 4800H with Radeon Graphics         
BenchmarkEndpoint_buildHTTPRequestWithHostHeader-16    	  231338	      5460 ns/op	    4896 B/op	      36 allocs/op
```

PythonGermany: I understand, it might take significantly longer tough if slower/older hardware is used in combination with a large configured endpoint body with many placeholders that need to be replaced when building the request. I realize this is a very unlikely scenario and the resulting displayed difference will be 1ms at most, so you are probably right that it's not necessary.

config/endpoint/endpoint.go
```diff
@@ -161,29 +163,36 @@ func (e *Endpoint) IsEnabled() bool {
...
+	case "tcp":
 		return TypeTCP
-	case strings.HasPrefix(e.URL, "sctp://"):
+	case "sctp":
 		return TypeSCTP
-	case strings.HasPrefix(e.URL, "udp://"):
+	case "udp":
 		return TypeUDP
-	case strings.HasPrefix(e.URL, "icmp://"):
+	case "icmp":
```
TwiN: I understand and appreciate your intention to simplify the code, but personally, having the ability to do a global search for `<protocol/endpoint-type>://` (e.g. `icmp://`) makes navigating the code much easier (reasoning here being that if you search for just `http` instead of `http://`, you're going to get hundreds of results as opposed to exactly what you're looking for).

Could you revert the changes here and update the new `domain://` prefix accordingly?

config/endpoint/endpoint_test.go
```diff
@@ -280,104 +301,124 @@ func TestEndpoint_IsEnabled(t *testing.T) {
 	}
 }
 
+type testEndpoint_typeArgs struct {
+	URL string
+	DNS *dns.Config
+	SSH *ssh.Config
+}
+
+var testEndpoint_typeData = []struct {
```
TwiN: The whole `Test<something>_<something>` notation is more intended to be for the struct name followed by the method name (and occasionally further details).

Is there any reason why you moved those outside of the function? They're only used in the function, you could just name them `typeArgs` and `typeData` within the `TestEndpoint_Type` test function itself

TwiN (changes requested):
Small nitpicks

config/endpoint/endpoint_test.go
qianlongzt: I just want `BenchmarkEndpoint_Type` to reuse the same test data as `TestEndpoint_Type`.

This benchmark is for both the old and new `Type` implementations, and I would like to keep it in the tests.

config/endpoint/endpoint.go
qianlongzt: Should we add quotation marks to the search string, for example `"icmp"` and `"http"`?

Image

Image

Image

qianlongzt:
@TwiN any update?

config/endpoint/endpoint_test.go
qianlongzt: removed

config/endpoint/endpoint.go
qianlongzt: reverted

qianlongzt:
@TwiN, could you please review it again? I restored the Endpoint.Type() to its original implementation.

Issue #1319, opened 2025-10-10
Allow html tags or markdown in announcements


AlexCK-STFC:
### Describe the feature request

Announcement messages are[ escaped right now](https://github.com/TwiN/gatus/blob/master/web/app/src/components/AnnouncementBanner.vue#L93). It would be useful to allow some form of formatting in them.

TwiN:
I wouldn't be against allowing Markdown, provided the frontend would do the rendering from markdown to HTML.

AlexCK-STFC:
That would be good.

Another thing that's missing right now is being able to add newlines, too.

TwiN:
Markdown rendering should make newlines work out of the box.

TwiN:
Support for markdown in announcements has been added

PR #1371, opened 2025-10-28
feat(announcements): add markdown support

Sworyz:
## Summary

Add markdown support to announcements to allow link and more format to be written here. The idea was from #1319 .

---
One problem that seems to stay on multi-line is this :
```css
<div data-v-06677c1a="" class="absolute left-3 w-0.5 bg-gray-300 dark:bg-gray-600 pointer-events-none" style="top: 1.5rem; height: 9rem;"></div>
```
Putting height: 11rem fix the problem as we add 2 more lines.

---

## Checklist

It was tested with a build.
image


Sworyz:
Example of announcements :

```yaml
announcements:
  - timestamp: 2025-08-15T14:00:00Z
    type: outage
    message: "Scheduled maintenance on database servers from 14:00 to 16:00 UTC"
  - timestamp: 2025-08-15T16:15:00Z
    type: operational  
    message: "Database maintenance completed successfully. All systems operational. [google link](https://google.fr)"
  - timestamp: 2025-08-15T12:00:00Z
    type: information
    message: "New *monitoring* dashboard features <br />will be deployed <br />next week"
```

Sworyz:
Multiline timeline resolved by #1373 
image

TwiN:
@Sworyz hah, I'm surprised somebody noticed that I fixed the multiline timeline issue. 

Alright, I'm sorry for being difficult, but I'm second guessing whether it's worth it to add an entire dependency for markdown, when realistically speaking, users will likely only care about adding support for links, bold text, italic text, and _maybe_ headers.

What do you think? Would it make sense to just create some code in the frontend to support those, as opposed to importing an entire markdown dependency?

Sworyz:
Hi @TwiN,

Yes, the multiline issue bothered me and I was glad to seen it resolved so I really wanted to mention it!

No worries about you being difficult, it’s a good thing and I really appreciate it. 👍 
I also understand wanting to avoid unnecessary complexity and dependencies. 

I’m going talk about our use case for this PR : we scrape incident and issue data from Jira using a sidecar container and write it to YAML. The markdown support is really valuable for formatting this data into clear, readable announcements. Without it, we may have a harder time presenting the information effectively.

Marked and Dompurify seems to be very widely used without too much dependencies (one for dompurify only). Hence why I’ve gone with them to make this working.

I also tend to agree that many users might only need basic formatting like links, bold text, and headers. However, I'm a little concerned that limiting the markdown support too much could restrict users to a single announcement structure impacting the presentation of the information to only one or two structures that may not fit everyone use case.

I was also thinking before adding markdown this simple thing : just adding a link to the announcement yaml part and making the announcement bubble clickable, but it wasn’t enough for our use case.

TwiN:
@Sworyz Alright, what you're saying makes sense to me. I'll review the code when I get a minute

TwiN:
I just added support for archiving announcements (past announcements) in #1382

@Sworyz Could you please resolve the conflicts & regenerate the static assets with the new changes from the master branch? I think you'll also need to update the new `PastAnnouncements.vue` component to also support markdown.

image

Sorry again for the trouble!

Sworyz:
Hi @TwiN,

I've done the resolution of the merge conflict; a quick fix for a markdown link rendering issue and the past announcement that can now use md. I've also rebuilt the static files like you wanted!

image
image

web/app/src/components/AnnouncementBanner.vue
```diff
@@ -232,6 +237,55 @@ const getTypeClasses = (type) => {
...
+    return ''
+  }
+
+  return String(value)
+    .replace(/&/g, '&amp;')
+    .replace(/</g, '&lt;')
+    .replace(/>/g, '&gt;')
+    .replace(/"/g, '&quot;')
+    .replace(/'/g, '&#39;')
+}
```
TwiN, line 251: There's probably a more secure way to do this and I'm sure some security scan will complain, but since the content of the announcement is controlled by the configuration file, XSS isn't a concern.

web/app/src/components/AnnouncementBanner.vue
```diff
@@ -232,6 +237,55 @@ const getTypeClasses = (type) => {
...
+
+  const href = tokenObject ? tokenObject.href : tokenOrHref
+  const resolvedTitle = tokenObject ? tokenObject.title : title
+  const resolvedText = tokenObject ? tokenObject.text : text
+
+  const url = escapeHtml(href || '')
+  const titleAttribute = resolvedTitle ? ` title="${escapeHtml(resolvedTitle)}"` : ''
+  const linkText = resolvedText || ''
+
+  return `<a href="${url}" target="_blank" rel="noopener noreferrer"${titleAttribute}>${linkText}</a>`
```
TwiN, line 268: Is there a reason why you don't specify tailwind classes here instead of using the `<style>` tag in the same file?

Also, now that there's two files where announcement messages are rendered with markdown, this duplicated code should probably be extracted to a shared utility file. I'll merge this and take care of it myself in a little bit

TwiN (approved):

TwiN:

TwiN:
Made a few improvements in 379ec2983dd0f7ea69fa3da65b30f66cc828075d

amai2012:
The feature basically is nice to have.
But the `Past Announcements` lists should be sparse, not feature an entry for each day. Otherwise the list will grow too big.

Can it be hidden?

TwiN:
@amai2012 This comment is under the wrong PR, but to answer your question, if there are no archived announcements, the Past Announcement section is hidden. This is explained in the documentation.

PR #1372, opened 2025-10-28
feat(docker): add multistage docker to prevent the needs of building static if change happens

Sworyz:
## Summary

The dockerfile is now multistage to prevent using the makefile in a full docker build environment.

## Checklist

Working with the modifications I made for markdown support without any static in app folder

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the frontend assets
-FROM node:20-alpine AS frontend
+FROM node:24-alpine AS frontend
 WORKDIR /app/web/app
 COPY web/app/package*.json ./
 RUN npm install


TwiN:
Why is this needed?

I understand your intentions, but my trust in the npm package ecosystem is very low

Sworyz:
Hello,

I completly agree with you about the trust in npm packages.
However having this allow to build fork or changes on the frontend without using the make and rebuilding with the makefile or npm commands : everything is inside the docker.

It even allows for a docker compose file with build without rebuilding the frontend externally.

To add more, if you are afraid of vulnerabilities, a trivy scan can for example be run against the image or files.

Regards,

TwiN:
I understand and appreciate your intentions, but right now, this isn't something Gatus _needs_, and I would rather not focus my limited time on things that aren't necessities and that would require further analysis (e.g. what to do about discrepancies Docker static files vs pushed static files, what impact this would have on container image publish time, what impact this would have from a security perspective, etc.)

Unfortunately, Gatus is just a side project for me, and while I would love to do this full time, I can't, and thus I have no choice but to prioritize what I spend my time on.

Next time, please create an issue so we can discuss it first before you spend time implementing something that I may have to turn down. 

Regardless, I do really appreciate your contribution, even if I'm going to close this PR, and I hope you understand.

PR #1380, opened 2025-11-07
feat(settings): add keyboard shortcut to refresh data on 'R' key press

nillpoe:
## Summary

Added keyboard shortcut (R) to refresh data.

## Checklist


TwiN:
I appreciate your eagerness to contribute, but this isn't something I'm interested in adding at the moment.

Next time, it may be a better idea to first create an issue so we can discuss it prior to jumping to the implementation.

PR #1390, opened 2025-11-16
feat(client): Add ssh private-key support

### LINKED ISSUE #1257: No option for SSH key

When testing a server via the SSH endpoint, you can specify a username and password for an actual connection. You can also leave them empty to just test whether there is a daemon listening. However, you cannot seem to use SSH-keys (in the case of passwordless authentication). Could you please add this option?

In our case, we hardened our servers by (amongst others) disabling password authentication. This makes for a more secure setup. I still want to test the possibility of actually logging in and running a command.

Just started

### Additional information

_No response_

mufeedali:
Fixes #1257

## Summary
Adds support for using an unencrypted SSH key for SSH endpoint testing. Closes #1257 . I don't know if I should be adding a separate test for private-key, please let me know if I should :)

## Checklist

### DIFF
--- a/config/endpoint/endpoint_test.go
+++ b/config/endpoint/endpoint_test.go
@@ -511,24 +511,38 @@ func TestEndpoint_ValidateAndSetDefaultsWithSSH(t *testing.T) {
 		name        string
 		username    string
 		password    string
+		privateKey  string
 		expectedErr error
 	}{
 		{
-			name:        "fail when has no user",
+			name:        "fail when has no user but has password",
 			username:    "",
 			password:    "password",
 			expectedErr: ssh.ErrEndpointWithoutSSHUsername,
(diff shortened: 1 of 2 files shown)


TwiN:
Excellent work! I'd appreciate a test for the private key, as you've suggested, though, and then we're good to go! No need to make it super complex.

mufeedali:
I looked into what we could be testing here. I've added a few cases to the existing config testing. But I can't think of a way to implement proper integration testing. Password-only could maybe be verified using https://test.rebex.net/ but can't think of a decent way to test private key auth.

TwiN (approved):

mufeedali:
@TwiN I think the tests failed due to GitHub facing issues. Could you try re-running the tests?

PR #1405, opened 2025-11-28
fix(ui): Show correct oldest result timestamp

### LINKED ISSUE #1404: Incorrect oldest result time in home view
### Describe the bug

Is is possible for the oldest result timestamp of the EndpointCard component to be incorrect as it is currently used in the home view.

### What do you see?

For an endpoint interval of 1m the timestamp will shows  "2 hours ago" even tough the last visible timestamp is only from ~50min ago.

Image

Image

Timestamp displaying "50 minutes ago" like in the EndpointDetails view:

Image

1. Create endpoint with short interval
2. Wait for result list to fill up and then a little longer

### Version

v5.33.0

### Additional information

_No response_

PythonGermany:
## Summary

Closes #1404

## Checklist

### DIFF
--- a/web/app/src/views/EndpointDetails.vue
+++ b/web/app/src/views/EndpointDetails.vue
@@ -89,13 +89,13 @@
                 <EndpointCard 
                   v-if="endpointStatus"
                   :endpoint="endpointStatus"
-                  :maxResults="50"
+                  :maxResults="resultPageSize"
                   :showAverageResponseTime="showAverageResponseTime"
                   @showTooltip="showTooltip"
                   class="border-0 shadow-none bg-transparent p-0"
                 />
                 <div v-if="endpointStatus && endpointStatus.key" class="pt-4 border-t">
-                  <Pagination @page="changePage" :numberOfResultsPerPage="50" :currentPageProp="currentPage" />
+                  <Pagination @page="changePage" :numberOfResultsPerPage="resultPageSize" :currentPageProp="currentPage" />
                 </div>
               </div>
             </CardContent>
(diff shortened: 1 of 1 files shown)


PythonGermany:
Are there any relevant UI tests that I can extend with a check to verify that the timestamps match the displayed result range?

TwiN:
@PythonGermany I'm not a fan of frontend tests to be honest, so feel free to skip testing the frontend changes. I do most of my frontend tests manually, though I'm having an issue replicating your issue on my end.

PythonGermany:
The issue can also be observed in the [demo](https://status.twin.sh/). In the EndpointCard component of the blog-external endpoint of the home view it says something like "2 hours ago" for oldest result and "31 seconds ago" for latest result. However if you then look at the timestamp of the first and last displayed result in the green status bar above the oldest displayed result is about ~50 minutes ago. This issue does not occur on the [details page ](https://status.twin.sh/endpoints/core_blog-external) of the endpoint. There the oldest result timestamp is displayed correctly.

web/app/src/views/Home.vue
```diff
@@ -433,7 +434,7 @@ const fetchData = async () => {
   }
   try {
     // Fetch endpoints
-    const endpointResponse = await fetch(`${SERVER_URL}/api/v1/endpoints/statuses?page=1&pageSize=100`, {
+    const endpointResponse = await fetch(`${SERVER_URL}/api/v1/endpoints/statuses?page=1&pageSize=${resultPageSize}`, {
```
TwiN, line 437: I think this may have been the issue honestly. Good catch; looks like some bugs fell through the cracks when I updated the UI

PythonGermany: Yes, this issue and the one in the EndpointCard component combined caused the UI bug. Each on their own wouldn't have caused any UI issues.

TwiN:
Thank you for the contribution @PythonGermany !

PR #1407, opened 2025-11-28
fix(ui): Show correct avg response time for N/A value

PythonGermany:
## Summary
Fixes issue where value of average response time field is "N/Ams" if currently selected result page has no result with duration value:
grafik

## Checklist

### DIFF
--- a/web/app/src/views/EndpointDetails.vue
+++ b/web/app/src/views/EndpointDetails.vue
@@ -89,13 +89,13 @@
                 <EndpointCard 
                   v-if="endpointStatus"
                   :endpoint="endpointStatus"
-                  :maxResults="50"
+                  :maxResults="resultPageSize"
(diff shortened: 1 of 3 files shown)


web/app/src/views/EndpointDetails.vue
```diff
@@ -274,17 +274,17 @@ const pageResponseTimeRange = computed(() => {
...
+    if (duration) {
+      min = Math.min(min, duration)
+      max = Math.max(max, duration)
       hasData = true
     }
   }
   
   if (!hasData) return 'N/A'
-  const minMs = Math.round(min)
-  const maxMs = Math.round(max)
```
PythonGermany, line 288: I made this change to make it consistent to with pageAverageResponseTime implementation property of the same component.

TwiN:
@PythonGermany Thank you for the contribution!

PR #1408, opened 2025-11-29
ci: Add build stage to generate static ui assets

PythonGermany:
## Summary

Add build stage to dockerfile to automatically generate static assets instead of having them in version history and having to update them manually.

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,6 +20,8 @@ jobs:
         with:
           go-version: 1.24.4
       - uses: actions/checkout@v5
+      - name: Build frontend
+        run: make frontend-install-dependencies frontend-build
       - name: Build binary to make sure it works
         run: go build
       - name: Test


Dockerfile
```diff
@@ -1,9 +1,18 @@
+# Build the static web assets
+FROM node:alpine AS web-builder
```
PythonGermany, line 2: Would pinning the node version to a specific one be appropriate?

PythonGermany:
A few questions I have about the topic:
- Is there a reason as for why the static generated files were in version history so far?
- Doesn't this cause unnecessary merge conflicts and other unforeseen issues?
- Are there advantages I can't think of?

TwiN:
It occasionally causes merge conflicts, but because the static assets are minified, this only really happens when the PR is outdated. 

Personally, I would rather have a GitHub workflow that would regenerate the static assets when I post a specific comment, something like `/generate-static-assets`.

The big pro here is that whatever is in the repository is exactly what I expect, and it can be historically verified and audited. 

With your approach, if there was a compromised npm library with malicious code, because the frontend is built when the container is built, we'd never know, because the assets generated are not committed.

It's more work for me, but before I do releases and I know the static assets have changed due to a change in package.json, I do my due diligence and look at what changed, manually.

Your suggested change wouldn't necessarily prevent me from doing the same thing, but it assumes that the version hasn't been poisoned, which sure, is far fetched with the checksums in place that should prevent them and all, but it does make it easier to become complacent about it.

PythonGermany:
Makes sense. When the static assets are not in the repo, they would only be in the container registries.

TwiN:
Btw, I still think a GitHub workflow that would regenerate the static assets when I post a specific comment in a PR and commit the generated assets, something like `/generate-static-assets`, would have great value, if that's something you're interested in building

PR #1418, opened 2025-12-06
build(docker): Non tidy go mods fail image build

PythonGermany:
## Summary

The Dockerfile to build gatus runs `go mod tidy` which might modify the go.mod and go.sum files inside the build stage image without failing if changes to those files are required to match the module source code.

Not failing the build in a situation like this might be a decision made on purpose for a reason that I do not know about or don't understand but as far as my current understanding goes it does not make sense to allow a build to succeed when the files are changed implicitly inside the build stage.

What are your thoughts on this?

If there speaks nothing against it I would suggest adding the diff flag which will make the build fail in those situations as explained [here](https://go.dev/ref/mod#go-mod-tidy):
> The -diff flag causes go mod tidy not to modify go.mod or go.sum but instead print the necessary changes as a unified diff. It exits with a non-zero code if the diff is not empty.

## Checklist

  - Locally modified go.mod to not match module source code and tried to build image -> Build failed with diff output

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM golang:alpine AS builder
 RUN apk --update add ca-certificates
 WORKDIR /app
 COPY . ./
-RUN go mod tidy
+RUN go mod tidy -diff
 RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o gatus .
 
 # Run Tests inside docker image if you don't have a configured go environment


TwiN (approved):

TwiN:
@PythonGermany Just want to say that I appreciate all of these little improvements you're making around Gatus.

I got back to work from paternity leave a few weeks ago and haven't been able to get to this polishing part during my leave, as I was focusing on big picture changes while being a little neglectful of the edges that needed polishing.

Issue #1403, opened 2025-11-27
Allow regex search in UI - search by regex by default


hrvoj3e:
### Describe the feature request

I propose a feature for UI - search by regex by default

This would not break the current behaviour but it would improve search for advanced users.

https://github.com/TwiN/gatus/blob/9f960fdd27371bf1abfe77d387a5d4743101ee9c/web/app/src/views/Home.vue#L234-L236

Simple change would be required:
```js
  if (searchQuery.value) {
    const query = searchQuery.value.toLowerCase()
    const regexp = new RegExp(query, 'g');
    filtered = filtered.filter(endpoint => 
      null !== (endpoint.name.toLowerCase().match(regexp) ?? (endpoint.group && endpoint.group.toLowerCase().match(regexp)) ?? null)
    )
  }
```

### Why do you personally want this feature to be implemented?

I want to be able to search more complex names and endpoint across groups when just plain grouping does not provide the needed granularity.

PR #1420, opened 2025-12-07
feat(ui): Add home view query parameter functionality

### LINKED ISSUE #1263: Allow passing dashboard filters as query parameters

The state of the input fields on the dashboard (search, refresh interval, sort, filter, ...) is currently stored in the browsers local storage. This ensures that changed filter settings will persist on page reload.

It should be possible to (also) pass the filter values through query parameters. That could either replace the current local-storage mechanism completely, or the values stored in the local storage could serve as default fallback values if no query parameter is supplied.

Passing the filter values as query parameters allows for better link sharing and bookmarking of specific settings.
Say i have a lot of endpoints and i want a bookmark to the dashboard that shows all failed backups, i can imagine a URL like this:
```
https://gatus.example.com?q=backup&filter=failing
```

Or i want someone to check out the unstable services, sorted by group, i'd quickly share this link:
```
https://gatus.example.com?filter=unstable&sort=group
```

 

_No response_

### Additional information

_No response_

PythonGermany:
## Summary

It might be a useful useful feature to set the search, sort and filter options as query parameters so that links with pinned values for those options are shareable across users.

This could enable valuable use cases and could be further enhanced with #1403

I propose the functionality should work like this for sort and group settings:
- Behavior on page load:
  - [x] Local storage: Do not update
  - [x] Query parameter: If exists set app config value to query parameter value if it does not set it from app config (local storage -> default server side config -> hardcoded config)
- Behavior for user input via UI:
  - [x] Local storage: Set to user input value
  - [x] Query parameter: Set to user input value
- Query parameter changes without reload (Forward/backward navigation)
  - [ ] Local storage: ?
  - [x] Query parameter: If exists set app config value to query parameter value
  - [x] Query parameter: If exists set app config value to query parameter value if it does not use app config but do **NOT** set query parameter

Closes #1263

## Checklist


web/app/src/components/SearchBar.vue
```diff
@@ -41,14 +43,37 @@
...
 import { Search } from 'lucide-vue-next'
 import { Input } from '@/components/ui/input'
 import { Select } from '@/components/ui/select'
 
-const searchQuery = ref('')
-const filterBy = ref(localStorage.getItem('gatus:filter-by') || (typeof window !== 'undefined' && window.config?.defaultFilterBy) || 'none')
-const sortBy = ref(localStorage.getItem('gatus:sort-by') || (typeof window !== 'undefined' && window.config?.defaultSortBy) || 'name')
+const STORAGE_KEYS = {
+  FILTER_BY: 'endpointFilterBy',
+  SORT_BY: 'endpointSortBy',
```
PythonGermany:
```suggestion
  FILTER_BY: 'gatus:filter-by',
  SORT_BY: 'gatus:sort-by',
```

amai2012:
@PythonGermany  Do you plan to finish this?
According to the checklist at least docs and maybe some tests are missing.

PythonGermany:
@amai2012 I am not working on it at the moment. Maybe I'll pick it up again when the summer ends sometime around October. No promises though. If you are interested in the feature and want progress sooner you are free to continue my work and open your own PR.

Also tests are probably not needed since it's only UI changes and this project does not have unit tests for the UI as far as I can remember.

PR #1421, opened 2025-12-07
ci(dependabot): Add npm package ecosystem

PythonGermany:
## Summary

Also monitor npm dependency versions with dependabot.

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,3 +11,9 @@ updates:
     labels: ["dependencies"]
     schedule:
       interval: "daily"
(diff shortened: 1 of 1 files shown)


TwiN:
I don't know about this. NPM has too many packages with security issues lately. I would rather not update npm packages as often, at the cost of trivial frontend CVEs, than risk accidentally releasing Gatus with a malicious package because everything gets auto updated.

PythonGermany:
I understand. We could set the interval to weekly or monthly instead of daily or maybe even quarterly?

PR #1422, opened 2025-12-07
chore(ui): Remove unnecessary eslint rule disables

PythonGermany:
## Summary

I have no prior experience with eslint so maybe I did not understand the reason as for why the rule disable statements where there in the first place. With the disable statements removed the build still succeeds and there are no linting rule violations.

## Checklist


TwiN:

PR #1423, opened 2025-12-08
fix(client): Switch websocket library

joy4eg:
- Switch to https://github.com/gorilla/websocket for WS probes
  - See https://github.com/golang/go/issues/18152
  - Also enable compression extension by default. 
- Add `t.Parallel()` for client's tests


TwiN:
I understand the changes you've made, but I don't understand the why. What made you look into changing the websocket library? Did you have any issues with the prior implementation?

joy4eg:
We have our internal WS service that we monitor. The issue with it is that it _requires_ WebSocket compression/deflate support in the handshake to connect to it.

TwiN:
Makes sense

client/client.go
```diff
@@ -394,48 +394,53 @@ func ShouldRunPingerAsPrivileged() bool {
...
-	}
-	if headers != nil {
-		if wsConfig.Header == nil {
-			wsConfig.Header = make(http.Header)
-		}
-		for name, value := range headers {
-			wsConfig.Header.Set(name, value)
+	var (
+		dialer = websocket.Dialer{
+			EnableCompression: true,
```
TwiN, line 401: Does setting `EnableCompression` break websocket requests that don't have compression enabled?

joy4eg: It does not. Basically, it just sets a `permessage-deflate` http header. The server may use it to enable the compression or just ignore it (if it does not support it)

TwiN:

PR #908, opened 2024-11-20
Add subpath support 

simonhammes:
## Summary

#88

## Checklist

- [ ] Tested and/or added tests to validate that the changes work as intended, if applicable.
- [ ] Updated documentation in `README.md`, if applicable.

PR #1424, opened 2025-12-08
Base context path can be changed (#88)

s4s0l:
## Summary

Issue: #88 (changing base context root for hosting behind url rewriting reverse proxy)

- based on work done here: https://github.com/simonhammes/gatus/tree/88-dynamic-base, that exists as a draft PR (https://github.com/TwiN/gatus/pull/908)

- Additions:
   - docker compose for testing reverse proxy scenario
   - rolled back 'SERVER_URL' in order to allow hosting ui via node during development
   - updated manifest.json so it will honour relative paths

## Checklist
  - make run && make frontend-run, seems to work
  - make docker-build-and-run works also
  - I've allowed myself to add compose file that runs gatus behind caddy, to proove it actually works.
    - While behind proxy installed an 'app' from edge - icons and stuff were present in my 'start' menu
- [x] Updated documentation in `README.md`.
 
## Notes:
- I didn't try to make 'make test' not - fail on my PC. Before and after my changes same 2 tests fail - i assume they have nothing to do with UI anyway (DNS and ICMP tests).
- For anyone wanting to test it or use my version there is 'ghcr.io/s4s0l/gatus:v0.0.0.0.4' image that contains this PR.


.examples/docker-compose-reverse-proxy/compose.yml
PythonGermany, line 1: Moving this to `.examples/docker-compose-caddy/compose.yaml` or `.examples/docker-compose-proxy/compose.yaml` would make the example showcasing the new functionality consistent with other docker compose examples.

Makefile
```diff
@@ -33,6 +33,8 @@ docker-run:
 
 docker-build-and-run: docker-build docker-run
 
+docker-run-proxied: frontend-build docker-build
```
PythonGermany: I don't think this is needed as there are also no makefile targets for the other examples

.examples/docker-compose-reverse-proxy/compose.yml
```diff
@@ -0,0 +1,46 @@
...
+        base: /gatus/
+  caddyfile:
+    content: |
+      {
+        auto_https off
+        http_port 8080
+      }
+      
+      http://localhost:8080 {
+        handle_path /gatus* {
```
PythonGermany:
```suggestion
        handle_path /gatus/* {
```
In order to not proxy path `/gatus` to the gatus service

s4s0l: Ok, but why?

Other thing is that I've just tested OIDC and it does not work when using ui.base:/ I'll need a moment to get it working. Should I add an OIDC example? ... anyway sory for bothering with unfinished PR...

README.md
```diff
@@ -267,6 +267,7 @@ If you want to test it locally, see [Docker](#docker).
 | `ui.dashboard-subheading`    | Dashboard description between header and endpoints                                                                                       | `Monitor the health of your endpoints in real-time` |
 | `ui.header`                  | Header at the top of the dashboard.                                                                                                      | `Gatus`                    |
 | `ui.logo`                    | URL to the logo to display.                                                                                                              | `""`                       |
+| `ui.base`                    | `href` attribute of the HTML `<base>` tag. Use this if you want to host Gatus on a subpath (e.g. `/status/`). Has to end with '/'.       | `/`                        |
```
PythonGermany: I think this option might actually be better fit in the web config section and maybe also renaming it to make it easier to read without the actually reading the docs: `web.base-path`

PythonGermany: It should either be validated that the configured base path ends with a "/" or it should be appended at runtime if not present. The same check might also make sense for the leading "/": https://github.com/TwiN/gatus/blob/b0629773e59ab76d78b87dcd077ff8d1d250dd43/config/web/web.go#L63

Makefile
```diff
@@ -33,7 +33,6 @@ docker-run:
 
 docker-build-and-run: docker-build docker-run
```
PythonGermany, line 35:
```suggestion


```

.examples/docker-compose-reverse-proxy/compose.yml
PythonGermany: > Ok, but why?

To prevent proxying unwanted paths like /gatus or /gatus123. If /gatus should also lead to the gatus site a redirect to /gatus/ should be added.

>  ... anyway sory for bothering with unfinished PR...

No worries about outstanding issues in an opened PR. It's good to open rather sooner than later to hopefully receive valuable feedback early on.

README.md
s4s0l: This time I'll ask before doing anything. The value of the 'base path' is needed by the 'UI config' (to render index.html) and by the 'security config' (for proper redirects and cookie paths). Currently, these parts are quite isolated, and I'm about to introduce a dependency from UI & security to web config. I can think of a plethora of ways of doing it, but I get a feeling there will be a few back-and-forth discussions over it 😉. What do you think about passing 'basePath' as an argument to 'ValidateAndSetDefaults' in these different configuration objects that need it? This way it won't introduce dependencies to 'web' config everywhere, will be easily mocked in tests, and this method is mutating everything anyway...or is somewhere else a pattern I can use?

.examples/docker-compose-reverse-proxy/compose.yml
s4s0l: personally i find it annoying when i forget to write final / in url ;)

README.md
s4s0l: Currently commited version does that. It is by no means final; still missing some validation and tests, but it works.

s4s0l: went with appending.

.examples/docker-compose-reverse-proxy/compose.yml
PythonGermany: If the path `/gatus` should work it can be configured like this:
```caddyfile
http://localhost:8080 {
  redir /gatus /gatus/
  handle_path /gatus/* {
    reverse_proxy http://gatus_behind_proxy:8080
  }
}
```
For the example I would leave it out though since it is not required for it to work and I think keeping the examples as simple as possible is probably preferred.

README.md
PythonGermany: Needs to be updated

README.md
PythonGermany: > What do you think about passing 'basePath' as an argument to 'ValidateAndSetDefaults' in these different configuration objects that need it?

I think your suggested solution passing it as argument might not scale very well in case there are more options that have the same problem are added in the future.

I opened a [PR](https://github.com/s4s0l/gatus/pull/1) on your fork with another solution that so far looks like it is consistent with the `MaximumNumberOfResults` storage option where we would not need to add any additional arguments or package dependencies.

s4s0l: Doing this 'your' way has a drawback, that it makes it difficult to validate the RedirectURL in oidc configuration, as validation occurs before settings are distributed. But I'm not entirely sure there is a universe where not including  a base-path in redirect url is a desired thing.

Anyway I think I've finished.

.examples/docker-compose-reverse-proxy/compose.yml
s4s0l: ;) exact reason why i did /gatus* ;) But I get it: some llm will train on it and someone, somewhere will be suprised by 404:)

security/oidc.go
```diff
@@ -57,19 +67,20 @@ func (c *OIDCConfig) initialize() error {
 }
 
 func (c *OIDCConfig) loginHandler(ctx *fiber.Ctx) error {
+	// ??? state and nonce are not 'secure random'
```
PythonGermany, line 70: I do not quite understand what you mean. Can you elaborate on it?

s4s0l: Note to myself to think about it and open separate issue. Didn't intend to push it. Not related to this PR. 

What I was wondering was that 'state' in oauth flow is kinda important. This value should not be easily guessable. I don't know how UUID is implemented here but UUIDs are usually quite 'guessable'. By 'secure random' i meant something like https://docs.oracle.com/javase/8/docs/api/java/security/SecureRandom.html. crypto in go probably has something like this.

PythonGermany:
We missed a location where we need to update the base path:
https://github.com/TwiN/gatus/blob/862613ef3b4c7de91895508699b2bf93d88fc897/web/app/src/components/Settings.vue#L133-L135

s4s0l:
There is https://github.com/TwiN/gatus/pull/1430, I think i'll wait with rebase until it  gets merged.

web/app/src/components/ResponseTimeChart.vue
PythonGermany, line 36: Should the default also be changed?

s4s0l: Probably. In my opinion default should be removed. It is used only once, and where it is used it is set explicitly. Default for this kind of stuff does not make much sense to me. If it is not set consequently throughout the application it will not work, but will not fail during type checking / linting.

oschrenk:
Would love to see this moving forward. Great work!

m-emelchenkov:
Please tell, what's its current status of this PR? Why stalled at Dec 17?

config/config.go
```diff
@@ -301,13 +301,13 @@ func parseAndValidateConfigBytes(yamlBytes []byte) (config *Config, err error) {
...
-		if err := ValidateSecurityConfig(config); err != nil {
+		if err := ValidateWebConfig(config); err != nil {
 			return nil, err
 		}
-		if err := ValidateEndpointsConfig(config); err != nil {
+		if err := ValidateSecurityConfig(config); err != nil {
 			return nil, err
 		}
-		if err := ValidateWebConfig(config); err != nil {
+		if err := ValidateEndpointsConfig(config); err != nil {
```
TwiN, line 310: Why was the order changed?

web/app/src/App.vue
```diff
@@ -133,7 +133,7 @@
           </div>
           
           <a
-            :href="`${SERVER_URL}/oidc/login`"
```
TwiN, line 136: Just thinking out loud, are we making sure that SERVER_URL is never something like "https://example.com", which in this case would cause this string to parse as "https://example.orgoidc/login"

web/app/src/views/Home.vue
```diff
@@ -64,14 +64,14 @@
...
-                <span v-if="calculateUnhealthyCount(items.endpoints) + calculateFailingSuitesCount(items.suites) > 0" 
+                <span v-if="calculateUnhealthyCount(items.endpoints) + calculateFailingSuitesCount(items.suites) > 0"
                       class="bg-red-600 text-white px-2 py-1 rounded-full text-sm font-medium">
                   {{ calculateUnhealthyCount(items.endpoints) + calculateFailingSuitesCount(items.suites) }}
                 </span>
                 <CheckCircle v-else class="h-6 w-6 text-green-600" />
               </div>
             </div>
-            
+
```
TwiN, line 74: Can you revert these empty changes?

web/app/src/views/Home.vue
```diff
@@ -229,88 +229,88 @@ const resultPageSize = 50
 
 const filteredEndpoints = computed(() => {
   let filtered = [...endpointStatuses.value]
-  
+
   if (searchQuery.value) {
     const query = searchQuery.value.toLowerCase()
-    filtered = filtered.filter(endpoint => 
+    filtered = filtered.filter(endpoint =>
```
TwiN, line 235: ?

TwiN (changes requested):
Looks good, but there's a handful of linting changes that seem to be IDE-specific, can you revert those?

shornabho:
@s4s0l are you still working on this? Would love to see this moving forward, and it would solve a major problem for many of us. 

Keep up the good work, thanks!

shornabho:
Considering this PR has become stale and diverged from current master. I have taken the liberty to create a squashed and rebased version in a _draft_ PR in #1725. Still testing it, will make it ready for review once done.

PR #1427, opened 2025-12-09
Add SFTP support

ss-rstocchi:
## Summary

Added SFTP support to Gatus. We were in need of the ability to check for files on an SFTP server without TTY. When trying with the original SSH method an error would occur that no TTY was available.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ Have any feedback or questions? [Create a discussion](https://github.com/TwiN/ga
   - [Monitoring an endpoint using ICMP](#monitoring-an-endpoint-using-icmp)
   - [Monitoring an endpoint using DNS queries](#monitoring-an-endpoint-using-dns-queries)
   - [Monitoring an endpoint using SSH](#monitoring-an-endpoint-using-ssh)
+  - [Monitoring an endpoint using SFTP](#monitoring-an-endpoint-using-sftp)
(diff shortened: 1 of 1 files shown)


TwiN:
> When trying with the original SSH method an error would occur that no TTY was available

Does this occur before, or after the authentication has succeeded? Because if it occurs _after_ the authentication succeeded, it may be possible to just have the code that verifies SSH also handle endpoints prefixed with `sftp://` by passing a flag that marks the `no TTY was available` as a success, as opposed to introducing an additional dependency

ss-rstocchi:
@TwiN I don't recall. After the holidays I'll give it a shot and adjust the PR accordingly. In the meantime Happy Holidays!

ss-rstocchi:
@TwiN

Good morning. I've attempted to just allow for `sftp://` for ssh in `endpoint.go`:

image

This allows a connection for SFTP without shell to happen but I cannot run any commands using:

` body: |
      {
        "command": "ls -la /test"
      }
`

And picking it up in the body. I think this is because it's attempting to run commands as a shell when we don't have a shell or shell is disallowed in SFTP?

So allowing for `sftp://` would work if we want to only monitor if SFTP without shell (or disallowed shell) is able to be connected to but not for checking files within any directories on the SFTP server.

PR #1429, opened 2025-12-10
refactor(ui): Improve dev environment

PythonGermany:
## Summary

I was trying to test UI changes in a GitHub codespace using `npm run serve`. However I ran into issues when opening the dashboard because of the hardcoded backend url `http://localhost:8080` and in  ports are forwarded with a hostname pattern like this `https://CODESPACENAME-ID-PORT.app.github.dev/`.

I figured out how to use the vue [devServer](https://webpack.js.org/configuration/dev-server/) config to solve the issue.

## Checklist

- [x] Tested and/or added tests to validate that the changes work as intended in GitHub codespace and local setup.
  - Automatic reloading via websocket connection works as expected

### DIFF
--- a/web/app/src/views/EndpointDetails.vue
+++ b/web/app/src/views/EndpointDetails.vue
@@ -213,7 +213,6 @@ import Settings from '@/components/Settings.vue'
 import Pagination from '@/components/Pagination.vue'
 import Loading from '@/components/Loading.vue'
 import ResponseTimeChart from '@/components/ResponseTimeChart.vue'
-import { SERVER_URL } from '@/main.js'
 import { generatePrettyTimeAgo, generatePrettyTimeDifference } from '@/utils/time'
 
 const router = useRouter()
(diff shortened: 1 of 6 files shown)


web/app/src/views/EndpointDetails.vue
```diff
@@ -228,7 +227,6 @@ const resultPageSize = 50
 const showResponseTimeChartAndBadges = ref(false)
 const showAverageResponseTime = ref(false)
 const selectedChartDuration = ref('24h')
-const serverUrl = SERVER_URL === '.' ? '..' : SERVER_URL
```
PythonGermany, line 231: I do not quite understand what the logic here was supposed to achieve. Maybe someone can give me context to help me understand.

s4s0l: In other PR I've removed it also. Don't know the history but I'm pretty sure it is a dead code, this condition will never be fulfilled as server_url can be only http://localhost:8080 or empty. Looks like some time ago fetch requests must have been made with relative paths, and endpoint-details are in subpath hence '..', so condition was detecting if server_url is relative or absolute?

TwiN (approved):
So long as you start the backend first and then the frontend, things should work out locally, but yeah with codespace I imagine the hardcoded server URL wouldn't work

PR #1430, opened 2025-12-10
refactor(ui): Cleanup template option handling

PythonGermany:
## Summary

In the development environment the UI config placeholders are not replaced.

Improves how configs which have template placeholders as default value (replaced at go application startup by actual config value, as far as I know) are initialized and accessed in the web app.

## Checklist


web/app/src/App.vue
```diff
@@ -176,26 +177,18 @@ const tooltipIsPersistent = ref(false)
 let configInterval = null
 
 // Computed properties
-const logo = computed(() => {
-  return window.config && window.config.logo && window.config.logo !== '{{ .UI.Logo }}' ? window.config.logo : ""
-})
```
PythonGermany, line 181: I have yet to understand why sometimes `typeof window !== 'undefined'` is checked before accessing a property of `window` and sometimes it is not. Seems inconsistent unless I'm missing something here.

TwiN: Where? None of the cide you changed seems to have it

PythonGermany: In those two files it is checked before the `config` property is accessed:
https://github.com/TwiN/gatus/blob/2beaca57004d2c06dc1b10631ac75024c3d1c7af/web/app/src/components/Pagination.vue#L53
https://github.com/TwiN/gatus/blob/2beaca57004d2c06dc1b10631ac75024c3d1c7af/web/app/src/components/SearchBar.vue#L50-L51

s4s0l: SSR, unit tests or linting rule? None of which apply to this codebase IMO.

PythonGermany: Is it possible to check somewhere at initialization if window is undefined once and if it is because of the environments you mentioned just define it and set it to null? To be on the "safe" side but also not have to remember checking it each time.

PythonGermany: The [globalThis](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) property seems to be the solution I was searching for.

web/app/package.json
```diff
@@ -39,7 +39,8 @@
   "eslintConfig": {
     "root": true,
     "env": {
-      "node": true
+      "node": true,
+      "es2024": true
```
PythonGermany, line 43: I'm not sure about this yet. Without this eslint complains about `globalThis` not being defined. But I still haven't understood the impact of the change and which version to choose.

PythonGermany:
@TwiN  Just as I had worried about, the rights for me are obviously not the same as in my fork which prevents me from running the command, even as author of the PR :disappointed:

Can you test if you can trigger it in this PR to commit to the branch?

We might have to whitelist the author of this PR using their username after all.

PR #1432, opened 2025-12-10
fix(ui): Do not store config locally on load

### LINKED ISSUE #1431: Default sort does not apply existing users
### Describe the bug

The documentation [here](https://github.com/TwiN/gatus/tree/a55eb7da406fe7d4ec3bd19845eace0d7dde4132?tab=readme-ov-file#how-do-i-sort-by-group-by-default) is inconsistent with the application behaviour:

> Note that if a user has already sorted the dashboard by a different field, the default sort will not be applied unless the user clears their browser's localstorage.

### What do you see?

Once a user has visited the site for the first time they will not receive any application wide sort and filter default value changes until they clear their browser's localstorage.

The server side global config changes are not applied anymore for a user ONLY after they manually changes their local defaults via the UI elements which in turn stores them in localstorage.

1. Visit site with clear localstorage
2. Change application wide default of service via yaml config
3. Visit site again (reload or new tab)
4. Observe how application wide default is not applied

### Version

v5.33.0

### Additional information

_No response_

PythonGermany:
## Summary

Closes #1431

## Checklist

  - Local test successful. Issue not reproducible anymore.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,7 +28,7 @@ jobs:
         # was configured by the "Set up Go" step (otherwise, it'd use sudo's "go" executable)
         run: sudo env "PATH=$PATH" "GOROOT=$GOROOT" go test ./... -race -coverprofile=coverage.txt -covermode=atomic
       - name: Codecov
-        uses: codecov/codecov-action@v5.5.1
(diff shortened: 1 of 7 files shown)


TwiN:

PR #1437, opened 2025-12-11
ci: Add platform input for custom action workflow

PythonGermany:
## Summary

Adds a choice input for the publish-custom workflow docker build platform. Unfortunately there is no option to configure a structured multiple choice input. The alternative would be to have a string input with the same default. Then a comma separated list could select multiple platforms at once.

This workflow seems to be the only one that doesn't require a dockerhub credentials as far as I have seen so far. I would like to select the platform as I want to deploy development versions via docker on arm devices.

## Checklist


TwiN (approved):

TwiN:

PR #1438, opened 2025-12-12
fix(ui): Persist show average response time setting

PythonGermany:
## Summary

Adds local storage persistence to the average response time setting. This way the setting does not have to be changed again each time a user visits the site a value different from the default is preferred.

## Checklist


web/app/src/views/EndpointDetails.vue
```diff
@@ -225,7 +225,7 @@ const events = ref([])
 const currentPage = ref(1)
 const resultPageSize = 50
 const showResponseTimeChartAndBadges = ref(false)
-const showAverageResponseTime = ref(false)
```
PythonGermany, line 227: I saw that the default here is `false` and in the Home view it is `true`. Is there a reason for this difference? Should they be aligned?

With the added persistence should the setting in one view also affect the configured value of the other view?

TwiN: If we're going to persist it via local storage, then they should be aligned, yeah

PythonGermany: Ok. Would average response time or min max response time be preferred as application wide default?
Edit: I changed it to average as consistent default. I hope that's ok.

PR #1441, opened 2025-12-12
fix(ui:): Resolve endpoint card result tooltip issues

### LINKED ISSUE #1440: Tooltip flickers when moving between status result bars
### Describe the bug

Because of the spacing in between the status result bars the tooltip flickers when moving the mouse left and right over multiple status results.

### What do you see?

This is the element I call "status result bar":
Image

Tooltip stays open and updates data according to currently hovered status result.

_No response_

### Version

latest

### Additional information

_No response_

PythonGermany:
## Summary

Fixes #1440

In addition it also fixes various other issues regarding the endpoint card tooltip. See checklist.

## Checklist

- [x] ~~ISSUE: Tooltip persistence not enabled when clicking between the results~~
- [x] ~~ISSUE: Empty tooltip is shown for a short amount of time before tooltip is hidden~~
  - ~~Found the cause -> Because the transition has a time configured. If I disable the transition the artifact is gone, but then location changes between results are also instant~~
- [x] ISSUE: Close persistent tooltip when clicking on result with no data
- [x] TODO: Set cursor to always to pointer when hovering over history bar
  -  Only makes sense once enabling persistence always works
  -  With the exception of hovering over results that have no data
- [x] TODO: Test how it behaves with suites (haven't used that feature so far)
- [x] ISSUE: Edge cases possible leading to outdated tooltip (especially relevant when using pinned tooltip)
  - [x] TODO: Handle for pinned tooltips
  - [x] TODO: Handle for unpinned tooltip


PythonGermany:
The updated UI behavior:

https://github.com/user-attachments/assets/33e47378-75ab-490c-a100-105270adbdbb

PythonGermany:
image
It's possible for the tooltip to be outdated by pinning the tooltip for one result and then waiting for the site to refresh. Which might already happen after 10s. Once outdated clicking on the updated result will update the tooltip inplace.

To solve this one option would be to always hide the tooltip when the results are updated, however this would lead to other possible issues, e.g tooltip disappears without user input just after opening it.

Updating the location of the tootip might be more complicated since I would need to figure out which one of the updated results is the previously selected one, if it still exists, since it might also have moved out of scope of the displayed results.

So updating the content of the tooltip might be the best way to move forward.

PythonGermany:
I cannot get to a working solution for the two oustanding issues:
- Still no persistent tooltip enabled when clicking beween results
- Tooltip opens before it is has the full data loaded and correct position set due to transition duration

PythonGermany:
> I cannot get to a working solution for the two oustanding issues:
> 
>     * Still no persistent tooltip enabled when clicking beween results
> 
>     * Tooltip opens before it is has the full data loaded and correct position set due to transition duration

My suggestion would be to leave it as is and tackle the remaining issues in a new PR/Issue -> https://github.com/TwiN/gatus/issues/1458

TwiN:
@PythonGermany Would you say the two issues you outlined is a regression compared to the current behavior in the master branch (despite the new feature)?

I have no problem splitting an issue in smaller sub-issues, but if it leaves the latest version of Gatus in a state where user experience is impacted until the sub-issue(s) are resolved, I would rather find a solution before merging the initial/faulty implementation.

PythonGermany:
> Would you say the two issues you outlined is a regression compared to the current behavior in the master branch (despite the new feature)?

Both issues mentioned are already present in the master branch with exactly the same observable behavior. They are **not** introduced by the changes of this PR. Just haven't found a good solution so far, that's why they are still unresolved.

PR #1442, opened 2025-12-12
feat(ui): Allow to configure config refresh interval

PythonGermany:
## Summary

Adds config option to customize config refresh interval in the UI (refresh is used to get update announcements) 

## Checklist


config/ui/ui_test.go
```diff
@@ -277,3 +297,41 @@ func TestConfig_ValidateAndSetDefaults_DefaultFilterBy(t *testing.T) {
...
+		Name                  string
+		ConfigRefreshInterval time.Duration
+		ExpectedError         error
+		ExpectedValue         time.Duration
+	}{
+		{
+			Name:                  "ZeroConfigRefreshInterval",
+			ConfigRefreshInterval: 0,
+			ExpectedError:         nil,
+			ExpectedValue:         defaultConfigRefreshInterval,
```
PythonGermany: Not sure if that is whats expected. Like this it behaves the same as the endpoint interval option, there setting the value to zero does not make the config invalid but instead just uses the `1m` default

PythonGermany: The issue could be solved in a similar way to how config bool values are handled:
https://github.com/TwiN/gatus/blob/1df0801a61c3341cad7e495229e151b5241cdc6f/config/ui/ui.go#L118-L120
For a duration this would be:
- nil -> set default
- zero -> return error

Only question would whether to make a new pr where this is also "solved" for endpoint interval config and possible other occurrences. Changing this would only be a breaking change for invalid configurations.

TwiN:
Is there really value in making this configurable? I'm tempted to just pick a sensible value like 15 minutes and go with that. After all, announcements currently require a config change (which means a redeployment for most), so anything faster than that likely doesn't make a big difference in the ux in practice.

At some point, maybe we'll switch to websockets instead of repeatedly fetching from the API, but for now, I think it's fine to keep it simple.

PythonGermany:
I think you're right. Currently the added complexity is not really justified.

PR #1443, opened 2025-12-13
feat: Add build information

### LINKED ISSUE #833: Show version number on page

Display the application version number on the Dashboard page

to see what version the application is running

not long

### Additional information

_No response_

### LINKED ISSUE #901: Provide a way to know which version of Gatus the binay is

I've not found any mechanism to know which version of Gatus I'm running.
It is important to ensure that I run the latest version, and for rollback purposes.

There are multiple way to provide the information, I'll describe a few here.

## In the log
Having the version displayed on startup/in the logs would be a good starting point.

**Pros**
It is private to the person running Gatus.
Logs often get exported in different places where they can be acted upon too.
**Cons**
On its own it might be limited.

```
2024/11/18 10:36:51 [StartupMessage] Current version of Gatus is v5.13.1
2024/11/18 10:36:51 [config.LoadConfiguration] Reading configuration from config/config.yaml
2024/11/18 10:36:51 [config.validateAlertingConfig] Alerting is not configured
2024/11/18 10:36:51 [config.validateEndpointsConfig] Validated 1 endpoints
2024/11/18 10:36:51 [config.validateEndpointsConfig] Validated 0 external endpoints
2024/11/18 10:36:51 [controller.Handle] Listening on 0.0.0.0:8080
```

Example **forgejo**:
```
# journalctl -b -u forgejo|grep version|head -n 1
cmd/web.go:111:showWebStartupMessage() [I] Forgejo version: 8.0.3+gitea-1.22.0 built with GNU Make 4.4.1, go1.22.7 : bindata, timetzdata, sqlite, sqlite_unlock_notify
```

## As a cli parameters
Having a dedicated `-v` or `--version` or `version` as cli switch to get the version is useful to check the right binary is at hand without running the configuration.

**Pros**
It is private to the person running Gatus.
Can be done without running the server with a configuration
**Cons**
There is no cli switches in Gatus right now, so it might be a design decision not to have any. 

Example **goatcounter**:
```
$ goatcounter version
version=v2.5.0; go=go1.21.5; GOOS=linux; GOARCH=amd64; race=false; cgo=true
```

## In the UI
Having the version in the UI, maybe as a footer could help quickly get the version without login to a server

**Pros**
It is quick to see the version
**Cons**
It is not private to the person running Gatus.

Example **owncast** :
```
Powered by [Owncast v0.1.3](https://owncast.online/)
```

Having a programmatically sound way to know the version would help me automate my Gatus installs with more confidence

1 months

### Additional information

_No response_

PythonGermany:
## Summary

Add application wide build informations via
- startup via info log
- UI footer with link to github release page (configurable if it should be displayed)
  - [x] TODO: Don't add link to keep it simple? -> Removed
  - [x] TODO?: Only add link to github releases page for correct version pattern `^v\d+\.\d+\.\d+$`
  - [x] TODO?: Make link to release page configurable. If it is enabled and/or where it points to.
- prometheus endpoint metric `gatus_build_info`
  - [x] Add go_version label

Closes #833
Closes #901

## Checklist

  - [x] Test workflows
    - [x] Version
    - [x] Build date
    - [x] Commit hash/revision
  - [x] Test docker and direct build
    - [x] Check version with and without tag
      - [x] via make
      - [x] via docker
    - [x] Check commit hash via make, docker and directly building (without setting any custom -X ldflags)
    - [x] Check build date via make and docker and directly building (without setting any custom -X ldflags)
  - [x] Test new `ui.show-version` option (make sure if it is false that the version is not embedded in the UI files) -> Version placeholder in index.html is empty if set to false


buildinfo/buildinfo.go
```diff
@@ -0,0 +1,19 @@
+package buildinfo
+
+var version = "dev"
```
PythonGermany: What should the default be if not replaced at build time? `dev`, `development`, `unknown` or something different altogether?

web/app/src/App.vue
```diff
@@ -101,6 +101,9 @@
           <div class="flex flex-col items-center gap-4">
             <div class="text-sm text-muted-foreground text-center">
               Powered by <a href="https://gatus.io" target="_blank" class="font-medium text-emerald-800 hover:text-emerald-600">Gatus</a>
+              <a v-if="buildVersion" :href="releaseLink" target="_blank">
+                &nbsp;{{ buildVersion }}
```
PythonGermany: I'm not quite satisfied with how this turns out in the UI. The space between the text before and the version is wider than a normal space but I haven't figured out how to easily make it work with a normal space.

Dockerfile
```diff
@@ -1,10 +1,18 @@
...
+
 RUN apk --update add ca-certificates
 WORKDIR /app
 COPY . ./
 RUN go mod tidy -diff
-RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o gatus .
+RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo \
+-ldflags "-X github.com/TwiN/gatus/v5/buildinfo.version=${VERSION} \
+-X github.com/TwiN/gatus/v5/buildinfo.commitHash=${COMMIT_HASH} \
+-X github.com/TwiN/gatus/v5/buildinfo.date=$(date -u +'%Y-%m-%dT%H:%M:%SZ' || echo 'unknown')" \
```
PythonGermany: This causes issues because it kills the cache mechanism of the layer. Maybe it's fine this way but it would certainly be an improvement to somehow have the date only update if the source of the copy layer changed.

joy4eg: You could extract build information using [debug](https://pkg.go.dev/runtime/debug#BuildInfo) package.
Something like:
```go
func readVersionString() (commitDate, commit string) {
	info, ok := debug.ReadBuildInfo()
	if !ok {
		return "unknown", "unknown"
	}

	for _, setting := range info.Settings {
		switch setting.Key {
			case "vcs.revision":
				commit = setting.Value
			case "vcs.time"
				commitDate = setting.Value
		}
	}

	return commitDate, commit
}
```

PythonGermany: I did not know about that package, thanks for the help! I added it to check for build info if none was set via ldflags. The only issue remaining is that there is no git inside to golang dockerimage, so there I still have to pass the build info retrieved in the host environment.

buildinfo/buildinfo.go
PythonGermany: Went with `development` for now.

main.go
```diff
@@ -26,6 +27,8 @@ func main() {
 		logr.Infof("Delaying start by %d seconds", delayInSeconds)
 		time.Sleep(time.Duration(delayInSeconds) * time.Second)
 	}
+	buildInfo := buildinfo.Get()
+	logr.Infof("Starting Gatus (version: %s, revision: %s, build-date: %s)", buildInfo.Version, buildInfo.Revision, buildInfo.Date)
```
PythonGermany:
```suggestion
	logr.Infof("Starting Gatus (version: %s, revision: %s, revision-date: %s)", buildInfo.Version, buildInfo.Revision, buildInfo.Date)
```
I'm also thinking about removing it altogether since the revision date doesn't say anything about the actual build date the date can always be looked up in version history anyway if needed.

README.md
```diff
@@ -275,6 +275,7 @@ If you want to test it locally, see [Docker](#docker).
 | `ui.dark-mode`               | Whether to enable dark mode by default. Note that this is superseded by the user's operating system theme preferences.                   | `true`                     |
 | `ui.default-sort-by`         | Default sorting option for endpoints in the dashboard. Can be `name`, `group`, or `health`. Note that user preferences override this.    | `name`                     |
 | `ui.default-filter-by`       | Default filter option for endpoints in the dashboard. Can be `none`, `failing`, or `unstable`. Note that user preferences override this. | `none`                     |
+| `ui.show-build-info`         | Whether to show build information (version in footer).                                                                                   | `true`                      |
```
TwiN: The fact that you felt the need to clarify what build info meant between parentheses makes me think it may be more straightforward to write it as `ui.show-version` rather than `ui.show-build-info`. 

Furthermore, I'm a little perplexed on whether it should be `show-version` or `hide-version`, the difference being that the former would cause the default to not display the version in the footer, while the latter would default to showing the version in the footer.

Would the average Gatus user want the version to be displayed out of the box?

PythonGermany: > The fact that you felt the need to clarify what build info meant between parentheses makes me think it may be more straightforward to write it as ui.show-version rather than ui.show-build-info.

You're right. I thought at the beginning that I might add some more fields, but now I think more than the version should, most of the time, not be needed.

> Would the average Gatus user want the version to be displayed out of the box?

Good point. I think as long as the version is always available in the logs logs and the metrics endpoint most running instances will have no need to have this enabled. So not showing it by default in the UI is probably a suitable default.

Only question then is if we then want to name it `show-version` with default **false** or `hide-version` with default **true**. The name `hide-version` would be in line with the endpoint specific UI option names.

PythonGermany:
Currently implemented ways to get the build info and info types available:
- Logs: Version, Revision, Revision date
- UI: Version (configurable)
- Prometheus metrics endpoint: Version, Revision, Revision date, Go version

PythonGermany:
@TwiN Would exposing the build info via the API be of use and a wanted addition? Or would that be redundant/irrelevant/unwanted since the API already has it's own version?

README.md
PythonGermany: Went with `show-version` and default `false` for now. Hope that's ok. Personally I prefer `show-*` over `hide-*` because the former is easier to read in my opinion.

TwiN:
Exposing it via the UI is enough for now.

At some point, I imagine we'll also have to expose it via cli (`gatus version` or `gatus --version`), but there's no need to implement this right now.

amai2012:
I would like to see this.
What's next? 

Thorough review by TwiN?

amai2012:
@TwiN  Any chance to proceed here?

PR #1445, opened 2025-12-14
feat: Add custom states

### LINKED ISSUE #227: RFE: Add support for severity
I see two ways this could be implemented, with Option 2 being super versatile but becomes more complex.

#### Simple Severity
Each test has an additional option that allows you to set a 1-10 severity rating.
```
# Notice the prefixed severity on each condition.  This should make existing configurations compatible.
  - name: webserver1
    group: customer_portal
    url: "http://web1/api/v3/status"
    interval: 5m
    conditions:
      - "10::[STATUS] == 200"
      - "5::[BODY].errors == false"
      - "3::[BODY].unknownErrors == false"
      - "2::[RESPONSE_TIME] < 150"
 ```
>None = Green = 0  #This would mean the test was successful
Low = Yellow = 1-4
Medium = Orange = 5-8
High = Red = 9-10

#### Advanced Severity
Each test has an additional option that allows you to set a 1-10 severity rating.
After **all** tests are completed, sum up the severity points and then evaluate on the sum for the entire set of tests.
```
# This uses a different yaml structure and would require admins to update their configuration.
- name: webserver1
  group: customer_portal
  url: http://web1/api/v3/status
  interval: 5m
  conditions:
    status:
      severity: 10
      test: '[STATUS] == 200'
    errors:
      severity: 5
      test: '[BODY].errors == false'
    unknown_errors:
      severity: 3
      test: '[BODY].unknownErrors == false'
    response_time:
      severity: 2
      test: '[RESPONSE_TIME] < 150'
```
>None = Green = 0  #This would mean the test was successful
Low = Yellow = 1-4
Medium = Orange = 5-8
High = Red = 9-10

### LINKED ISSUE #1274: Issues with maintenance-windows
### Describe the bug

Maintenance Windows are not working for end points or across the whole board.
I've tried the following configuration and also tried it with no time zone and matched it with UTC  and the end point won't go into maintenance and is marked as healthy and will alert if left for too long.

Image

### What do you see?

Not going into maintenance and is being marked as unhealthy.

The end point going into maintenance or the board going into maintenance when set globally.

1. Set maintenance window configuration per the docs.
2. Wait until next refresh.
3. Down the end point.
4. Watch it get marked as unhealthy on next interval.

### Version

5.23.0

### Additional information

```
  - name: pangolin
    group: core
    url: "https://nuhuh.com"
    maintenance-windows:
      - start: "15:54"
        duration: 2h
        timezone: "Pacific/Auckland"
    interval: 3m
    conditions:
      - "[CONNECTED] == true"
    alerts:
      - type: discord
        failure-threshold: 2
        success-threshhold: 2
        description: "A core application HAS GONE DOWN! Please manually create an incident and log what happened!"
    ui:
      badge:
        response-time:
          thresholds: [1100, 1300, 1500, 1750, 1900]
```

I have tried with my timezone and just via UTC which is what the docker container operates and it won't ignore nor will it go into maintenance.

Can update to the very latest to see if this can be reproduced.

I am probably still doing something wrong, have tripled checked config and docs this time.

### PR DESCRIPTION
**Pull request by PythonGermany**
## Summary

Implement the requested severity feature by introducing custom states.

If implementation moves forward as  suggested:
Closes #227, close #1274.

If we structure it correctly we could start working on #1444 as a followup if this PR gets merged to make it possible to configure multiple themes for each state to allow adding colorblind  and other themes which can then be selected via the UI.

### Implementation tasks:
The following logic needs to be implemented (backend):
- [x] New config section `states` defining the available states
  - [x] No custom config is given -> Set default states `healthy`, `unhealthy` and `maintenance`
  - [x] Merge custom config with default config
  - [x] Validate config
    - [x] Check name conflicts
    - [x] Check priority conflicts
- [x] New UI section configuring colors/themes for states
  - [x] No custom config is given -> Set default color config for default states
  - [x] Merge custom config with default config
  - [x] Validate config
    - [x] Verify that names in state and color config match
    - [x] Validate color code format
 - [ ] Extend alerting config
   - [ ] Add state priority threshold below which no alerting should take place (set state in config and resolve to priority at runtime)
- [x] Logic to evaluate state of result/endpoint
  - [x] In maintenance state
  - [x] Default healthy/unhealthy states -> One properly implemented also covered by severity logic
  - [x] Custom states -> Input will be condition evaluation results
    - [x] Select correct state based on highest priority
- [x] New API functionality
  - [x] Send color and UI relevant information for states
  - [x] Send new state information of result
  - [x] Include state in events to display useful info and correct color in UI
  - [ ] Handle custom states for external endpoints
    - [x] Handle healthy and unhealthy defaults
    - [ ] Allow for custom state input
      - [ ] Validate custom state against current config? If not valid default to healthy unhealty?
    - [ ] Handle maintenance state 
- [ ] Handle custom states for suites
- [x] DB Migration adding new result state column if not present and set appropriate state based on Success value

The following logic has to be implemented (frontend):
- [x] Add new placeholders to embed color config when starting gatus
- [ ] Handle current status UI text value in endpoint details view
- [x] Use new API state fields
  - [x] to display correct color and text in status badges
  - [x] to display correct color and text in endpoint events
  - [x] to display correct color and text in endpoint result history bar
  - [x] to display correct color and text for events in response time trend chart
- [x] Handle absence of custom color config (to prevent issues in development environment or state config changes)
  - Handled by using a hardcoded frontend local color used to communicate missing custom state color config 
### Outstanding issues/tasks/questions

- [x] Colors are broken when in frontend development mode -> `npm run serve`
- [x] Add new event on state change and not just on success change?
  - [ ] Make configurable?
- [ ] Test impacts of color stuff on remote instance experimental feature
- [ ] Reserve `unknown` state and allow state color config?
- [ ] Show state for result in tooltip
- [ ] Set success field based on new config option configurable for each state
- [ ] Apply custom color for default states to everything relevant regarding suites
- [x] How is config file merging handled? First merged then validated? (interest because of new config sections)
  - readme says first merged then validated

### Final tasks

### CODE REVIEW

**Comment by TwiN** [comment]
Just FYI before you commit time to implementing this - there has been a few attempts at implementing severity, and the issue is usually that it's a breaking change and I don't like the implementation. I understand some people would love a way to add some level of severity, but to me, I'm not even convinced I would want this feature in Gatus in the first place because of all the confusion it adds around health % calculation, uptime calculation, the increase in complexity for specifying what severity a specific condition impacts, etc. 

I've seen the comment you posted on the issue, and it's a good idea to avoid breaking changes, but I think it would be too difficult to understand for most people (and also `states` is perhaps not the right word here for this configuration, as it's not really self-explanatory like `severities` or something similar would be), and again, it would add a lot of complexity for people to understand in the configuration.

Long story short, it adds a whole lot of complexity for what I think is very little advantage, and for that reason, at this stage, I'm not sure I'm interested in adding this to Gatus

**Comment by PythonGermany** [comment]
Thanks for the heads up! I understand  that working on this might lead nowhere and that it's a complicated thing to add since it impacts a lot of core components. Currently I'm just exploring and trying to figure out what the biggest road blocks are towards a fully working implementation/prototype.

Out of my perspective it's a good topic because this way I need  to have a look at many different core components and features of Gatus. So even if it leads nowhere it will not be wasted time for me, I'm sure.

> The issue is usually that it's a breaking change

Breaking change in configuration compatibility is probably the most critical I imagine. How problematic are breaking changes in API and database structure (e.g. added state member, in addition to the currently existing success field)?

> I'm not even convinced I would want this feature in Gatus in the first place because of all the confusion it adds around health % calculation, uptime calculation

My goal would be to not change the default behavior in any way. Only people trying to would have to think about the implications of customizing it.

In regards to added complexity I definitely see your point. However I think is the structure on how to configure it is well designed one detailed entry in the FAQ section of the readme would be enough for most people to understand.

I'll keep this PR in draft as long as I'm experimenting and once the base of the implementation is solid and thorough enough I'll open it. Maybe you'll see the potential once I have hopefully solved some of the bigger issues, and if not so be it. I know what I signed up for!

**Comment by TwiN** [comment]
Making a breaking change on the database schema would only imply removing or modifying existing fields. It wouldn't be preferable, but it's not a big deal if we have to do that. There is no public database API contract for Gatus, meaning there is no guarantee that the database schema will not have breaking changes.

The REST API, however, is a different story.
In your case, you'd likely be adding to the API, not removing or changing existing fields in a way that would actually constitute a breaking change.

Changes are just breaking if users currently relying on a specific field are no longer able to rely on that field due to a change. Such a change would be breaking, and would require a major version bump, which can be done if absolutely necessary.

If you're not sure whether something is a breaking change, feel free to ping me and ask.

**Comment by PythonGermany** [comment]
> states is perhaps not the right word here for this configuration, as it's not really self-explanatory like severities or something similar would be

I've thought about this and I'm not sure if `severities` is more self-explanatory. I thought about calling it `endpoint-states` but that doesn't really make sense either. The best suggestion I currently have would be to call the option `result-states`, EDIT: New idea, maybe `health-states`?

**Comment by PythonGermany** [comment]
image

I've gotten the prototype far enough that it already provides the following functionality:
- Default states (healthy, unhealty and maintenance)
- States can be linked to conditions and if the condition is not successful the state will be selected
- Configurable colors for each state in the UI

**Comment by PythonGermany** [comment]
Events are now displayed with a custom text and color:
<details>
<summary>Showcase screenshot</summary>
image
</details>

In development mode the template placeholder for the `window.config` is not replaced and no custom colors are available. The current implementation uses violet if the color for the displayed state is unknown. 

Could also be changed to fallback on the Success field of the endpoint results. This situation should only arise during development if api and webserver instance are bundled as one, since the state and color config link is validated at server startup:
<details>
<summary>Showcase development screenshot</summary>
image
</details>

**Comment by PythonGermany** [comment]
@TwiN I'm getting ready for a feedback phase.

There are still some parts (and many small details) to be implemented/discussed and most of the tests are still missing but enough is already working to showcase it in a usable way.

> Long story short, it adds a whole lot of complexity for what I think is very little advantage, and for that reason, at this stage, I'm not sure I'm interested in adding this to Gatus

I'd move forward with implementing/discussing the details and cleaning up the rough edges if you are more interested after looking at the rough implementation for this feature. After working on it for a little while I see the added complexity but also the potential.

~~I've added new columns to the corresponding tables in the db schemas, and I am not sure if it will break things for existing DBs. I just did a short test locally and when switching to the feature branch with an old db gatus complains about missing columns, which makes sense. I'm just not sure how to move forward to mitigate that.~~ Figured it out!

**Comment by Xyz00777** [comment]
i really like the picture with the different colors for the state from above, im excited to see how the feature will potentially look in the future :)

**Comment by FuLLeNN** [comment]
Is there any updates on this? 
I'm waiting for this feature so I can use it at my workplace (this maintenance/multiple states would be a huge thing to us) 

@TwiN @PythonGermany

**Comment by Zophar78** [comment]
I've tested your branch and it works as intended for me (And i'm really looking forward to add a "degrated" mode for my use case which is a requirement).
Just my two-cents on the configuration: it would be "easier" to have the cutom state color directly in the "states" section like this:
```bash
states:
  - name: degraded
    priority: 1
    color: "#FFA500"  # Orange color for degraded state
```

instead of:
```bash
states:
  - name: degraded
    priority: 1

ui:
  state-colors:
    degraded: "#FFA500"  # Orange color for degraded state
```

It would replacate states in ui section as well.

marcaurele:
> Just FYI before you commit time to implementing this - there has been a few attempts at implementing severity, and the issue is usually that it's a breaking change and I don't like the implementation. I understand some people would love a way to add some level of severity, but to me, I'm not even convinced I would want this feature in Gatus in the first place because of all the confusion it adds around health % calculation, uptime calculation, the increase in complexity for specifying what severity a specific condition impacts, etc.
> 
> I've seen the comment you posted on the issue, and it's a good idea to avoid breaking changes, but I think it would be too difficult to understand for most people (and also `states` is perhaps not the right word here for this configuration, as it's not really self-explanatory like `severities` or something similar would be), and again, it would add a lot of complexity for people to understand in the configuration.
> 
> Long story short, it adds a whole lot of complexity for what I think is very little advantage, and for that reason, at this stage, I'm not sure I'm interested in adding this to Gatus

One interesting case for the degraded status is the TLS certificate expiry. It's more granular to be able to turn an endpoint as degraded if the certificate will expire in the next 7 days, and turning to unhealthy when it's expired. _Unhealthy_ is too strong in that case as it's a warning and not a service downtime state.

christophdb:
**A concrete production use case for custom states — with a note on the uptime concern**

We monitor SeaTable Cloud with Gatus on a customer-facing status page. Our backend runs **multiple dtable-servers**, but we deliberately *don't* want one Gatus endpoint per server — that's too technical and confusing for customers. Instead we expose a single **"meta" health endpoint** on our side that internally checks all dtable-servers and rolls them up into one status.

What we're missing is exactly a **third state**:
- 🟢 healthy — all servers fine
- 🟡 degraded — at least one server has issues, but the service is still reachable/usable
- 🔴 down — everything is unavailable

**Where the logic lives:** the actual aggregation ("which of our N servers is unhealthy?") stays entirely on our side, inside that meta endpoint. Gatus polls it like any normal endpoint and only needs to map the response to a state. Our endpoint always returns `200` with a small JSON body, e.g.:
```json
{ "status": "degraded", "affected": ["dtable-3"] }
```
and Gatus would select the state via a simple condition such as `[BODY].status == healthy` / `== degraded` / `== down`. So Gatus isn't computing severity across many conditions — it's a trivial status→state mapping on a single endpoint, using the same condition mechanism it already has. The complexity @TwiN was worried about doesn't enter the picture here.

**On the uptime question (the main worry raised):** in our case `degraded` must count as `up` for the uptime/availability %. Our SLA is "reachable/usable", and the service *is* available when only one backend server is degraded — the yellow state communicates the partial issue visually without distorting the availability metric. That said, I can easily imagine other operators wanting degraded to count as downtime, which is exactly why the **per-state `success` mapping** already on your checklist (configurable whether a state counts toward uptime) is the right design — it lets each operator pick the semantics instead of Gatus imposing one.

This is the same spirit as @marcaurele's TLS-expiry example: a "warning, not an outage" state that's genuinely useful and shouldn't be forced into the binary up/down model. Happy to help test this against a real multi-server setup if useful.

christophdb:
I tested this feature against the current master and it works great. 🎉

Since the branch was ~4 months behind, I merged the latest master into it first. The conflicts were all trivial (master's new LoginSubtitle vs. this branch's StateColors, plus regenerated frontend assets), none in the custom-states logic. After that: go build ./... ✅, the ui / state / sql tests pass ✅, and the three states render correctly end-to-end (the API returns state: degraded and the custom color shows up).

Here's a real-world example of why this matters for us: we run SeaTable Cloud with multiple dtable-servers behind a single "meta" endpoint. We don't want one endpoint per server (too technical for customers). We just want 🟢 healthy / 🟡 degraded (≥1 server has issues but the service is still usable) / 🔴 down. This feature does exactly that:

Screenshot_2026-06-12_11-03-51

I pushed the rebased/merged branch as a PR into your fork so it's easy to pull in if useful: PythonGermany/gatus#26. No pressure at all, just trying to help move this forward. Really hope this lands; thanks for the excellent work! 🙏

christophdb:
Sorry for the next comment... but two more thoughts:

### 1) another use case:
I've been playing around with this some more and honestly it's fantastic. 🙌

I found another use case that I think is even more universal than the multi-backend one I mentioned earlier: using a custom state for slow response times. "Slow but reachable" really shouldn't show up as a full outage, and with this feature it just works:

```
conditions:
  - "[STATUS] == 200"                  # no/error response -> unhealthy (red)
  - "degraded::[RESPONSE_TIME] < 100"  # slow -> degraded (yellow); else healthy
``` 
Here it is running on a endpoint (a login page). The result-history bar nicely shows the degraded buckets where the response time crossed the threshold, with the condition breakdown right there in the tooltip:

Screenshot_2026-06-12_13-06-29

### 2) alerting + uptime
One more thing I ran into while testing. The interaction between custom states and alerting + uptime, which I think is the key question for real-world use.

Right now a `degraded::` condition that fails sets `result.Success = false` (same as any failing condition), and both alerting and the uptime % key off that boolean. So as it stands, a degraded result triggers notifications just like a full outage (counts toward `failure-threshold`) and counts as downtime in the uptime %. I see this is already on your radar (the TODO#227 in `watchdog/endpoint.go`) and the two unchecked alerting items in the checklist.

What makes this interesting is that the same degraded state needs opposite alerting behavior depending on the check. Two real examples from our setup:

1. Slow response time (`degraded::[RESPONSE_TIME] < 1000`) should show 🟡 on the dashboard but not page anyone. It's usually transient, "slow but reachable" isn't an emergency, and it shouldn't drag down the availability SLA.
2. TLS cert expiring in <7 days (`degraded::[CERTIFICATE_EXPIRATION] > 168h`) should show 🟡 and absolutely send a notification, because that's exactly the early warning you want in order to renew in time. Here a missed alert is the whole problem.

So a single global "degraded doesn't alert" rule wouldn't work for either of us. Case 1 wants silence, case 2 wants a page, both on the same state. That's what makes me think an alertable switch per check (alongside a per-state up/down mapping for the uptime %) would be the flexible primitive here.

Is that roughly the direction you're considering for the success/priority config? Happy to help test that part once it exists and thanks again, the visual side already works great.

breart:
Are there any updates? This functionality would be extremely valuable. 

How we can move this forward? @TwiN

K-e-l-v-i-n-12:
Unfortunately, we can't use Gatus because it doesn't distinguish between maintenance and outages. We're eagerly awaiting this feature.
@TwiN : Would it be possible to get a quick response as to whether this is planned and whether it's worth waiting, or if we'll have to rely on other heavyweights after all—which we'd really rather not do?

PR #1450, opened 2025-12-17
docs: Update config section and add env var faq

PythonGermany:
## Summary

Separate explanation on environment variable functionality in config files into its own faq entry

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -3313,7 +3313,7 @@ web:
   port: ${PORT}
 
 ui:
-  title: $TITLE $$ Gatus Monitoring # With env var set to "Example" this will expand to "Example $ Gatus Monitoring"
+  title: $TITLE
 ```
 ⚠️ When your configuration parameter contains a `$` symbol, you have to escape `$` with `$$`.
 


README.md
```diff
@@ -234,14 +238,10 @@ subdirectories are merged like so:
...
 > 💡 You can also use environment variables in the configuration file (e.g. `$DOMAIN`, `${DOMAIN}`)
 >
-> See [examples/docker-compose-postgres-storage/config/config.yaml](.examples/docker-compose-postgres-storage/config/config.yaml) for an example.
+> ⚠️ When your configuration parameter contains a `$` symbol, you have to escape `$` with `$$`.
 >
-> When your configuration parameter contains a `$` symbol, you have to escape `$` with `$$`.
-
-If you want to test it locally, see [Docker](#docker).
-
+> See [Use environment variables in config files](#use-environment-variables-in-config-files) or [examples/docker-compose-postgres-storage/config/config.yaml](.examples/docker-compose-postgres-storage/config/config.yaml) for examples.
```
PythonGermany, line 245: Moved this section here because before clicking on **Configuration** in the table of contents directly showed the top level config option table. For the top level config as a new user I'd first expect a _how_ to configure and then the _what_ to configure part.

TwiN:
Can't post a review comment on GH mobile due to a bug, but to avoid confusing 99% of users who don't need to escape `$`, I don't think the example should include $$ (in `ui.title`). Just leaving it as a note below the example should be sufficient.

TwiN:

PR #1452, opened 2025-12-17
fix(ui): Inconsistent time values in UI

### LINKED ISSUE #1451: Displayed response time in UI does not match placeholder time
### Describe the bug

In the prettified placeholder the decimal places are truncated.
In the UI the value seems to be rounded.

### What do you see?

Image

Observe the response time in the placeholder is 214 and in the response time section of the tooltip it is 215. A difference is not guaranteed though -> The range where truncating the value of the prettified placeholder in the backend and rounding in the frontend leads to the same integer.

Both times 214ms.

1. Configure a response time condition that is not successful
2. Observe the difference via the UI tooltip in the home view

### Version

latest

### Additional information

_No response_

PythonGermany:
## Summary

Fixes #1451 

## Checklist

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -157,7 +157,6 @@ func TestShouldRunPingerAsPrivileged(t *testing.T) {
 	}
 }
 
-
 func TestCanPerformStartTLS(t *testing.T) {
 	type args struct {
 		address     string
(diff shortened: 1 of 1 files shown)


TwiN:
Good catch! I had actually never noticed that discrepancy

TwiN:
Can you `make frontend-build`?

TwiN:
Great work!

PR #1462, opened 2025-12-21
feat(alerting): ClickUp alerting provider

### LINKED ISSUE #1426: feat(alerting): Add support for ClickUp as an alerting provider

Add support for ClickUp as an alerting provider
REST API Docs: https://developer.clickup.com/

We use ClickUp as our project management tool and have an internal fork of this repo that has the provider already built, but would love to contribute back to the project.

~ 1 year

### Additional information

I would love to contribute our provider that we wrote internally here, just let me know if I should go ahead and clean it up, add tests, docs and create a PR. We have been using that provider internally for about a year now.

TheBinaryGuy:
## Summary

Closes #1426

## Checklist

### DIFF
--- a/alerting/provider/clickup/clickup.go
+++ b/alerting/provider/clickup/clickup.go
@@ -136,20 +136,16 @@ func (provider *AlertProvider) Send(ep *endpoint.Endpoint, alert *alert.Alert, r
 	if err != nil {
 		return err
 	}
-
 	if resolved {
 		return provider.CloseTask(cfg, ep)
 	}
-
 	// Replace placeholders
 	name := strings.ReplaceAll(cfg.Name, "[ENDPOINT_GROUP]", ep.Group)
 	name = strings.ReplaceAll(name, "[ENDPOINT_NAME]", ep.Name)
-
 	markdownContent := strings.ReplaceAll(cfg.MarkdownContent, "[ENDPOINT_GROUP]", ep.Group)
 	markdownContent = strings.ReplaceAll(markdownContent, "[ENDPOINT_NAME]", ep.Name)
 	markdownContent = strings.ReplaceAll(markdownContent, "[ALERT_DESCRIPTION]", alert.GetDescription())
 	markdownContent = strings.ReplaceAll(markdownContent, "[RESULT_ERRORS]", strings.Join(result.Errors, ", "))
-
 	body := map[string]interface{}{
 		"name":             name,
 		"markdown_content": markdownContent,
@@ -160,7 +156,6 @@ func (provider *AlertProvider) Send(ep *endpoint.Endpoint, alert *alert.Alert, r
 	if cfg.Priority != "none" {
 		body["priority"] = priorityMap[cfg.Priority]
 	}
-
 	return provider.CreateTask(cfg, body)
 }
 
@@ -169,7 +164,6 @@ func (provider *AlertProvider) CreateTask(cfg *Config, body map[string]interface
 	if err != nil {
 		return err
 	}
-
 	createURL := fmt.Sprintf("%s/list/%s/task", cfg.APIURL, cfg.ListID)
 	req, err := http.NewRequest("POST", createURL, bytes.NewBuffer(jsonBody))
 	if err != nil {
@@ -177,41 +171,34 @@ func (provider *AlertProvider) CreateTask(cfg *Config, body map[string]interface
 	}
 	req.Header.Set("Content-Type", "application/json")
 	req.Header.Set("Authorization", cfg.Token)
-
 	httpClient := client.GetHTTPClient(nil)
 	resp, err := httpClient.Do(req)
 	if err != nil {
 		return err
 	}
 	defer resp.Body.Close()
-
 	if resp.StatusCode >= 400 {
 		return fmt.Errorf("failed to create task, status: %d", resp.StatusCode)
 	}
-
 	return nil
 }
 
 func (provider *AlertProvider) CloseTask(cfg *Config, ep *endpoint.Endpoint) error {
 	fetchURL := fmt.Sprintf("%s/list/%s/task?include_closed=false", cfg.APIURL, cfg.ListID)
-
 	req, err := http.NewRequest("GET", fetchURL, nil)
 	if err != nil {
 		return err
 	}
 	req.Header.Set("Authorization", cfg.Token)
-
 	httpClient := client.GetHTTPClient(nil)
 	resp, err := httpClient.Do(req)
 	if err != nil {
 		return err
 	}
 	defer resp.Body.Close()
-
 	if resp.StatusCode >= 400 {
 		return fmt.Errorf("failed to fetch tasks, status: %d", resp.StatusCode)
 	}
-
 	var fetchResponse struct {
 		Tasks []struct {
 			ID   string `json:"id"`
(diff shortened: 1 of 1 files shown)


alerting/provider/clickup/clickup.go
```diff
@@ -0,0 +1,246 @@
...
+
+	if resp.StatusCode >= 400 {
+		return fmt.Errorf("failed to create task, status: %d", resp.StatusCode)
+	}
+
+	return nil
+}
+
+func (provider *AlertProvider) CloseTask(cfg *Config, ep *endpoint.Endpoint) error {
+	fetchURL := fmt.Sprintf("https://api.clickup.com/api/v2/list/%s/task?include_closed=false", cfg.ListID)
```
PythonGermany: If I understand correctly `cfg.APIURL` should be used here instead of the default one.

alerting/provider/clickup/clickup.go
```diff
@@ -0,0 +1,246 @@
...
+		if err := provider.UpdateTaskStatus(cfg, taskID, "closed"); err != nil {
+			return fmt.Errorf("failed to close task %s: %v", taskID, err)
+		}
+	}
+
+	return nil
+}
+
+func (provider *AlertProvider) UpdateTaskStatus(cfg *Config, taskID, status string) error {
+	updateURL := fmt.Sprintf("https://api.clickup.com/api/v2/task/%s", taskID)
```
PythonGermany: `cfg.APIURL` here as well.

alerting/provider/clickup/clickup.go
```diff
@@ -0,0 +1,246 @@
...
+
+	name := cfg.Name
+	if name == "" {
+		name = "Health Check: [ENDPOINT_GROUP]:[ENDPOINT_NAME]"
+	}
+
+	markdownContent := cfg.MarkdownContent
+	if markdownContent == "" {
+		markdownContent = "Triggered: [ENDPOINT_GROUP] - [ENDPOINT_NAME] - [ALERT_DESCRIPTION] - [RESULT_ERRORS]"
+	}
```
PythonGermany: I think setting those defaults should be moved to the `Valiate` method.

README.md
```diff
@@ -893,6 +895,64 @@ If the `access-key-id` and `secret-access-key` are not defined Gatus will fall b
 Make sure you have the ability to use `ses:SendEmail`.
 
 
+#### Configuring ClickUp alerts
+| Parameter                            | Description                                                                                | Default       |
+|:-------------------------------------|:-------------------------------------------------------------------------------------------|:--------------|
+| `alerting.clickup`                   | Configuration for alerts of type `clickup`                                                 | `{}`          |
+| `alerting.clickup.list-id`           | ClickUp List ID where tasks will be created                                                | Required `""` |
+| `alerting.clickup.token`             | ClickUp API token                                                                          | Required `""` |
+| `alerting.clickup.api-url`           | Custom API URL (optional, defaults to `https://api.clickup.com/api/v2/list/{list-id}/task`) | `""`          |
```
PythonGermany: Default value should be moved to default column.

README.md
```diff
@@ -893,6 +895,64 @@ If the `access-key-id` and `secret-access-key` are not defined Gatus will fall b
...
+| `alerting.clickup.assignees`         | List of user IDs to assign tasks to                                                        | `[]`          |
+| `alerting.clickup.status`            | Initial status for created tasks                                                           | `""`          |
+| `alerting.clickup.priority`          | Priority level for created tasks (1-4)                                                     | `0`           |
+| `alerting.clickup.name`              | Custom task name template (supports placeholders)                                          | `""`          |
+| `alerting.clickup.content`           | Custom task content template (supports placeholders)                                       | `""`          |
```
PythonGermany: Documented default values do not match default values in code.

alerting/provider/clickup/clickup.go
TheBinaryGuy: Sorry about that, updated, `APIURL` is now more generic (just the base) and other parts are now using it to construct the full URL.

alerting/provider/clickup/clickup.go
TheBinaryGuy: Fixed.

alerting/provider/clickup/clickup.go
TheBinaryGuy: Updated.

README.md
TheBinaryGuy: Fixed.

README.md
TheBinaryGuy: Fixed.

PythonGermany:
This provider does not support global overriding of config values based on the group the endpoint is in. I've not noticed most other alert providers (not all though) support it. @TwiN Is it a requirement that new providers support this?

TheBinaryGuy:
@PythonGermany Should I port it from the [Datadog Provider](https://github.com/TwiN/gatus/blob/40b1576ec788107d783b51532f5389fbdb52ecc4/alerting/provider/datadog/datadog.go#L56)?

PythonGermany:
If it's not too much trouble. If you want to be sure I'd wait for feedback from TwiN as I am not sure why some providers implement it and some not.

The changes required in the provider `GetConfig` method seem to be straightforward. To me it looks like many providers reimplement the same `GetConfig` method, I added the same implementation as to the alert provider I'm currently working on.

I wonder why this functionality is not implemented somewhere once like a template and then reused for every provider.

TheBinaryGuy:
@PythonGermany Not a big deal, just lifted it from the datadog provider. Let me know if it requires any changes! If you decide to go the template route, I'd be happy to change the provider to match that.

alerting/provider/clickup/clickup.go
PythonGermany: Is priority '0' a possible valid input as override priority in ClickUp alerts? If yes, a the config priority member could be a bool pointer and then check for nil to see if it is set in the override config.

alerting/provider/clickup/clickup.go
```diff
@@ -0,0 +1,282 @@
...
+	DefaultConfig Config `yaml:",inline"`
+
+	// DefaultAlert is the default alert configuration to use for endpoints with an alert of the appropriate type
+	DefaultAlert *alert.Alert `yaml:"default-alert,omitempty"`
+
+	// Overrides is a list of Override that may be prioritized over the default configuration
+	Overrides []Override `yaml:"overrides,omitempty"`
+}
+
+// Override is a case under which the default integration is overridden
```
PythonGermany:
```suggestion
// Override is a case under which the default configuration is overridden
```
I'm not sure what `integration` is.

TheBinaryGuy: Ahh just got it from the datadog provider

alerting/provider/clickup/clickup.go
TheBinaryGuy: No, priority has to be between 1 to 4, I have added validation for it now as well in the `Validate` function

PythonGermany: Wouldn't priority 3 be a reasonable default then if no manual priority was set in the config or override config? What happens with the task priority if it is not set and the POST request body for to create the task has value '0' set as priority?

TheBinaryGuy: Yeah I think you are right, setting `0` / `nil` both are creating a task with no priority assigned (we are using `Urgent` in our deployment, but I think `Normal` is a better default for others).

Should I make the default `3` then?

Also while `0` is working, the ClickUp docs suggests to just omit it, it might not work in the future, so should I also add a bool pointer as you suggested for folks who want to deliberately set `No Priority`.

What I am thinking is `no-priority` config option, if set, `priority` won't be validated or sent to ClickUp, if not set, provided `priority` (or default = `3`) will be validated and used.

PythonGermany: > What I am thinking is no-priority config option, if set, priority won't be validated or sent to ClickUp

Good point, I did not think about the scenario where no priority might be wanted instead of the default. An alternative to adding another option (where we would have to explain in the docs how it impacts the other option) I would suggest to allow `0` as input for `priority` and just add the priority config value to the create task request body if its not `nil` and not `0`.

> so should I also add a bool pointer as you suggested for folks who want to deliberately set No Priority

> Should I make the default 3 then?

Yes, making it a bool pointer and default priority `Normal` sounds good.

Do the people who might configure ClickUp as an alert provider usually know which priority number corresponds to which priority name? 

If not it might make improve usability to set the priority option as a string and map the string values to the right number before the API call, with an additional `None` value input allowed which maps to `null` in the API call.

Sorry about all the questions. It's interesting how the devil's always in the details :sweat_smile:

TheBinaryGuy: I don't think we can assume they will know, so I was adding the description in the README, but I like your idea of a map a lot more, better ux + also supports the None scenario well.

TheBinaryGuy: Done, I have also added instructions on how to get assignee ids and made `notify-all` configurable (defaults to `true`)

TwiN:
GitHub mobile is being silly and won't let me submit a code review, but please remove the overrides from the example. I try to keep the examples as short as possible since the documentation is already quite lengthy .

TheBinaryGuy:
@TwiN Done

TwiN:
GitHub mobile is still being silly, but please remove the unnecessary empty newlines inside functions, and then I think we're good to merge!

TheBinaryGuy:
@TwiN No worries, removed

TwiN:
@TheBinaryGuy Thank you for the contribution, and great work!

@PythonGermany Thank you for the assist!

TheBinaryGuy:
Thank you @PythonGermany and @TwiN and thanks for maintaining this amazing project!

PR #1463, opened 2025-12-21
feat(altering): Ntfy click link placeholder support

PythonGermany:
## Summary

Simple addition to allow placeholders in the click link. This way it's easy to add a generic click link that links to the correct gatus endpoint details view:
```yaml
alerting:
  ntfy:
    url: "http://ntfy"
    topic: "gatus-alerts"
    click: "https://${GATUS_DOMAIN}/endpoints/[ENDPOINT_KEY]"
```

## Checklist

### DIFF
--- a/alerting/provider/ntfy/ntfy.go
+++ b/alerting/provider/ntfy/ntfy.go
@@ -171,6 +171,17 @@ type Body struct {
 	Click    string   `json:"click,omitempty"`
 }
 
+func resolveEndpointPlaceholders(data string, ep *endpoint.Endpoint) string {
+	placeholders := map[string]string{
+		"[ENDPOINT_URL]": ep.URL,
+		"[ENDPOINT_KEY]": ep.Key(),
+	}
+	for placeholder, value := range placeholders {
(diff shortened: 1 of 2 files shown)


alerting/provider/ntfy/ntfy.go
```diff
@@ -171,6 +171,17 @@ type Body struct {
...
+func resolveEndpointPlaceholders(data string, ep *endpoint.Endpoint) string {
+	placeholders := map[string]string{
+		"[ENDPOINT_URL]": ep.URL,
+		"[ENDPOINT_KEY]": ep.Key(),
+	}
+	for placeholder, value := range placeholders {
+		data = strings.ReplaceAll(data, placeholder, value)
+	}
+	return data
+}
```
PythonGermany, line 183: I'm not sure if there is a need or interest for more supported placeholders for this and other alert providers. In case there is it would probably make sense to move this out to a file that is shared by all providers and implement it in a more generic way.

This is especially the case if, like in the case of the click url. The part of the input data that is not replaced is static. It does not really make sense to search and replace the same placeholders each time an alert is sent.

PR #1467, opened 2025-12-22
docs: Add missing alert provider group override options

PythonGermany:
## Summary

The following providers seem to support group overrides but have the documentation for its config keys missing:
https://github.com/TwiN/gatus/blob/40b1576ec788107d783b51532f5389fbdb52ecc4/alerting/provider/awsses/awsses.go#L189-L197
https://github.com/TwiN/gatus/blob/40b1576ec788107d783b51532f5389fbdb52ecc4/alerting/provider/custom/custom.go#L173-L181

## Checklist


TwiN:
Good catch

PR #1468, opened 2025-12-22
feat(alerting): Add support for Threema Gateway as an alerting provider (basic mode)

### LINKED ISSUE #1464: feat(alerting): Add basic support for Threema Gateway as an alerting provider

Add implementation for [Threema Gateway](https://threema.com/en/products/work/gateway) as an alerting provider (only basic mode)

The following send modes should be covered:
- Basic mode (no end to end encryption)

Resources: [API documentation](https://gateway.threema.ch/en/developer/api)

End-to-end encrypted mode will be covered in this followup ticket:
- #1470

_No response_

_No response_

### Additional information

_No response_

PythonGermany:
## Summary

Closes #1464

## Checklist

- [x] Check for outstanding TODO#1464 or TODO#1468 in code


alerting/provider/threemagateway/threemagateway.go
```diff
@@ -0,0 +1,363 @@
...
+	default:
+		r.Type = defaultRecipientType
+		r.Value = parts[0]
+	}
+	return nil
+}
+
+func (r Recipient) MarshalText() ([]byte, error) {
+	return []byte(r.Value), nil
+}
```
PythonGermany: I have not seen a solution to preprocess/parse string values from the yaml config like this anywhere else in the repo. This new provider was a good playground to test it out.

I think applied to the condition system it could improve code quality (and maybe performance, I have not enough experience with go to know how much) significantly, because then parsing of the condition and runtime evaluation would be completely separated.

@TwiN What do you think about my approach?

PythonGermany:
I have requested a Gateway ID from Threema. Once I have a working Gateway I will be able to test the alert provider against their API and once the tests are successful this PR will be ready.

PythonGermany:
> I have requested a Gateway ID from Threema. Once I have a working Gateway I will be able to test the alert provider against their API and once the tests are successful this PR will be ready.
Screenshot_20251228-184437

I was able to test the alert provider in basic mode with the official Gateway API. Sending the alert via basic mode works as expected.

TwiN (changes requested):
Instead of splitting the code across many files, please use the same approach as the other providers (one file for the implementation of the provider, one file for testing the implementation).

TwiN:
```
        ErrInvalidThreemaId          = errors.New("Must be 8 characters long and alphabetic characters must be uppercase")
```

Errors in Go shouldn't have the first character capitalized, can you fix that?

alerting/provider/threemagateway/threemagateway.go
```diff
@@ -23,26 +23,26 @@ const (
...
 	validRecipientTypes       = map[string]RecipientType{
 		"id":    RecipientTypeId,
 		"phone": RecipientTypePhone,
 		"email": RecipientTypeEmail,
 	}
 
-	ErrInvalidThreemaId          = errors.New("Must be 8 characters long and alphabetic characters must be uppercase")
-	ErrInvalidPhoneNumberFormat  = errors.New("invalid phone number: must contain only digits and may start with '+'")
-	ErrInvalidEmailAddressFormat = errors.New("invalid email address: must contain '@'")
+	errInvalidThreemaId          = errors.New("invalid id: must be 8 characters long and alphabetic characters must be uppercase")
```
TwiN: In Go, acronyms and abbreviations must be uppercased unless there are multiple back-to-back uppercased words in the field/variable name, in which case either all back-to-back acronym/abbreviations are `camelCase`, or all but the last one are `camelCase` and the last may be all uppercase.

For instance, in this case, it should be `errInvalidThreemaID`, but let's say it was an API ID, it could be either `errInvalidThreemaApiId` or `errInvalidThreemaApiID`.

The only exception is when the first word in an exported field is an abbreviation, e.g. APIID would be valid (though some might use `ApiID` or `ApiId` instead).

TLDR, this should be `errInvalidThreemaID`

alerting/provider/threemagateway/threemagateway.go
```diff
@@ -23,26 +23,26 @@ const (
 )
 
 var (
-	ErrApiIdentityMissing   = errors.New("api-identity is required")
-	ErrApiAuthSecretMissing = errors.New("auth-secret is required")
-	ErrRecipientsMissing    = errors.New("at least one recipient is required")
+	errApiIdentityMissing   = errors.New("api-identity is required")
+	errApiAuthSecretMissing = errors.New("auth-secret is required")
```
TwiN: These two should use `API` instead of `Api`

TwiN (changes requested):
Couple small changes, but we should be good to go afterwards

alerting/provider/threemagateway/threemagateway.go
PythonGermany: Where does this style requirement come from? The google styleguide [here](https://google.github.io/styleguide/go/decisions#initialisms)?

PR #1483, opened 2026-01-11
feat: add protection to metrics with security middleware

### LINKED ISSUE #1482: `/metrics` is publicly accessible even when `security.basic` / `security.oidc` is configured
When configuring Gatus with `security.basic` or `security.oidc`, the UI/API endpoints are protected as expected, but the Prometheus metrics endpoint (`GET /metrics`) remains accessible without authentication.

I’m not sure if this is intentional (to simplify Prometheus scraping), but it was unexpected and can lead to information disclosure (monitored services inventory + status).

Is `/metrics` intentionally excluded from auth middleware? If yes, could we document it clearly / add an option to not exclude it?

### Steps to reproduce

1. Run Gatus with metrics enabled and Basic auth enabled:
```yaml
metrics: true

security:
    basic:
        username: "john.doe"
        password-bcrypt-base64: "<bcrypt-base64>"

endpoints:
    - name: example
    url: "https://example.org"
    interval: 1m
    conditions:
        - "[STATUS] == 200"
```

2. Verify UI (or API) is protected:

```bash
curl -i http://localhost:8080/
# => 401 or auth challenge (expected)
```
3. Fetch metrics without auth:

```bash
curl -i http://localhost:8080/metrics
# => 200 and metrics payload (unexpected)
```

I also observed the same behavior with `security.oidc` enabled: `/` is protected but `/metrics` is still reachable without auth.

### Expected behavior

One of the following (depending on intended design):

* `/metrics` should be protected by the configured security mechanism (same as UI/API), OR
* The documentation should clearly state that `/metrics` is intentionally not protected and must be secured at the network / reverse-proxy level, OR
* Provide a configuration option to protect `/metrics` (e.g. `metrics.require-auth: true` and/or dedicated basic auth / bearer token for metrics scraping).

### Actual behavior

- `GET /metrics` returns metrics without any authentication, even when `security.basic` or `security.oidc` is enabled.

### Why this matters (impact)

The `/metrics` payload can reveal:

* monitored endpoint names/groups/keys (service inventory),
* status/HTTP codes, success/failure counters, latency,
* certificate expiration, etc.
* Go/process runtime metrics.

In our case, we run an internet-facing Gatus instance protected with OIDC.
Even though the UI/API is gated behind OIDC, the unauthenticated `/metrics` endpoint effectively leaks parts of our internal configuration and service inventory through metric labels (group/name/key). This is problematic for us and we’d like to avoid exposing this information publicly.

### Suggested approaches

* Apply the existing auth middleware to `/metrics` when `security.basic` is enabled (Prometheus supports basic auth), and document limitations for OIDC (interactive auth not suitable for Prometheus).
* Alternatively, add a dedicated “metrics auth” option (basic auth or bearer token), leaving the main UI as OIDC if desired.
* Or, if unauthenticated `/metrics` is intentional, add an explicit note in the docs + recommend reverse-proxy/IP allowlisting.

### Environment

* Version: c93a321 (it seems like this behavior exists at least as far back as the Fiber migration @ 6bb65f4)
* Deployment: docker
* Config: `security.basic` and/or `security.oidc`, `metrics: true`

---

I'd be more than happy to contribute the fix (or a PR adding a `metrics.require-auth` / `metrics.auth` option), if you'd be open to it, just point me to the preferred approach and any conventions you’d like me to follow!

ggmolly:
## Summary

adds optional protection for `/metrics` when security is configured, and rejects configs that enable it without auth.

- optionally protect `/metrics` with existing security middleware (`basic` or `oidc`) via `security.protect-metrics: true`
- validate protect-metrics requires a security provider
- cover protected metrics behavior in api and config tests
- document the protect-metrics setting with an example

closes #1482 

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -2704,9 +2704,10 @@ endpoints:
 ### Security
 | Parameter        | Description                  | Default |
 |:-----------------|:-----------------------------|:--------|
-| `security`       | Security configuration       | `{}`    |
-| `security.basic` | HTTP Basic configuration     | `{}`    |
-| `security.oidc`  | OpenID Connect configuration | `{}`    |
+| `security`                 | Security configuration                                        | `{}`    |
+| `security.protect-metrics` | Whether to protect `/metrics` with configured security      | `false` |
+| `security.basic`           | HTTP Basic configuration                                     | `{}`    |
+| `security.oidc`            | OpenID Connect configuration                                 | `{}`    |
 
 
 #### Basic Authentication
@@ -2773,6 +2774,17 @@ web:
 ### Metrics
 To enable metrics, you must set `metrics` to `true`. Doing so will expose Prometheus-friendly metrics at the `/metrics`
 endpoint on the same port your application is configured to run on (`web.port`).
+
+By default, `/metrics` is unauthenticated. To protect it using your configured security provider, set `security.protect-metrics`
+to `true` and configure `security.basic` or `security.oidc`.
+
+```yaml
+security:
+  protect-metrics: true
+  basic:
+    username: "john.doe"
+    password-bcrypt-base64: "JDJhJDEwJHRiMnRFakxWazZLdXBzRERQazB1TE8vckRLY05Yb1hSdnoxWU0yQ1FaYXZRSW1McmladDYu"
+```
 
 | Metric name                                  | Type    | Description                                                                | Labels                          | Relevant endpoint types |
 |:---------------------------------------------|:--------|:---------------------------------------------------------------------------|:--------------------------------|:------------------------|


api/api.go
```diff
@@ -68,7 +68,14 @@ func (a *API) createRouter(cfg *config.Config) *fiber.App {
...
 		}))
-		app.Get("/metrics", adaptor.HTTPHandler(metricsHandler))
+		metricsRouter := app.Group("/metrics")
+		if cfg.Security != nil && cfg.Security.ProtectMetrics {
+			if err := cfg.Security.ApplySecurityMiddleware(metricsRouter); err != nil {
+				panic(err)
+			}
+		}
+		metricsRouter.Get("", adaptor.HTTPHandler(metricsHandler))
+		metricsRouter.Get("/", adaptor.HTTPHandler(metricsHandler))
```
PythonGermany, line 78:
```suggestion
```
Not needed to keep previous routing behaviour since strict routing is not enabled: https://github.com/gofiber/fiber/blob/8b5d061088ca47b2fe801d14e33b95b7d445a54e/app.go#L121-L125 so I think it could be removed as far as I understand the routing behaviour of fiber.

PR #1486, opened 2026-01-12
feat(conditions,ui): Add endpoints[].ui.resolve-successful-conditions

lorenzoppx:
I’ve been using Gatus in my organization for about six months, and it has been an excellent experience. Recently, I identified a use case where it is necessary to display each condition value that was successfully met when an endpoint passes. This is especially helpful for users who need to monitor and better understand the behavior of an endpoint.

**Before:**
Sem título1

**After:**
Sem título
##Checklist

- [X] Added a test for the new functionality: TestEndpoint_ResolveSuccessfulConditions.
- [X] Verified that test results are identical when running both the master branch and this branch.
- [X] Updated the documentation in README.md, specifically in the UI subsection, to demonstrate the newly added option.


TwiN:
Makes sense to me!

Issue #1393, opened 2025-11-18
fallback notification


Xyz00777:
### Describe the feature request

Hi, it would be really nice to setup a fallback notification service like, first send per notify but when no response is coming/403 from the first backend, than send per telegram or something like that?

### Why do you personally want this feature to be implemented?

a fallback notification option would be really nice

### How long have you been using this project?

~10 minutes but i dont find it in the documentation

Christoph922:
I have exactly the same usecase, upvote!

PR #1496, opened 2026-01-18
Implement alert failover mechanism for notification/Alert providers

Xyz00777:
## Disclaimer
Its created with Claude Sonnet and Gemini 3 Pro.
I have absolutely 0 knowledge about Go so i used AI. But it tested it and as far as i can see and tested it works as expected.

## Summary
This PR introduces a simple notification failover mechanism for alerts. If the primary alerting provider fails to send a notification (e.g., due to an API outage or network error), Gatus will now automatically attempt to deliver the alert through a list of backup providers in the order they are configured.

## Checklist
- [x] **Core Logic**: Implemented `sendWithFailover` in the watchdog to handle sequential delivery attempts.
- [x] **Config Schema**: Added an optional `failover` field (list of provider types) to the `Alert` struct.
- [x] **Stability**: Added safety checks for `nil` alerting configurations to prevent runtime panics.
- [x] **Logging**: Added human-readable logs for failover events (e.g., `Trying failover provider 'telegram'`, `Alert sent successfully via failover...`).
- [x] **Testing**: Created `watchdog/alerting_failover_test.go` with 6 new test cases covering success, partial failure, total failure, and empty/unconfigured providers.
- [x] **Documentation**: Updated `README.md` with configuration parameters, behavior explanations, and a clear usage example.
- [X] **Tested**: Works on my machine ;D

### How it works
The failover is triggered only when the primary provider returns an error. It works for both **triggered** and **resolved** alerts. If a failover provider is not configured in the global `alerting` section, it is skipped with a warning log.

**Example Configuration:**
```yaml
alerting:
  pagerduty:
    integration-key: "********************************"
  telegram:
    token: "..."
    id: "..."

endpoints:
  - name: my-critical-app
    url: "https://example.org/health"
    alerts:
      - type: pagerduty
        failover: [telegram, slack]  # If PagerDuty fails, it tries Telegram, then Slack
        description: "Healthcheck failed"
```

Please test and check the Code. Im also not so sure if the current implementation is good like it is or if you have another idea @TwiN , how this should be implemented. Im open for ideas :)

I hope we can implement that feature because it would be really usefull.

fix https://github.com/TwiN/gatus/issues/1393

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,7 @@ Have any feedback or questions? [Create a discussion](https://github.com/TwiN/ga
     - [Configuring Zapier alerts](#configuring-zapier-alerts)
     - [Configuring Zulip alerts](#configuring-zulip-alerts)
     - [Configuring custom alerts](#configuring-custom-alerts)
+    - [Configuring alert failover](#configuring-alert-failover)
     - [Setting a default alert](#setting-a-default-alert)
   - [Maintenance](#maintenance)
   - [Security](#security)
@@ -803,6 +804,7 @@ Alerts are configured at the endpoint level like so:
 | `alerts[].send-on-resolved`          | Whether to send a notification once a triggered alert is marked as resolved.                                                                              | `false`       |
 | `alerts[].description`               | Description of the alert. Will be included in the alert sent.                                                                                             | `""`          |
 | `alerts[].provider-override`         | Alerting provider configuration override for the given alert type                                                                                         | `{}`          |
+| `alerts[].failover`                  | List of fallback provider types to try if the primary provider fails. <br />See [Configuring alert failover](#configuring-alert-failover).               | `[]`          |
 
 Here's an example of what an alert configuration might look like at the endpoint level:
 ```yaml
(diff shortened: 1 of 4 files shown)


Xyz00777:
@TwiN can i help somehow that the PR gets implemented? would be really nice to have it upstream

TwiN:
I apologize, but at the moment, it's not something I'm interested in implementing/supporting due to time constraints. This is because to implement it properly, we'd have to track when and which fallover(s) are used, otherwise, you might end up with Slack having the trigger alert and another app having the resolve alert, which is inconsistent. Supporting this would also require updating the persistence layer, which I don't have time to adequately test right now.

You can always set additional alerts with different thresholds, though I understand it's not quite what you're looking for.

That being said, it's a very good idea and feel free to create an issue and link to this PR for future reference.

Xyz00777:
Thanks for the response :) 
No problem I can understand but hope that it will get worked on in the not so distanced future :) 

Oh yes that's a good point, I didn't thought about that! Thank you

Yeah like you sayed that not what I'm looking for, currently I'm "solved" it for my worth the workaround that I have another monitor on my primary notification provider and when this one is down I get an alert, so I know at least that this one is down...

There is already an issue created ;) #1393

I'm looking forward to see how this will get in the future :)

Issue #1499, opened 2026-01-21
Suite name duplication should verify by using suite key instead


overskylab:
https://github.com/TwiN/gatus/blob/1c532971cbac4bc243c6a82fd641ab58b1199993/config/config.go#L512

Like an endpoint verification using Endpoint.Key to check duplicate name

From
```
// Check for duplicate suite names
if suiteNames[suite.Name] {
	return fmt.Errorf("duplicate suite name: %s", suite.Key())
}
```

To
```
// Check for duplicate suite names
if suiteNames[suite.Key] {
	return fmt.Errorf("duplicate suite name: %s", suite.Key())
}
```

Reproduce
- Add 2 suit with same name but difference group name

Result
- Program throws an error
```
panic: error parsing config: duplicate suite name: xxxx

goroutine 1 [running]:
main.main()
    /app/main.go:32 +0x1d4
```

PR #1500, opened 2026-01-21
fix(suite): Use key instead of name for deduplication

overskylab:
## Summary
Related issue #1499 

## Checklist


config/config.go
```diff
@@ -509,10 +509,10 @@ func ValidateSuitesConfig(config *Config) error {
 	suiteNames := make(map[string]bool)
 	for _, suite := range config.Suites {
 		// Check for duplicate suite names
-		if suiteNames[suite.Name] {
+		if suiteNames[suite.Key()] {
```
TwiN, line 512: Can you also rename the map name to suiteKeys?

PR #1506, opened 2026-01-23
fix: handle HTTP header names case-insensitively

### LINKED ISSUE #1491: HTTP header names should be case insensitive
### Describe the bug

where we configure headers like this:

```
endpoints:
  - headers:
      user-agent: Mozilla/5.0
```

due to how gatus is coded and maps in golang work, it's a toss up will we get mozilla or gatus header effectively applied to request, so results are random. Can we fix this so that the header names are truly case insensitive as they should be?

### What do you see?

endpoints fail intermittently due to wrong header sent

the header is sent as specified in configuration without another header of the same name being sent by gatus itself

1. use the headers config similar to above
2. observe on the server polled that headers received are mixed

### Version

5.34.0

### Additional information

_No response_

AnuragEkkati:
Details: HTTP header names should be case-insensitive per RFC 7230. When the users configure the headers like 'user-agent' (lowercase), gatus was adding its own 'User-Agent' header because the case-sensitive map lookup failed to find the existing header.

Fixed headers:
- User-Agent
- Content-Type
- Host

Fixes #1491

## Checklist

### DIFF
--- a/alerting/provider/gitea/gitea_test.go
+++ b/alerting/provider/gitea/gitea_test.go
@@ -12,6 +12,18 @@ import (
 	"github.com/TwiN/gatus/v5/test"
 )
 
+// isIgnorableTestError checks if an error is expected during testing when making API calls with dummy credentials
+func isIgnorableTestError(err error) bool {
+	if err == nil {
+		return false
+	}
+	errStr := err.Error()
+	return strings.Contains(errStr, "user does not exist") ||
+		strings.Contains(errStr, "no such host") ||
+		strings.Contains(errStr, "invalid username, password or token") ||
+		strings.Contains(errStr, "dial tcp")
+}
+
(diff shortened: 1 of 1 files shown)


config/endpoint/endpoint.go
```diff
@@ -189,6 +189,16 @@ func (e *Endpoint) Type() Type {
 	}
 }
 
+// hasHeader checks if a header exists in the map using a case-insensitive lookup
+func hasHeader(headers map[string]string, name string) bool {
```
TwiN: Can you move this function at the bottom of the file?

AnuragEkkati: Sure. Will update it now. 

Thanks for the review

AnuragEkkati: Done.

TwiN:

PR #1511, opened 2026-01-27
config: suites: use suite group and name to generate key

joy4eg:
This allows us to have more generic configs:
```yaml
suites:
  - name: node1-1
    group: XXX
    <<: *default-suite-config
    endpoints:
      - name: HTTP check
        url: https://example.com
        <<: *default-http

      - name: Write check
        url: wss://example.com
        <<: *default-write

      - name: Read check
        url: wss://example.com
        <<: *default-read
        
  - name: node1-2
    group: XXX
    <<: *default-suite-config
    endpoints:
      - name: HTTP check
        url: https://foo.bar
        <<: *default-http

      - name: Write check
        url: wss://foo.bar
        <<: *default-write

      - name: Read check
        url: wss://foo.bar
        <<: *default-read
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -561,9 +561,9 @@ func ValidateUniqueKeys(config *Config) error {
 			return fmt.Errorf("duplicate key '%s': suite '%s' conflicts with %s", suiteKey, suite.Key(), existing)
 		}
 		keyMap[suiteKey] = fmt.Sprintf("suite '%s'", suite.Key())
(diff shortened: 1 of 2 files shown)


joy4eg:
@TwiN it would be nice to get your thoughts on that, as copilot's comment is a bit confusing.

Goal: ability to use the same names across the suite.

joy4eg:
Moved to discussion - https://github.com/TwiN/gatus/discussions/1513

Issue #1237, opened 2025-09-01
Schedule heartbeat checks after startup sooner than the interval


aristaeus:
### Describe the feature request

Currently after restarting gatus, external endpoint heartbeats are checked only after the interval has passed. I would like some way to force it to check heartbeats sooner, e.g. when the time since the last push equals the heartbeat, or at a fixed time of day.

### Why do you personally want this feature to be implemented?

I have some low frequency heartbeats being monitored by gatus (daily backup scripts pushing that they ran successfully). This works well - except if gatus is restarted (i.e. I keep screwing with the config) it resets the timer and the heartbeats won't be checked for another day.

### Additional information

This would probably cause problems for external endpoints with a short interval that miss their push while gatus is down.

perchbirdd:
I have an external-endpoint health check on a 25hr interval for daily backups. I also backup gatus every night for the historical data. My backup script always takes the service down for backup integrity. I tested the 25hr backup endpoint last night by disabling the backup script. Gatus has a health check on a 25hr interval that has not had a response in 30hrs but is still showing healthy. This is odd behavior to me and is not outlined in the documentation. If I hadn't found this issue I would have no idea of the cause.

PythonGermany:
> This would probably cause problems for external endpoints with a short interval that miss their push while gatus is down.

For shorter intervals the chance of gatus being down to become an issue when it is started again is higher but it can happen for any interval. Just depends on when and how long gatus is down. The advantage is that endpoints with shorter intervals being unhealthy due to this issue would be healthy again faster since the next push will happen sooner.

PythonGermany:
> I have some low frequency heartbeats being monitored by gatus (daily backup scripts pushing that they ran successfully)

It might make sense to be able to configure a push window for external endpoints like this instead of an interval similar to [how maintenance windows are configured](https://github.com/TwiN/gatus/tree/77cedb080db5c27f1620b68bbce3154bf423783f?tab=readme-ov-file#maintenance).

manuelgrabowski:
Huh, I think this explains the odd behavior I'm seeing. I also have a heartbeat check with a 24 hour interval (1445 minutes, 5 minute extra to account for hiccups). I ping it at 7 in the morning every day.

[...]

PR #1521, opened 2026-02-01
docs: Improve heartbeat interval description

PythonGermany:
## Summary
As a result of the issue https://github.com/TwiN/gatus/issues/1322 and discussion https://github.com/TwiN/gatus/discussions/1507 I would like to improve the description for the heartbeat interval config to better align with it's current behaviour.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,16 +325,16 @@ For instance:
 - You can monitor services that are not supported by Gatus
 - You can implement your own monitoring system while using Gatus as the dashboard
 
-| Parameter                                 | Description                                                                                                                       | Default        |
-|:------------------------------------------|:----------------------------------------------------------------------------------------------------------------------------------|:---------------|
-| `external-endpoints`                      | List of endpoints to monitor.                                                                                                     | `[]`           |
-| `external-endpoints[].enabled`            | Whether to monitor the endpoint.                                                                                                  | `true`         |
-| `external-endpoints[].name`               | Name of the endpoint. Can be anything.                                                                                            | Required `""`  |
-| `external-endpoints[].group`              | Group name. Used to group multiple endpoints together on the dashboard. <br />See [Endpoint groups](#endpoint-groups).            | `""`           |
-| `external-endpoints[].token`              | Bearer token required to push status to.                                                                                          | Required `""`  |
-| `external-endpoints[].alerts`             | List of all alerts for a given endpoint. <br />See [Alerting](#alerting).                                                         | `[]`           |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -325,16 +325,16 @@ For instance:
...
+| `external-endpoints[].name`               | Name of the endpoint. Can be anything.                                                                                          | Required `""`  |
+| `external-endpoints[].group`              | Group name. Used to group multiple endpoints together on the dashboard. <br />See [Endpoint groups](#endpoint-groups).          | `""`           |
+| `external-endpoints[].token`              | Bearer token required to push status to.                                                                                        | Required `""`  |
+| `external-endpoints[].alerts`             | List of all alerts for a given endpoint. <br />See [Alerting](#alerting).                                                       | `[]`           |
+| `external-endpoints[].heartbeat`          | Heartbeat configuration for monitoring when the external endpoint stops sending updates.                                        | `{}`           |
+| `external-endpoints[].heartbeat.interval` | Interval used for update checks. If no update is received within this interval, alerts will be triggered. Must be at least 10s. | `0` (disabled) |
```
TwiN: I'm not sure if this really makes it clearer. Rather than reusing the same parameter name (interval) to explain what it does, it might make more sense to describe it with something like `Duration after which...` or `Period of time between...` (doesn't have to be exactly that, just something more clear that basically doesn't say "an interval is an interval" 😆)

TwiN (changes requested):

README.md
PythonGermany: I get your point... but `Period of time between` is just a longer way to say interval in my opinion. Also it was described as interval before. What about `Duration between update checks`?

TwiN: I have no idea what "between update checks" means. Is it `since the last status was pushed`?

PythonGermany: I have made an attempt to improve it and also added a warning message block further explaining the behaviour of the interval when gatus restarts.

README.md
```diff
@@ -350,6 +350,8 @@ external-endpoints:
         send-on-resolved: true
 ```
 
+> ⚠️ When you restart gatus the heartbeat interval timer that triggers alerting resets. Because of this the option might not be what you are looking for when you want to reliably monitor the health of jobs that push status updates with very low frequency eg. daily or weekly.
```
TwiN:
```suggestion
```

Honestly, this is more of a bug than something that should be documented. Rather than specifying this, we should fix the bug

PythonGermany: Ok, that that means instead of this note we complete #1237 to fix it. I've put it on my list of issues I want to work on at some point :smiley:

PR #1528, opened 2026-02-03
fix: Set first interval based on last result time

### LINKED ISSUE #1237: Schedule heartbeat checks after startup sooner than the interval

PythonGermany:
## Summary

Closes #1237 by introducing a check after startup adjusting the time to wait for the first heartbeat check based on the time that has passed since the last received result

## Checklist

### DIFF
--- a/watchdog/external_endpoint.go
+++ b/watchdog/external_endpoint.go
@@ -8,11 +8,28 @@ import (
 	"github.com/TwiN/gatus/v5/config/endpoint"
 	"github.com/TwiN/gatus/v5/metrics"
 	"github.com/TwiN/gatus/v5/storage/store"
+	"github.com/TwiN/gatus/v5/storage/store/common/paging"
 	"github.com/TwiN/logr"
 )
 
 func monitorExternalEndpointHeartbeat(ee *endpoint.ExternalEndpoint, cfg *config.Config, extraLabels []string, ctx context.Context) {
-	ticker := time.NewTicker(ee.Heartbeat.Interval)
+	timeToNextCheck := ee.Heartbeat.Interval
+
+	lastStatus, err := store.Get().GetEndpointStatusByKey(ee.Key(), paging.NewEndpointStatusParams().WithResults(0, 1))
+	if err != nil {
+		logr.Errorf("[watchdog.monitorExternalEndpointHeartbeat] Failed to get last endpoint status: %s", err.Error())
+	} else if lastStatus != nil {
+		if results := lastStatus.Results; len(results) > 0 {
+			timeSinceLastResult := time.Since(results[0].Timestamp)
+			logr.Debugf("[watchdog.monitorExternalEndpointHeartbeat] Time since last result: '%s'", timeSinceLastResult)
+			if timeSinceLastResult < timeToNextCheck {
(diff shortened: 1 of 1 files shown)


watchdog/external_endpoint.go
```diff
@@ -21,6 +38,7 @@ func monitorExternalEndpointHeartbeat(ee *endpoint.ExternalEndpoint, cfg *config
 			return
 		case <-ticker.C:
 			executeExternalEndpointHeartbeat(ee, cfg, extraLabels)
+			ticker.Reset(ee.Heartbeat.Interval)
```
PythonGermany, line 41: Not sure if this is considered a "hack". I'd love to learn how to do this in a more clean looking way.

watchdog/external_endpoint.go
```diff
@@ -8,11 +8,28 @@ import (
...
 	"github.com/TwiN/gatus/v5/storage/store"
+	"github.com/TwiN/gatus/v5/storage/store/common/paging"
 	"github.com/TwiN/logr"
 )
 
 func monitorExternalEndpointHeartbeat(ee *endpoint.ExternalEndpoint, cfg *config.Config, extraLabels []string, ctx context.Context) {
-	ticker := time.NewTicker(ee.Heartbeat.Interval)
+	timeToNextCheck := ee.Heartbeat.Interval
+
+	lastStatus, err := store.Get().GetEndpointStatusByKey(ee.Key(), paging.NewEndpointStatusParams().WithResults(0, 1))
```
PythonGermany, line 18: Is this an appropriate way to do this here with this store method and results size 1?

watchdog/external_endpoint.go
```diff
@@ -8,11 +8,28 @@ import (
...
+	if err != nil {
+		logr.Errorf("[watchdog.monitorExternalEndpointHeartbeat] Failed to get last endpoint status: %s", err.Error())
+	} else if lastStatus != nil {
+		if results := lastStatus.Results; len(results) > 0 {
+			timeSinceLastResult := time.Since(results[0].Timestamp)
+			logr.Debugf("[watchdog.monitorExternalEndpointHeartbeat] Time since last result: '%s'", timeSinceLastResult)
+			if timeSinceLastResult < timeToNextCheck {
```
PythonGermany, line 25: One question still open for me is which way to handle what should happen if the time since the last result is longer than the configured heartbeat interval. Currently it will wait for the duration of the heartbeat interval before the first check.

Alternatively the behaviour could be that a failure is created immediately. However I suspect this could often lead to false positives, especially for very short intervals. 

I think I prefer the currently implemented behaviour. Just wanted to throw the different possible approaches out there!

PR #1531, opened 2026-02-04
fix(security): Do not allow invalid security configs

### LINKED ISSUE #1530: Invalid security config allowed during startup
As I discovered and first mentioned here: https://github.com/TwiN/gatus/pull/1350#discussion_r2756344229:

> If more than one auth method is configured and the config for at least one of them is valid the validation check result for all other config sections is disregarded.
> 
> In such cases Gatus will still start without any indication of an invalid config even though it should show an error and quit immediately.

PythonGermany:
## Summary

Closes #1530 

## Checklist

### DIFF
--- a/security/config.go
+++ b/security/config.go
@@ -28,7 +28,7 @@ type Config struct {
 
 // ValidateAndSetDefaults returns whether the security configuration is valid or not and sets default values.
 func (c *Config) ValidateAndSetDefaults() bool {
-	return (c.Basic != nil && c.Basic.isValid()) || (c.OIDC != nil && c.OIDC.ValidateAndSetDefaults())
+	return (c.Basic == nil || c.Basic.isValid()) && (c.OIDC == nil || c.OIDC.ValidateAndSetDefaults())
 }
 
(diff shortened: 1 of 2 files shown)


security/config.go
```diff
@@ -28,7 +28,7 @@ type Config struct {
 
 // ValidateAndSetDefaults returns whether the security configuration is valid or not and sets default values.
 func (c *Config) ValidateAndSetDefaults() bool {
-	return (c.Basic != nil && c.Basic.isValid()) || (c.OIDC != nil && c.OIDC.ValidateAndSetDefaults())
+	return (c.Basic == nil || c.Basic.isValid()) && (c.OIDC == nil || c.OIDC.ValidateAndSetDefaults())
```
TwiN, line 31: Can you add a comment? It's not super clear what we're trying to achieve here

TwiN (changes requested):

security/config.go
TwiN: Nevermind, makes sense

TwiN (approved):

PR #1542, opened 2026-02-09
feat(alerting): Add Alertmanager provider

day1118:
## Summary

Adds Alertmanager as an alerting provider, allowing Gatus to post alerts via the Alertmanager API v2.

Rather than reimplementing features like silencing, inhibition rules, grouping, and routing directly in Gatus, this lets users route alerts through Alertmanager and use its full feature set. Alertmanager can then forward to any of its supported receivers (Slack, PagerDuty, email, webhooks, etc.), giving users a single place to manage alert routing and suppression — especially useful alongside an existing Prometheus stack.

- Prometheus-compatible labels (alertname, instance, job, severity)
- Multiple URLs for HA deployments; Alertmanager handles deduplication
- Configurable extra-labels and extra-annotations
- Group-based and per-alert provider overrides
- Lowers minimum-reminder-interval from 5m to 1m to support  Alertmanager's default 5m resolve_timeout (warns if < 5m for other providers)

## Checklist

### DIFF
--- a/alerting/alert/alert.go
+++ b/alerting/alert/alert.go
@@ -16,7 +16,7 @@ var (
 	// ErrAlertWithInvalidDescription is the error with which Gatus will panic if an alert has an invalid character
 	ErrAlertWithInvalidDescription = errors.New("alert description must not have \" or \\")
 
-	ErrAlertWithInvalidMinimumReminderInterval = errors.New("minimum-reminder-interval must be either omitted or be at least 5m")
+	ErrAlertWithInvalidMinimumReminderInterval = errors.New("minimum-reminder-interval must be either omitted or be at least 1m")
 )
 
 // Alert is endpoint.Endpoint's alert configuration
@@ -80,8 +80,11 @@ func (alert *Alert) ValidateAndSetDefaults() error {
 	if alert.SuccessThreshold <= 0 {
 		alert.SuccessThreshold = 2
 	}
-	if alert.MinimumReminderInterval != 0 && alert.MinimumReminderInterval < 5*time.Minute {
+	if alert.MinimumReminderInterval != 0 && alert.MinimumReminderInterval < time.Minute {
 		return ErrAlertWithInvalidMinimumReminderInterval
+	}
+	if alert.MinimumReminderInterval > 0 && alert.MinimumReminderInterval < 5*time.Minute && alert.Type != TypeAlertmanager {
+		logr.Warnf("[alert.ValidateAndSetDefaults] minimum-reminder-interval is set to %s, which is below the recommended 5m and may cause alert spam for most providers", alert.MinimumReminderInterval)
 	}
 	if strings.ContainsAny(alert.GetDescription(), "\"\\") {
 		return ErrAlertWithInvalidDescription
--- a/alerting/alert/alert.go
+++ b/alerting/alert/alert.go
@@ -83,8 +83,8 @@ func (alert *Alert) ValidateAndSetDefaults() error {
 	if alert.MinimumReminderInterval != 0 && alert.MinimumReminderInterval < time.Minute {
 		return ErrAlertWithInvalidMinimumReminderInterval
 	}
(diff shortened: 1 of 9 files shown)


alerting/alert/alert.go
```diff
@@ -16,7 +16,7 @@ var (
 	// ErrAlertWithInvalidDescription is the error with which Gatus will panic if an alert has an invalid character
 	ErrAlertWithInvalidDescription = errors.New("alert description must not have \" or \\")
 
-	ErrAlertWithInvalidMinimumReminderInterval = errors.New("minimum-reminder-interval must be either omitted or be at least 5m")
```
TwiN, line 19: Please don't mix unrelated changes in the PR, this should be discussed separately

TwiN (changes requested):
Great work, just one change I'd rather discuss separately

alerting/alert/alert.go
day1118: Thanks for the feedback!

This change is related. Alertmanager's default `resolve_timeout` is 5m, so it expects to receive a repeat alert within that window or it auto-resolves the incident. With a 5m minimum repeat interval, alerts end up in a constant `fire` → `resolve` → `fire` cycle. See `resolve_timeout` in the
[Alertmanager config docs](https://prometheus.io/docs/alerting/latest/configuration/#file-layout-and-global-settings).

I can split it into a separate PR, but the alertmanager provider needs a sub-5m interval to function correctly. A few options:

1. Lower the global minimum to 1m with a warning if below 5m (simplest - what I've done)
2. Keep 5m for other providers, allow 1m only for alertmanager (Easy enough - I can update this PR)
3. Split it out and update the docs to require users to increase `resolve_timeout` on their Alertmanager instance to accommodate the 5m minimum

Happy to go with whichever you prefer.

TwiN: Ok, this makes sense.

day1118:
Would you like me to update the PR to address the comments from Copilot @TwiN?

TwiN:
@day1118 If you could address the last one, I'd appreciate it, because it does have a good point

day1118:
> @day1118 If you could address the last one, I'd appreciate it, because it does have a good point

Comments have been addressed @TwiN

alerting/alert/alert.go
```diff
@@ -80,9 +80,12 @@ func (alert *Alert) ValidateAndSetDefaults() error {
 	if alert.SuccessThreshold <= 0 {
 		alert.SuccessThreshold = 2
 	}
-	if alert.MinimumReminderInterval != 0 && alert.MinimumReminderInterval < 5*time.Minute {
+	if alert.MinimumReminderInterval != 0 && alert.MinimumReminderInterval < time.Minute {
 		return ErrAlertWithInvalidMinimumReminderInterval
 	}
+	if alert.MinimumReminderInterval > 0 && alert.MinimumReminderInterval < 5*time.Minute && alert.Type != TypeAlertmanager {
+		logr.Warnf("[alert.ValidateAndSetDefaults] minimum-reminder-interval is set to %s, which is below the recommended 5m and may cause alert spam for most providers", alert.MinimumReminderInterval)
+	}
```
TwiN: All provider-specific validation should be self-contained within the provider itself, or where the Config struct is validated (search for validateAlerting or something along those lines).

I understand why you'd put it here, but for the sake of consistency/maintainability and given that no other provider-specific validation happens here, we should move it elsewhere.

TwiN (changes requested):

alerting/alert/alert.go
day1118: Hi @TwiN - does this feel suitable? I didn't want to change all providers to perform a check themselves, so just set the min to 1m, and warn if it is below 5m.
Let me know what you think.

day1118:
@TwiN — anything else needed to get this merged?

zimiovid:
@TwiN any updates?

Issue #1172, opened 2025-07-24
Zero values in weekly and monthly charts


petrkutalek:
### Describe the bug

Weekly and monthly charts always start with a series of zero values. This is regardless of the metric; values are measured over many months. See the attached images.

<img width="1305" height="468" alt="Image" src="https://github.com/user-attachments/assets/9be697b9-8b4f-4b3b-9bd2-0ae9b779b90c" />
<img width="1321" height="478" alt="Image" src="https://github.com/user-attachments/assets/68647829-031e-4e65-8813-ce19e8314219" />

### Version

latest

PR #1560, opened 2026-02-26
fixes issue #1172 zero values in weekly and monthly charts

christophdb:
## Summary
fixes issue #1172 zero values in weekly and monthly charts

## Checklist


api/chart.go
```diff
@@ -144,7 +136,9 @@ func ResponseTimeHistory(c *fiber.Ctx) error {
 	if err != nil {
 		return c.Status(400).SendString("invalid key encoding")
 	}
-	hourlyAverageResponseTime, err := store.Get().GetHourlyAverageResponseTimeByKey(endpointKey, from, time.Now())
+	// Fetch one extra day of data before 'from' so the chart line extends to the left edge
+	// (the frontend clips the visible range to 'from' via the X-axis min/max)
+	hourlyAverageResponseTime, err := store.Get().GetHourlyAverageResponseTimeByKey(endpointKey, from.Add(-24*time.Hour), time.Now())
```
TwiN, line 141: This does not seem to be the right approach. We should be fixing this bug in the front end, not the back end

PR #1568, opened 2026-03-05
feat(jsonpath): add a real jsonpath parser

JannikBulow:
Library used is https://github.com/oliveagle/jsonpath I made sure to check the output of the library so tests would succeed. This technically means that it's not 100% correct JSONPath output.

## Summary

I decided to switch out the "half-baked json path implementation that needs some love" with an actual library as I needed a filter selector for my status page. I have done my best to keep the output the same as the older parser so no one will have to update their conditions.

I am not very good at writing tests or documentation so I simply haven't, but all the existing tests passed.

## Checklist


yurivangeffen:
We could really use this!

PR #1581, opened 2026-03-12
feat(alerting): Add configurable api-url for OpsGenie provider

meesns:
## Summary

The OpsGenie API URL is currently hardcoded to `https://api.opsgenie.com/v2/alerts`. This makes it impossible to use alternative API-compatible endpoints such as:

- **Atlassian JSM Operations API**: `https://api.atlassian.com/jsm/ops/integration/v2/alerts`
- **EU OpsGenie API**: `https://api.eu.opsgenie.com/v2/alerts`

Atlassian is [retiring standalone OpsGenie](https://www.itmethods.com/opsgenie-is-being-deprecated-how-to-prepare-for-migration/) (shutdown April 5, 2027) and migrating users to JSM Operations, which exposes the same alert API at a different base URL.

## Changes

Adds an optional `api-url` field to the OpsGenie `Config` struct, defaulting to the current URL for backwards compatibility:

```yaml
alerting:
  opsgenie:
    api-url: "https://api.atlassian.com/jsm/ops/integration/v2/alerts"
    api-key: "..."
```

## Prior art

Prometheus Alertmanager already supports this via a configurable `api_url` field, which enabled seamless JSM migration without code changes:

- [prometheus/alertmanager#4244](https://github.com/prometheus/alertmanager/issues/4244) — confirmed existing OpsGenie receiver works with JSM by changing the URL
- [prometheus/alertmanager#3949](https://github.com/prometheus/alertmanager/pull/3949) — confirmed API compatibility between OpsGenie and JSM

### DIFF
--- a/alerting/provider/opsgenie/opsgenie.go
+++ b/alerting/provider/opsgenie/opsgenie.go
@@ -17,7 +17,7 @@ import (
 )
 
 const (
-	restAPI = "https://api.opsgenie.com/v2/alerts"
+	defaultAPIURL = "https://api.opsgenie.com/v2/alerts"
 )
 
 var (
@@ -25,6 +25,11 @@ var (
 )
 
 type Config struct {
+	// APIURL to use for the OpsGenie API
+	//
+	// default: https://api.opsgenie.com/v2/alerts
+	APIURL string `yaml:"api-url"`
+
 	// APIKey to use for
 	APIKey string `yaml:"api-key"`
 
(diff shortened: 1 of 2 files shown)


HartmannVolker:
We would be highly interested in this feature as well! Maybe update the Readme as well so that people know that they can use the opsgenie integration for JSM as well.

TwiN:
Honestly, if OpsGenie is being removed, I would rather add a different provider for its replacement. Even if the API is compatible, once it's completely decommissioned, there's just no point in keeping the provider on Gatus, so may as well implement a separate provider for the replacement

PR #1583, opened 2026-03-16
feat(ui): Add customizable OIDC login screen branding

hluaces:
As per https://github.com/TwiN/gatus/discussions/1579, this PR allows for some customizations in the OIDC login screen:

- If a logo is set it will be displayed alongside the Gatus one
- New ui config `ui.login-subtitle`: customises the message in the screen
- If set, use `ui.header` will be used to customise the title

I haven't commited the web static assets. I believe that's triggered with a comment? let me know if I should be doing it. 

Also please clarify if I've forgotten about something here, this is my first pr.


hluaces:
This is how it looks like when a logo, header and login-subtitle is set:

Captura desde 2026-03-16 10-39-48

This is what it looks like with a login-subtitle and header, but no logo:

Captura desde 2026-03-16 10-41-05

And this is how it looks like with no header, logo or login-subtitle (matches existing default looks):

Captura desde 2026-03-16 10-50-44

TwiN:
Excellent work!

PR #1586, opened 2026-03-18
fix(alerting): Remove square brackets from email subject

### LINKED ISSUE #1563: Email alerts - threading doesn't work with Gmail due to square brackets in subject line
### Describe the bug

I've noticed some strange issues with threading of email alerts in Gmail. Specifically I see that different alerts get merged into a single email thread even though they have different subject lines. So for example I might have `Alert X` and `Alert Y` that both send emails - even though the subjects are different (`[Alert X] Alert triggered` vs. `[Alert Y] Alert triggered`) Gmail will merge those two emails into a single thread.

It looks like the reason for this is that Gmail ignores text in square brackets when creating threads, and since the alert name is in square brackets Gmail is just threading everything together since it's only matching on the "Alert triggered" text.

An easy fix would be to just remove the square brackets from the subject line here:

https://github.com/TwiN/gatus/blob/master/alerting/provider/email/email.go#L145-L151

However I realize that some people might prefer the square brackets, so another option could be to add a bool to the config that controls whether square brackets are used.

I have some Golang experience and I'd be happy to send a PR if you let me know how you'd like this to be implemented :)

Thanks so much for your work on this project!

### What do you see?

_No response_

_No response_

_No response_

### Version

5.35.0

### Additional information

_No response_

jaydeethree:
… resolve Gmail threading issues. Fixes #1563

## Summary

#1563

## Checklist

- [Not Applicable] Updated documentation in `README.md`, if applicable.

### DIFF
--- a/alerting/provider/email/email.go
+++ b/alerting/provider/email/email.go
@@ -143,10 +143,10 @@ func (provider *AlertProvider) Send(ep *endpoint.Endpoint, alert *alert.Alert, r
 func (provider *AlertProvider) buildMessageSubjectAndBody(ep *endpoint.Endpoint, alert *alert.Alert, result *endpoint.Result, resolved bool) (string, string) {
 	var subject, message string
 	if resolved {
-		subject = fmt.Sprintf("%s Alert resolved", ep.DisplayName())
+		subject = fmt.Sprintf("%s: Alert resolved", ep.DisplayName())
 		message = fmt.Sprintf("An alert for %s has been resolved after passing successfully %d time(s) in a row", ep.DisplayName(), alert.SuccessThreshold)
 	} else {
-		subject = fmt.Sprintf("%s Alert triggered", ep.DisplayName())
+		subject = fmt.Sprintf("%s: Alert triggered", ep.DisplayName())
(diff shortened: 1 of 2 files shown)


TwiN:
I think it'd make more sense to just remove the square brackets instead of swapping them for parentheses

jaydeethree:
> I think it'd make more sense to just remove the square brackets instead of swapping them for parentheses

Done.

alerting/provider/email/email.go
```diff
@@ -143,10 +143,10 @@ func (provider *AlertProvider) Send(ep *endpoint.Endpoint, alert *alert.Alert, r
 func (provider *AlertProvider) buildMessageSubjectAndBody(ep *endpoint.Endpoint, alert *alert.Alert, result *endpoint.Result, resolved bool) (string, string) {
 	var subject, message string
 	if resolved {
-		subject = fmt.Sprintf("[%s] Alert resolved", ep.DisplayName())
+		subject = fmt.Sprintf("%s Alert resolved", ep.DisplayName())
```
TwiN: Could you add `:` after `%s`? As in `%s: Alert resolved`

Sorry, I assumed there was already a colon after the brackets

jaydeethree: Done.

TwiN:

jaydeethree:
Sure thing, thank you for an awesome project!

PR #1597, opened 2026-03-24
docs: Update postgres volume bind mount

### LINKED ISSUE #1596: docs: Update the bind mount for the postgres docker example
In postgres 18 docker, the data directory has changed. The container will fail to start if the older /var/lib/postgresql/data is used.

[The new recommendation is to use /var/lib/postgresql.](https://hub.docker.com/_/postgres#pgdata)

The included docker compose example should reflect this change.

ClementPhan:
## Summary

This update the bind mount for the postgres docker compose example.

This will fix #1596

## Checklist

### DIFF
--- a/.examples/docker-compose-postgres-storage/compose.yaml
+++ b/.examples/docker-compose-postgres-storage/compose.yaml
@@ -2,7 +2,7 @@ services:
   postgres:
     image: postgres
     volumes:
-      - ./data/db:/var/lib/postgresql/data
+      - ./data/db:/var/lib/postgresql
     ports:
       - "5432:5432"
     environment:


amai2012:
If the data directory is version dependent, the example should use a proper tag for the postgres image.

ClementPhan:
I do not believe a version tag to be necessary. The example is already implicitly targeting the "latest" tag where the mount point has already changed and the example fails to launch the containers.

If a version was specified, it could indicate there is a specific version expected from Gatus. I have made this assumption in the past, for other projects.

Furthermore, while the data path is version dependent, the mount point is not. The Postgres container documentation shows that the database will keep the files in the new mount, even in future major versions.

There could be an issue if the user specified a version tag of < 18, but I do not believe this is in the scope of an example config. Please correct and edit the request if I am mistaken.

amai2012:
adding the image tag 18 would:

1. keep the example valid even if postgres 19 would change the data dir again
2. document a known combination to work

Actually the missing tag is not a weak point of your contribution, but on the example as it has been before.

PR #1603, opened 2026-03-28
fix: return error from rdapQuery when no expiration event found

### LINKED ISSUE #1570: domain_expiration function is jacked up
### Describe the bug

The domain within the ".net" TLD fails to properly check the domain registration date; however, it seems to work fine for ".com" TLDs.

Image

### What do you see?

_No response_

_No response_

_No response_

### Version

_No response_

### Additional information

_No response_

majiayu000:
Fixes #1570

`rdapQuery()` was silently returning a zero-value `ExpirationDate` when the RDAP
response didn't contain an expiration event. This meant `GetDomainExpiration` never
fell back to WHOIS for domains like .net where RDAP might not include expiration data.

Added a check after the event loop: if `ExpirationDate` is still zero, return an error
so the caller falls back to WHOIS as intended (client.go:73-74).

Also added a test case for a .net domain in `TestRdapQuery`.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -529,7 +529,7 @@ func rdapQuery(hostname string) (*whois.Response, error) {
 		}
 	}
 	if response.ExpirationDate.IsZero() {
-		return nil, errors.New("no expiration event found in RDAP response")
+		return nil, fmt.Errorf("no expiration event found in RDAP response for %s", hostname)
 	}
 	return &response, nil
 }
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -55,6 +55,13 @@ func TestRdapQuery(t *testing.T) {
 	} else if response.ExpirationDate.Unix() <= 0 {
 		t.Error("expected to have a valid expiry date, got", response.ExpirationDate.Unix())
 	}
+	// .net domain: rdapQuery must either return a valid expiration or an error,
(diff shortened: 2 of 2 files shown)


TwiN:

n3gwg:
I stand most greatly appreciative of everyone that looked into this issue.
Might I ask how I will know when this new code has made it into the next version of the container or what version number would represent this code fix?

The reason I am asking is because I just entirely deleted and downloaded whatever is the currently available version of the container which states it is 7 hours old and I see that the code was merged 7 hours ago. Thus, I presume the container I have is the latest code. If so, then I sadly must state that the bug seems to be remaining stubbornly persistent. However, if I simply must wait a bit longer for the latest code to merge into the container I understand entirely.

Stuart, N3GWG

TwiN:
It should be available under the latest tag, but you may still have it cached. FWIW there was a build issue yesterday that I just fixed, so maybe that was the issue

n3gwg:
I downloaded this image:
$ docker images
docker.io/twinproduction/gatus  latest      a405b2f3eaf3  36 hours ago  49.4 MB

It seems to still have the bug persistent. Am I using the proper image that has the fix in it or is this wrong?

Thanks for your hard work on this thus far!

Stuart, N3GWG

PR #1629, opened 2026-04-17
fix(config): walk the resolved config path, not the raw argument

### LINKED ISSUE #456: Config paths are not treated equally
### Describe the bug

Well, this is kinda weird so bear with me.

I've changed the hardcoded paths to config to a path to a directory and gatus failed to load config from there. Yes, I know that it is probably odd to report bugs after essentially manually breaking the code, but lemme finish. After little investigation I have noticed that the code probably uses the wrong variable to traverse a directory. In config package, in function `LoadConfiguration`, after the loop and call to `os.Stat()` the case when a path points to a directory `walkConfigDir` is called to walk in `configPath` provided to `LoadConfiguration` as an argument (which could be empty string), but probably should be called with `usedConfigPath`, which should be set to first non-empty valid path.

### What do you see?

Gatus fails to read config files from a directory if the hardcoded paths were paths to directories. I understand that this is probably not intended use of a project, but still.

Gatus successfully launches if config paths are set to correct existing locations.

1. Change the `DefaultConfigurationFilePath` and `DefaultFallbackConfigurationFilePath` to point to some directory with configs.
2. Recompile Gatus
3. Launch it, do not set `GATUS_CONFIG_PATH` environment variable.
4. It fails.

### Version

5.3.2

### Additional information

Probably related to #400.

SAY-5:
### Problem

`LoadConfiguration` chooses `usedConfigPath` by walking the raw
argument and then the two default paths (`DefaultConfigurationFilePath`,
`DefaultFallbackConfigurationFilePath`), keeping the first one that
stats cleanly.

When the resolved path turns out to be a directory it walks it, but the
call uses the function argument `configPath`, not the resolved
`usedConfigPath`:

```go
if fileInfo.IsDir() {
    err := walkConfigDir(configPath, func(path string, d fs.DirEntry, err error) error {
        ...
```

If `configPath` is empty (the common case, e.g. `GATUS_CONFIG_PATH`
unset) and one of the default paths happens to resolve to a directory,
`walkConfigDir` is handed `""`, descends from the current working
directory, and silently skips every config file. Gatus fails to load
anything, even though `os.Stat` on the default path succeeded. #456.

### Fix

Walk the path we already proved exists:

```diff
- err := walkConfigDir(configPath, func(path string, d fs.DirEntry, err error) error {
+ err := walkConfigDir(usedConfigPath, func(path string, d fs.DirEntry, err error) error {
```

No behaviour change when the caller explicitly passes a directory:
`configPath == usedConfigPath` in that case. Only the
"argument is empty, default resolves to a directory" path changes, and
that path is currently broken.

Fixes #456

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -218,7 +218,11 @@ func LoadConfiguration(configPath string) (*Config, error) {
 	}
 	var config *Config
 	if fileInfo.IsDir() {
-		err := walkConfigDir(configPath, func(path string, d fs.DirEntry, err error) error {
+		// Walk the directory we actually resolved, not the raw argument: when
+		// configPath is empty and the directory came from one of the default
+		// paths, walking configPath would descend from the current working
+		// directory and silently skip every file. See #456.
+		err := walkConfigDir(usedConfigPath, func(path string, d fs.DirEntry, err error) error {
 			if err != nil {
 				return fmt.Errorf("error walking path %s: %w", path, err)
 			}


wahajahmed010 (commented):
## Review: fix(config): walk the resolved config path, not the raw argument (#1629)

Nice catch, @SAY-5. This is the kind of subtle bug that's easy to miss. When `configPath` is empty and `usedConfigPath` resolves to a default path, walking the raw `configPath` would descend from CWD and silently skip everything — not a crash, just silently loading nothing.

**Verification:** The change is minimal and correct — `walkConfigDir(usedConfigPath, ...)` instead of `walkConfigDir(configPath, ...)` ensures you're walking the actual resolved directory. The comment referencing the relevant issue is helpful. +5/-1, clean and focused.

One minor suggestion: a test case with an empty config path + default directory resolution would make this regression-proof. But not a blocker.

**LGTM.**

PR #1644, opened 2026-04-26
fix: update lib/pq to v1.11.2 to fix Supabase PostgreSQL connection regression

### LINKED ISSUE #1633: PostgreSQL connection panic: EOF with Supabase (regression in v5.35.0)
### Describe the bug

Gatus v5.35.0 fails to connect to Supabase PostgreSQL databases with `panic: EOF` during initialization. This is a regression from v5.34.0 which works correctly with the same configuration. The issue prevents Gatus from starting when configured to use Supabase as the PostgreSQL storage backend.

### What do you see?

Gatus crashes immediately during startup with the following error:

```
panic: EOF
goroutine 1 [running]:
main.initializeStorage(0x...)
        /app/main.go:105 +0x10f0
main.main()
        /app/main.go:34 +0xb0
```

The container exits and Gatus never starts monitoring endpoints.

Gatus should successfully connect to Supabase PostgreSQL, initialize the storage, and start monitoring the configured endpoints. The application should remain running and display "✓ Gatus is ready!" message.

1. Configure Gatus storage with Supabase PostgreSQL:
   ```yaml
   storage:
     type: postgres
     path: ${DATABASE_URL}
   ```
2. Set environment variable with Supabase connection string:
   ```
   DATABASE_URL=postgres://postgres.[project-ref]:[password]@aws-[region].pooler.supabase.com:5432/postgres?sslmode=require
   ```
3. Run Gatus v5.35.0 (e.g., via Docker)
4. Observe the panic error during startup

### Version

- Gatus v5.35.0 (broken)
- Gatus v5.34.0 (working - last known good version)
- PostgreSQL: Supabase (PostgreSQL 17.6)

### Additional information

- Direct `psql` connection to the same Supabase database works fine
- The `lib/pq` driver works independently when tested with a simple Go program
- Both direct connection (port 5432) and pooler connection (port 6543) fail
- Tested SSL modes: `require`, `verify-full`, `verify-ca` - all fail with v5.35.0
- Intermediate versions v5.34.1-v5.34.5 don't exist as Docker tags
- Workaround: Use Gatus v5.34.0 instead of v5.35.0
- Supabase uses connection pooling (PgBouncer/Supavisor) with SSL enforcement enabled

wahajahmed010:
## Description

Fixes #1633

This PR updates `github.com/lib/pq` from v1.11.1 to v1.11.2 to resolve a PostgreSQL connection regression affecting Supabase users.

### Root Cause

`lib/pq` v1.11.0 introduced a Config struct refactor (commit c8c8c81) that unconditionally sends an empty `"options"` startup parameter during the PostgreSQL handshake. Supabase's Supavisor connection pooler rejects startup messages containing empty parameters, returning `:bad_startup_payload` and causing an `EOF` panic during Gatus initialization.

See upstream issue: https://github.com/lib/pq/issues/1259

### Fix

Bump `lib/pq` to v1.11.2 (commit 1412805), which fixes the issue by omitting empty startup parameters from the handshake.

### Changes

- `go.mod`: bump `github.com/lib/pq` v1.11.1 → v1.11.2
- `go.sum`: update checksums for `lib/pq` v1.11.2
- `storage/store/sql/supabase_test.go`: add regression test to verify lib/pq version requirement

### Verification

- [x] go.mod points to `github.com/lib/pq v1.11.2`
- [x] go.sum checksums updated from proxy.golang.org
- [x] Regression test verifies minimum lib/pq version at build time

### AI Disclosure

This contribution was AI-assisted. The issue was diagnosed by analyzing the `lib/pq` upstream bug report, bisecting the breaking commit, and applying the known fix.


TwiN:
The test you added doesn't make sense. Once versions are bumped up, there's very rarely any reason to bump it back down, rendering the test useless.

Please remove the test

wahajahmed010:
Hi TwiN, just following up on this PR. The test removal you requested has been addressed — the current diff only contains the lib/pq version bump in go.mod/go.sum with no test additions. Happy to make any further changes if needed. Thanks for your time.

TwiN (approved):

TwiN:
Thank you!

PR #1668, opened 2026-05-14
fix(config): use actual file modification time in UpdateLastFileModTime

### LINKED ISSUE #1657: Constant restart loop triggered by listenToConfigurationFileChanges preventing Slack alerts
### Describe the bug

I am experiencing a persistent restart loop in Gatus where the server shuts down and restarts every few minutes, citing configuration changes that have not occurred. This instability appears to be interfering with the alerting system, as Slack notifications are not being dispatched when endpoints fail.

### What do you see?

The logs show that Gatus detects a configuration change almost every minute:
`[main.listenToConfigurationFileChanges] Configuration file has been modified
`
This triggers a graceful shutdown:
`[controller.Handle] Server has shut down successfully
`
Because the server is constantly restarting and canceling current executions, the alerting logic for failing endpoints never seems to complete or fire the Slack webhook.

Relevant logs:

```
2026/05/07 12:46:33 [watchdog.executeEndpoint] Monitored group=core; endpoint=test; key=core_test; success=false; errors=0; duration=22ms; body={"status":"healthy", ...}
2026/05/07 12:46:33 [main.listenToConfigurationFileChanges] Configuration file has been modified
2026/05/07 12:46:33 [watchdog.monitorEndpoint] Canceling current execution of group=core; endpoint=test; key=core_test
2026/05/07 12:46:33 [controller.Handle] Server has shut down successfully
2026/05/07 12:46:34 [config.LoadConfiguration] Reading configuration from configFile=config/config.yaml
```

In the end: An alert in slack. But I guess this is due to the server restarts so I think the goal would be that the server does not restart unless there are real changes

1. Deploy Gatus to a Containerized Environment: Use a platform like Google Cloud Run or Kubernetes where the configuration file (config.yaml) is mounted via a volume or sidecar.
2. Configure an Endpoint to Fail: Create an endpoint with a condition that is guaranteed to fail, ensuring it is linked to a Slack alert.
```
endpoints:
  - name: reproduction-test
    url: "https://somethingwhichfails.comm"
    interval: 30s
    conditions:
      - "[BODY].test == ok"
    alerts:
      - type: slack
```
I have set this up in Google Secret Manager and Gatus itself is running in one instance (exactly one, no autoscalling) in Google Cloud Run.
3. Observe the Race Condition: Monitor the logs. Note that the moment the watcher detects a change, it initiates a Server has shut down successfully sequence.
4. Check for Alerts: Observe that because the watchdog.monitorEndpoint execution is canceled immediately after the failure is detected, the Slack notification is never dispatched.

### Version

v5.35.0

### Additional information

I guess the reason the Slack alerts aren't working is that Gatus is killing itself the exact same millisecond it realizes the test failed.
In my logs:
1. `14:46:33.243: core_test fails.`
2. `14:46:33.331: (Less than 100ms later) Gatus sees a "config change" and starts the shutdown.`
3. `14:46:33.332: It logs Canceling current execution. This kills the alert before it even leaves the server.`

Now the question is why this change is getting triggered.

RaminGe:
## Summary
Fixes #1657

Previously, `UpdateLastFileModTime()` set the last file modification time to `time.Now()`. In containerized environments with volume mounts (e.g. Kubernetes, Google Cloud Run), the mounted file's timestamp can be in the future relative to the container's internal clock. This caused the configuration watcher to continuously evaluate that the file had been modified (because `config.lastFileModTime` < `fileInfo.ModTime()`), triggering infinite restart loops and effectively disabling the alerting engine.

This PR updates `UpdateLastFileModTime()` to read the actual ModTime of the config file using `os.Stat` (and if it is a directory, using the newest ModTime inside that directory). It falls back to `time.Now()` only if the true modification time cannot be determined.

## Checklist

- [x] Tested and/or added tests to validate that the changes work as intended, if applicable. -> Tested locally
- [x] Updated documentation in `README.md`, if applicable. -> Not applicable


mgenoni:
@RaminGe 🐎🔥

config/config.go
```diff
@@ -191,7 +191,29 @@ func (config *Config) HasLoadedConfigurationBeenModified() bool {
 
 // UpdateLastFileModTime refreshes Config.lastFileModTime
 func (config *Config) UpdateLastFileModTime() {
-	config.lastFileModTime = time.Now()
+	fileInfo, err := os.Stat(config.configPath)
+	if err != nil {
+		config.lastFileModTime = time.Now()
+		return
+	}
+
```
TwiN:
```suggestion
```

config/config.go
```diff
@@ -191,7 +191,29 @@ func (config *Config) HasLoadedConfigurationBeenModified() bool {
...
+		_ = walkConfigDir(config.configPath, func(path string, d fs.DirEntry, err error) error {
+			if info, err := d.Info(); err == nil && info.ModTime().After(maxModTime) {
+				maxModTime = info.ModTime()
+			}
+			return nil
+		})
+	} else {
+		maxModTime = fileInfo.ModTime()
+	}
+
```
TwiN:
```suggestion
```

TwiN (approved):
Makes sense. Thank you for the contribution!

TwiN:
@RaminGe is there any chance you could add a test for your change?

RaminGe:
> @RaminGe is there any chance you could add a test for your change?

@TwiN done!

RaminGe:
Hey @TwiN  anything else needed on this? Would really appreciate a merge as we use Gatus for our production system monitoring and our custom alerting workaround sucks

RaminGe:
Hi @TwiN , the suggestions left by Copilot are just lint/robustness improvements for the test file (checking errors on the temp file `os.WriteFile` and `os.Stat` instead of using `_`). 

The core implementation is fully verified and the tests pass perfectly. I/You can apply Copilot's suggested changes with a single click if you prefer or we can merge this as-is since the logic is approved. Let me know what you think - we are eager to get this into production to fix our alerting loop :)

PR #1669, opened 2026-05-14
feat(client): add IMAP STARTTLS support

### LINKED ISSUE #689: STARTTLS does not work work IMAP, workds with SMTP, though
### Describe the bug

When I connect an IMAP STARTTLS endpoint, I get an error. This works for an SMTP endpoint.

When I check both endpoints with openssl, they work fine. One has to say, that for "openssl s_client", you have to specify which protocol to use with -starttls, like smtp, imap a.s.o., but I see no way how to make this work in the Gatus documentation.

### What do you see?

```
Conditions:
X ~ [CONNECTED] (false) == true
X ~ [CERTIFICATE _EXPIRATION} (0) > 3d (0)

Errors:
- short response: * OK [CAPABILITY IMAPrev1 SASL-IR .... STARTTLS] Dovecot ready.

```

No error.

I specify an endpoint using STARTTLS with SMTP like this:

```
endpoints:
  - name: SMTP Endpoint
    group: mail
    url: "starttls://smtp.localdomain:5870"
    interval: 5m
    conditions:
      - "[CONNECTED] == true"
      - "[CERTIFICATE_EXPIRATION] > 7d"

```

it works as expected. However, when I specify:

```
endpoints:
  - name: IMAP Endpoint
    group: mail
    url: "starttls://imap.localdomain:1430"
    interval: 5m
    conditions:
      - "[CONNECTED] == true"
      - "[CERTIFICATE_EXPIRATION] > 7d"

```

### Version

Gatus 5.7.0 (Dockerhub stable version)

### Additional information

The endpoints are provisioned with certificates issued by my own CA, which has been injected via a docker volume to /etc/ssl/certs/ca-certificates.crt. Both endpoints use the same multi-SAN certificate. The same error occurs when I add "client.insecure" to the definitions.

superShen0916:
## Summary
Fixes #689 - Adds IMAP STARTTLS support to `CanPerformStartTLS`, which previously only worked with SMTP.

When connecting to IMAP servers on port 143, the function now performs the IMAP-specific STARTTLS negotiation instead of failing with a "short response" error from the SMTP client.

## Changes
- `client/client.go`: Added IMAP protocol detection by port (143), with a new `canPerformIMAPSTARTTLS` helper implementing the IMAP STARTTLS flow

## How it works
1. Connect to server
2. Read IMAP greeting (`* OK [CAPABILITY ... STARTTLS ...]`)
3. Send `a001 STARTTLS` command
4. Read OK response
5. Upgrade to TLS

## Test plan
- [x] `go build ./client/` succeeds

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"crypto/tls"
@@ -186,6 +187,11 @@ func CanPerformStartTLS(address string, config *Config) (connected bool, certifi
 		}
 	}
 
+	// Detect protocol from port: 143 = IMAP, others = SMTP
+	if hostAndPort[1] == "143" {
+		return canPerformIMAPSTARTTLS(connection, hostAndPort[0], config)
+	}
+
 	smtpClient, err := smtp.NewClient(connection, hostAndPort[0])
 	if err != nil {
 		return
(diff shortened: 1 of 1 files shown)


superShen0916:
Closing this as I’m no longer actively maintaining the patch. Thanks for taking a look.

PR #1671, opened 2026-05-15
feat(alerting): Add new jsm integration based on opsgenie

HartmannVolker:
## Summary
Jira Service Management is the successor of Opsgenie and implements the same API on a different API endpoint. 
I copied over the code and adjusted the api-endpoint and the namings.

Additionally, I added the option to do group based overrides of the config to add e.g. multiple API-Keys.

Based on the comment in #1581

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -834,49 +834,50 @@ endpoints:
 > 📝 If an alerting provider is not properly configured, all alerts configured with the provider's type will be
 > ignored.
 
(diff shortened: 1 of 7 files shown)


HartmannVolker:
@TwiN would you mind having a look at this PR? I think it would be very valuable for anyone that wants to integrate with JSM or needs to migrate from OpsGenie to JSM

PR #1679, opened 2026-05-28
feat(dns): Support TXT query type

hernil:
Allows monitoring of things like spf records and others things found in DNS TXT records.

## Summary
This is my first foray into Go. TXT records can be a bit tricky as there can be multiple of them, and sort of single ones spanning more than 255 characters being broken up into multiples as well. Hopefully this should work - at least it has in my personal testing. 

Everything should work As Expected™ compared to other DNS monitoring so I didn't see much of a need to add noise to the documentation. Do let me know if you disagree! 

## Checklist
- [ ] ~~Updated documentation in `README.md`, if applicable.~~


TwiN (approved):

TwiN:

PR #1680, opened 2026-05-28
feat(pagerduty): include endpoint extra-labels as custom_details in alert payload

ogarcia:
## Summary

- Adds `custom_details` to the PagerDuty event payload when an endpoint defines `extra-labels`
- When no `extra-labels` are present the field is omitted entirely (no breaking change)
- Follows the same pattern already used by the `incidentio` and `email` providers

## Example

```yaml
endpoints:
  - name: My Service
    url: https://example.com
    extra-labels:
      environment: production
      region: es
    alerts:
      - type: pagerduty
```

Results in a PagerDuty event with:
```json
{
  "payload": {
    "summary": "TRIGGERED: My Service - ...",
    "source": "Gatus",
    "severity": "critical",
    "custom_details": {
      "environment": "production",
      "region": "es"
    }
  }
}
```

## Checklist

### DIFF
--- a/alerting/provider/pagerduty/pagerduty.go
+++ b/alerting/provider/pagerduty/pagerduty.go
@@ -122,9 +122,10 @@ type Body struct {
 }
 
 type Payload struct {
-	Summary  string `json:"summary"`
-	Source   string `json:"source"`
-	Severity string `json:"severity"`
+	Summary       string            `json:"summary"`
+	Source        string            `json:"source"`
+	Severity      string            `json:"severity"`
+	CustomDetails map[string]string `json:"custom_details,omitempty"`
 }
 
 // buildRequestBody builds the request body for the provider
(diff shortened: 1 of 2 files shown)


wahajahmed010 (commented):
Clean implementation, @ogarcia! Follows the same pattern as incidentio and email providers — good design choice.

**What I like:**
- omitempty means zero extra-labels = no custom_details in payload (zero-cost when unused)
- Tests cover both with-labels and without-labels scenarios
- Properly refactored the test struct to pass Endpoint instead of building inline — keeps it maintainable

**Minor observation:**
The existing tests use full-string comparison on marshaled JSON, while the new test uses json.Unmarshal + field-level assertions. Slightly inconsistent approach — works fine, but for future PRs it might be cleaner to pick one pattern. Not a blocker at all.

Solid contribution! 🦌

ogarcia:
Thanks for the review! Regarding the test inconsistency — the field-level assertion approach in the new tests is intentional: custom_details is serialized from a map[string]string, and Go doesn't guarantee map key ordering in JSON marshaling, so a full-string comparison would be non-deterministic. Happy to add a note in the test if that makes the intent clearer.

PR #1687, opened 2026-06-04
fix(client): allow SSH checks without command body

### LINKED ISSUE #1660: SSH Endpoint "unexpected end of JSON input"
### Describe the bug

My config:
```yml
endpoints:
  - name: SSH Simulation VM
      group: remote-0
      url: ssh://192.168.2:22
      ssh:
        username: "gatus"
        private-key: |
          -----BEGIN OPENSSH PRIVATE KEY-----
          ****
          -----END OPENSSH PRIVATE KEY-----
      interval: 1m
      conditions:
        - "[CONNECTED] == true"
```
With this I get an unhealthy endpoint even though the connection succeeds.

### What do you see?

My endpoint is unhealthy with the following message:
Timestamp
2026-05-08 14:10:26
Response Time
0ms
Conditions
✓[CONNECTED] == true
Errors
• unexpected end of JSON input

I expect a healthy endpoint because the connection succeeded.

1. Use the config with corresponding ip and private key
2. Start gatus
3. See unhealthy endpoint in UI

### Version

v5.35.0 docker image via ghcr.io/twin/gatus:latest

### Additional information

_No response_

puneetdixit200:
Fixes #1660.

Summary:
- Treat empty or whitespace-only SSH endpoint bodies as authenticated connection checks instead of parsing them as JSON command payloads.
- Keep SSH command execution behavior unchanged when a command body is provided.
- Add an in-process SSH server regression test for the empty-body path.

Tests:
- GOMODCACHE=/tmp/gatus-1660-gomodcache GOCACHE=/tmp/gatus-1660-gocache go test ./client -run TestExecuteSSHCommandWithEmptyBody -count=1
- GOMODCACHE=/tmp/gatus-1660-gomodcache GOCACHE=/tmp/gatus-1660-gocache go test ./client -count=1

Created with assistance from an AI coding agent (GPT-5), with human supervision.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -320,6 +320,9 @@ func ExecuteSSHCommand(sshClient *ssh.Client, body string, config *Config) (bool
 		Command string `json:"command"`
 	}
 	defer sshClient.Close()
+	if strings.TrimSpace(body) == "" {
+		return true, 0, nil, nil
+	}
 	var b Body
 	body = parseLocalAddressPlaceholder(body, sshClient.Conn.LocalAddr())
 	if err := json.Unmarshal([]byte(body), &b); err != nil {
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -2,8 +2,11 @@ package client
 
 import (
 	"bytes"
+	"crypto/rand"
+	"crypto/rsa"
 	"crypto/tls"
 	"io"
+	"net"
 	"net/http"
 	"net/netip"
 	"os"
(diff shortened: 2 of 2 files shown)


wahajahmed010 (commented):
Nice fix, this is a clean solution to #1660. The change itself looks correct — early return on empty/whitespace body prevents the JSON parse error.

**One concern: test flakiness potential**

The 50ms `time.After` in the test is a minor weakness. On a loaded CI runner, the goroutine handling `listener.Accept()` could be delayed past 50ms, causing the test to falsely pass (the channel never fires because the connection hasn't been fully processed yet).

A more robust approach would be to use a closed channel or context deadline instead of a fixed wall-clock timeout. Something like:

select {
case channelType := <-channelTypes:
    t.Fatalf("Expected no SSH channel, got %s", channelType)
case <-done:
}

where `done` is closed after verifying the SSH no-op path. Alternatively, increase the timeout to 500ms-1s so it only triggers on genuine hangs rather than scheduling delays.

Otherwise, the test covers both empty and whitespace bodies well, and the in-process SSH server setup is thorough. Thanks for the contribution.

PR #1696, opened 2026-06-19
Add condition for HTTP response headers

stblassitude:
## Summary
Allow conditions like:
```yaml
  - name: example.com
    url: "https://example.com"
    client:
      ignore-redirect: true
    conditions:
      - "[STATUS] == 302"
      - "[HEADERS].location == https://www.example.com/"
```

Created using Claude:
> Gatus has a number of conditions that can be checked, for example `[BODY]`. For HTTP requests, there is currently no way to check the value of any of the response headers. Implement a new condition `[HEADERS]` that make the response headers available as a map of `header-name` and an array of `value`s. When a header occurs only once, collapse the array to a single string. Make sure to add suitable tests; use the `Location` header that is used with HTTP redirects for that. Update the documentation in README.md to explain how to check header contents.

## Checklist


belitre:
Hi @stblassitude , so I was playing with Claude to get a version of Gatus that checks headers and while doing it, Claude found your PR which looks more or less what I need!!!

But at the same time and while playing with my changes and your changes... Claude found an issue worth flagging, the Claude explanation:

```
Issue: when a header appears more than once (e.g. multiple Set-Cookie values), [HEADERS].name resolves to a JSON-serialized array of all the values as one string:

jsonBytes, err := json.Marshal(values)
...
return string(jsonBytes), nil

That flattened string then goes through the existing isEqual/pattern.Match logic unchanged, which does a plain substring/glob match against whatever string it's given — it has no concept of "this string is actually several distinct values glued together."

The result: a pat() condition can match by spanning the JSON separator between two unrelated values, even when neither individual value actually contains the searched substring. Minimal repro:

values := []string{"session=abc", "theme=dark"}
resolved := `["session=abc","theme=dark"]` // what [HEADERS].Set-Cookie resolves to

isEqual(resolved, `pat(*abc","theme*)`) // → true

That succeeds even though neither "session=abc" nor "theme=dark" contains the substring abc","theme — it only exists because JSON puts "," between the two array elements, and the pattern happens to straddle that seam. Same risk applies to the bare [HEADERS] form (all headers flattened into one JSON object) — a pattern could match across two entirely different header names/values.
```

Of course I don't want to hijack your PR or anything, I have a PR on my fork that I'm going to test in the next days that doesn't have that problem, it compares value by value to avoid the problem, and again I think it's worth flagging since it can be an issue. The PR on the fork I have in case you find it interesting: https://github.com/belitre/gatus/pull/1

Cheers!

stblassitude:
> Hi @stblassitude , so I was playing with Claude to get a version of Gatus that checks headers and while doing it, Claude found your PR which looks more or less what I need!!!
> 
> But at the same time and while playing with my changes and your changes... Claude found an issue worth flagging [...]

Thanks for pointing that out, you are correct.

I had Claude implement a fix (I don't know how much it took from your PR).

@TwiN are we moving in the right direction here? I feel condition.go is getting a bit complicated, and I wonder if the right approach would be to find a proper expression language to implement the conditions with?

belitre:
Hi Again,

> I had Claude implement a fix (I don't know how much it took from your PR).

Yeah, no worries at all, as I said I was doing this on a fork to test it an if it works create a PR in the future when Claude found your PR and got code as well from it, sharing is caring!

Looking forward to see if this feature gets merged and released in the future so I don't need to work with custom versions since this application is so useful! Thanks @TwiN for this application by the way 😄

PR #1700, opened 2026-06-25
fix(UI): Use <a> instead of <span> for links on endpoint cards

### LINKED ISSUE #1676: [Feature Request] Make endpoint card "links" actual links

Thanks for creating this great project!

While the service name on the endpoint cards shown in the health dashboard behave like links (if you hover over them you get an underline, and if you click on them you go to the endpoint page), they are simple `<span>`s with `role="link"` and navigation implemented with JavaScript. It would make the dashboard a lot (!) more useful if they were actual links, so you can right-click them and open them in new tabs (which right now you cannot).

1. I want to open endpoint pages in new tabs, but I cannot. This is especially frustrating since you cannot see a historic overview for endpoints on the dashboard (#863), and opening several endpoint pages on several tabs would be a workaround.
2. It's best practice to use semantic HTML. Using a fake `<span role="link">` with link semantics implemented through JavaScript when you can simply use `<a>` provides worse UX and makes the implementation more complicated than it needs to be.

_No response_

### Additional information

_No response_

aaruni96:
This makes the links clickable and open-able in a new tabs

## Summary

Instead of using `<span role="link">` to _emulate_ a link with javascript, this change uses the standard HTML anchor `<a>` to have an actual link on the endpoint cards. This allows "open in new tab" functionality for endpoints from cards.

Fixes #1676 

## Checklist

  Tested on my local machine that things appear to work as intended
  I don't think this warrants a change in README


felixfontein:
Thanks for implementing this!

TwiN:
Did you check if this change impacts any of the CSS? Otherwise, this seems like a good idea

aaruni96:
I am not sure how to check for that. My line of reasons was that `git grep 'role="link"'` produces only two hits, one in this file, and one more in `web/app/src/components/SuiteCard.vue`. I amn't using suites, so I haven't touched `SuiteCard.vue`.

Based on my (admittedly limited) understanding, this change should have no effects outside of the intended change.

ffplexim:
I did test this, it works as expected, and the CSS looks good as well.

wahajahmed010 (commented):
## Review: TwiN/gatus#1700 — fix(UI): Use <a> instead of <span> for links on endpoint cards

**Summary:** Replaces a <span> with an <a> tag for endpoint detail links, removing the JS-based router.push navigation in favor of native <a href>. This is a clear accessibility and UX improvement.

**Logic:** ✅ Correct. Native <a> elements are keyboard-focusable by default, work with screen readers, support right-click → open in new tab, and degrade gracefully if JS fails.

**Tests:** No test changes needed — this is purely a markup change.

**Accessibility:** ✅ Major improvement. The old <span> with role="link" and tabindex="0" was a workaround; the <a> is the semantically correct element.

**Performance:** ✅ Removes the useRouter import entirely — one less dependency in the component.

**No issues found.** Clean improvement, good to merge.

PR #1706, opened 2026-07-01
feat: Implement zeroconf mdns discovery

TN-1:
## Summary
This PR enables Zeroconf service discovery. This is to enable the Home Assistant integration I am working on to automatically discover and configure itself when Gatus is running on the local network

Currently enabled by default ~~with no way to turn off. If its preferable, could add a config option to disable it. Thoughts?~~ There is now a config flag (`disable-zeroconf`) to disable this feature if required.

## Checklist

Tested locally, seems to be working as expected

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,7 @@ If you want to test it locally, see [Docker](#docker).
 | Parameter                    | Description                                                                                                                              | Default       |
 |:-----------------------------|:-----------------------------------------------------------------------------------------------------------------------------------------|:--------------|
 | `metrics`                    | Whether to expose metrics at `/metrics`.                                                                                                 | `false`       |
+| `disable-zeroconf`           | Whether to disable zeroconf service autodiscovery.                                                                                       | `false`       |
 | `storage`                    | [Storage configuration](#storage).                                                                                                       | `{}`          |
 | `alerting`                   | [Alerting configuration](#alerting).                                                                                                     | `{}`          |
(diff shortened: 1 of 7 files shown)


wahajahmed010 (commented):
## Review: TwiN/gatus#1706 — feat: Implement zeroconf mdns discovery

**Summary:** Adds mDNS/Zeroconf service discovery using hashicorp/mdns. Well-structured with a dedicated zeroconf package, config toggle, and proper lifecycle management.

**Logic:** ✅ Sound. Initialize() creates an mDNS service advertising Gatus on _http._tcp, Shutdown() tears it down gracefully. The localIPs() helper filters loopback, docker, and bridge interfaces — sensible defaults.

**Tests:** ✅ Config parsing test for disable-zeroconf: true covers the toggle.

**Security:** ⚠️ The disable-zeroconf flag defaults to false in the README, meaning mDNS is on by default. This is a potential concern for users who don't read docs. Consider making it opt-in (default true).

**Performance:** ✅ Minimal overhead — one goroutine from the mdns library, no polling.

**Nit:** The localIPs() function skips interfaces named docker0 or starting with br-, but doesn't handle other virtual interfaces (e.g., tun, veth). Consider a broader filter or making the interface list configurable.

**Verdict:** Good feature, clean implementation. Recommend making mDNS opt-in (default disabled) for security-conscious deployments.

TN-1:
The integration has been accepted into Home Assistant. If/When this PR is merged, I will be able to update the integration to support automatic configuration.

PR #1712, opened 2026-07-07
feat: Filter remote endpoints by group (filter-groups)

selfhoster1312:
I wanted to be able to monitor electrical downtime on my machines (rare, but does happen). Since gatus has no way of knowing when a check was not performed (due to the instance being offline), i'm using the remote instances feature.

In order to avoid adding too much stuff to each instance's UI, and only add remote reachability tests, i'm filtering by groups.

Example config `site1.yaml`:

```yaml
storage:
  type: "sqlite"
  path: "gatus.sqlite"
remote:
  instances:
    - endpoint-prefix: "from-site2 "
      filter-groups: [ "site1" ]
      url: "https://site2.example.com/api/v1/endpoints/statuses"
endpoints:
  - name: router
    url: "icmp://192.168.1.1"
    group: site1
    interval: 5s
    conditions:
      - "[CONNECTED] == true"
  - name: site2
    url: "icmp://IP2"
    group: site2
    interval: 5s
    conditions:
      - "[CONNECTED] == true"
```

Example config `site2.yaml`:

```yaml
storage:
  type: "sqlite"
  path: "gatus.sqlite"
remote:
  instances:
    - endpoint-prefix: "from-site1 "
      filter-groups: [ "site2" ]
      url: "https://site1.example.com/api/v1/endpoints/statuses"
endpoints:
  - name: router
    url: "icmp://192.168.2.1"
    group: site2
    interval: 5s
    conditions:
      - "[CONNECTED] == true"
  - name: site1
    url: "icmp://IP1"
    group: site1
    interval: 5s
    conditions:
      - "[CONNECTED] == true"
```

I did not add a test because the remote feature is marked alpha and i didn't find remote tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2839,6 +2839,9 @@ There are two main use cases for this:
 - You have multiple Gatus instances running on different machines, and you wish to visually expose the statuses through a single dashboard
 - You have one or more Gatus instances that are not publicly accessible (e.g. behind a firewall), and you wish to retrieve
 
+It is possible to filter imported endpoints with the `filter-groups` parameter. This is for example useful if you have separate
+gatus instances, but would like to perform inter-site connectivity tests.
+
 This is an experimental feature. It may be removed or updated in a breaking manner at any time. Furthermore,
 there are known issues with this feature. If you'd like to provide some feedback, please write a comment in [#64](https://github.com/TwiN/gatus/issues/64).
 Use at your own risk.
@@ -2848,6 +2851,7 @@ Use at your own risk.
 | `remote`                           | Remote configuration                           | `{}`          |
 | `remote.instances`                 | List of remote instances                       | Required `[]` |
 | `remote.instances.endpoint-prefix` | String to prefix all endpoint names with       | `""`          |
+| `remote.instances.filter-groups`   | List of groups to import (default all)         | `[]`          |
 | `remote.instances.url`             | URL from which to retrieve endpoint statuses   | Required `""` |
 | `remote.client`                    | [Client configuration](#client-configuration). | `{}`          |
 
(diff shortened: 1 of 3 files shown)


wahajahmed010 (commented):
## Review: TwiN/gatus#1712 — Filter remote endpoints by group

**Overall:** Simple, clean feature. The implementation is minimal and correct.

**Strengths:**
- Uses `slices.Contains` for O(n) filtering — fine for the expected small group lists
- Debug logging when endpoints are filtered is helpful for troubleshooting
- Backward compatible: empty `FilterGroups` means no filtering (default behavior preserved)
- README docs updated

**Nit:** The `ignoredEndpoints` slice collects names but is only used for logging. Consider also logging the group name for easier debugging.

**Verdict:** Looks good. No issues.

selfhoster1312:
Hello @wahajahmed010, i added the requested extra info in the debug log, and also included one more (small) commit for an obvious optimization i missed the first time around. I'm happy to squash these 3 commits if that's how you do things around here.

While I thank you for the suggestion, I would appreciate talking to a human. It's ok if review takes time, and if communication is sometimes hard. But i have no time and energy to waste talking to a LLM.

LLMs are a scourge on our craft and are destroying our planet, and i personally believe they should not be welcome anywhere and their use is shameful. I recognize this is not my project and not my rules, so i'm not going to argue for banning LLMs and will refrain from judging their use (outside of my own PRs). As common courtesy, i would really appreciate talking to you and not to your agents.

I felt really insulted receiving a response from a bot (especially one not marked as such), and i suspect you will likewise feel insulted by my reaction (which i assure you has already been toned down). I hope you are doing well and we find a way to build a better connection in the future.

Issue #329, opened 2022-09-12
Detail page of remote endpoints throw 404


praveen-livspace:
### Describe the bug

Clicking on any of the remote endpoint on the main page takes to a 404 page.

### What do you see?

A blank detail page 

### What do you expect to see?

Detailed status of the remote endpoint.

### List the steps that must be taken to reproduce this issue

1. Start the application with a remote endpoint.
2. Click on the remote endpoint hyperlink.
3. Notice that the detail page does not show any data.

### Version

4.1.0

TwiN:
Related: #64

praveen-livspace:
@TwiN please check this PR #330 . It handles the case of remote endpoints

m-emelchenkov:
Absolutely, it's a killer feature of Gatus (comparing e.g. to Uptime Kuma) and should works well. It's a shame to see that it hasn't been fully implemented. I see commit #330 with status "Closed with unmerged commits". Actually, I don't understand what this phrase means: work on the bug has been stopped?

PR #1715, opened 2026-07-09
fix: Route for remote endpoint detailed view

selfhoster1312:
I was bothered that when i clicked on a remote endpoint no data was shown at all. I implemented it.

It's a bit hacky:

- relies on iteration count to identify the endpoint (should only make problems when dynamically reloading config, and even then, it will only produce an error, not panic)
- takes the remote endpoints URI and rewrites it on the go, a future improvement would be to use a remote API URL and build a proper API client
- compiles a regex on the go (could be globally compiled once, or we could do without a regex at all as a performance improvement)

It's not perfect but since remote instances is an alpha feature i think it fits the bill, and it's useful for me so i thought i'd share it with others and not just keep it on my branch.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -77,14 +77,14 @@ func (a *API) createRouter(cfg *config.Config) *fiber.App {
 	////////////////////////
 	unprotectedAPIRouter := apiRouter.Group("/")
 	unprotectedAPIRouter.Get("/v1/config", ConfigHandler{securityConfig: cfg.Security, config: cfg}.GetConfig)
-	unprotectedAPIRouter.Get("/v1/endpoints/:key/health/badge.svg", HealthBadge)
-	unprotectedAPIRouter.Get("/v1/endpoints/:key/health/badge.shields", HealthBadgeShields)
-	unprotectedAPIRouter.Get("/v1/endpoints/:key/uptimes/:duration", UptimeRaw)
-	unprotectedAPIRouter.Get("/v1/endpoints/:key/uptimes/:duration/badge.svg", UptimeBadge)
-	unprotectedAPIRouter.Get("/v1/endpoints/:key/response-times/:duration", ResponseTimeRaw)
-	unprotectedAPIRouter.Get("/v1/endpoints/:key/response-times/:duration/badge.svg", ResponseTimeBadge(cfg))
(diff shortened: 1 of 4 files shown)


selfhoster1312:
Fix for #329

selfhoster1312:
I realized there's more routes to fix than just `EndpointStatus`, for example for the badges. Working on a more general solutions. Don't merge/review yet.

api/api.go
```diff
@@ -129,8 +129,10 @@ func (a *API) createRouter(cfg *config.Config) *fiber.App {
 		}
 	}
 	protectedAPIRouter.Get("/v1/endpoints/statuses", EndpointStatuses(cfg))
+	// Remote endpoint statuses cannot be proxied as-is because we
+	// need to edit the JSON on the go to replace local key with remote key
 	protectedAPIRouter.Get("/v1/endpoints/:key/statuses", EndpointStatus(cfg))
 	protectedAPIRouter.Get("/v1/suites/statuses", SuiteStatuses(cfg))
-	protectedAPIRouter.Get("/v1/suites/:key/statuses", SuiteStatus(cfg))
+	protectedAPIRouter.Get("/v1/suites/:key/statuses", queryRemoteOrLocalHandlerWithCfg(cfg, SuiteStatus, "SuiteStatus"))
```
selfhoster1312, line 136: This one may also need special-casing. Should add a remote suite test.

EDIT: Actually there was no remote integration at all in suites. Is this a good or a bad thing? I've never used suites so i can't really figure out what's a good UX for that.

selfhoster1312:
I added more generic methods that will pass through to the local handler or the remote handler depending on the key prefix. For now only `EndpointStatus` is special-cased because it needs to rewrite the JSON to add the `endpoint-prefix` again when the remote instance replies.

I'm not sure what to do about suites. I guess this could be reviewed as-is and if people want to implement remote support for suites that can be a new PR?

I added some tests, and the URL-rewriting (to pass the request to the remote instance) is now based on parsed URLs and not raw strings, to avoid messing with arguments. I'm still not a big fan of this, i think a `Instance.ApiURL` setting would be really helpful but that's a story for another day.

I think this is ready for review.

PR #1716, opened 2026-07-10
feat(client): Allow configuration of client cookie storage

### LINKED ISSUE #1711: Store cookie for suites/redirects

Apologies if this is already a feature but I couldn't find anything in the documentation and cookies don't seem to be stored.

I have a requirement to store cookies to allow an endpoint to complete a request.

I have an application I am trying to monitor. It has an auth feature but I have it disabled. It still goes through the auth cookie creation process though. As a result, it needs to store the cookies returned in the `Set-Cookie` header and transmit them in the `Cookie` for the next request.

I have tried to achieve this using an endpoint but cookies from 302 responses don't seem to be stored and used in the next request. I have also tried with suites and follow redirects disabled but get a similar failure due to the missing cookie.

2 days

### Additional information

_No response_

AaronLionsheep:
Fixes #1711 

## Summary
The client configuration now allows a CookieJar to be set by a new client setting, `store-cookies`. The default behavior does not change where cookies are not stored or sent to subsequent requests.

The motivation for this feature is to be able to monitor services behind an authentication sidecar (like [oauth2-proxy](https://github.com/oauth2-proxy/oauth2-proxy)) without bypassing the sidecar. [oauth2-proxy intercepts an unauthenticated request and redirects the client to a login area](https://oauth2-proxy.github.io/oauth2-proxy/#architecture). Once the client authenticates (automatically with mTLS for this example), they are redirected back to their original route with the addition of an authentication cookie. oauth2-proxy verifies this cookie and then allows the request to the targeted backend service.

Storing cookies allows an endpoint check to authenticate via mTLS through an authentication sidecar by storing and sending the cookie set during an authentication redirect chain. Subsequent checks are able to reuse the cookie and allow the authentication sidecar to validate or start a new authentication flow.

## Checklist

### DIFF
--- a/client/config.go
+++ b/client/config.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"net"
 	"net/http"
+	"net/http/cookiejar"
 	"net/url"
 	"regexp"
 	"strconv"
@@ -34,6 +35,7 @@ var (
 		IgnoreRedirect: false,
 		Timeout:        defaultTimeout,
 		Network:        "ip",
+		StoreCookies:   false,
 	}
 )
 
@@ -81,6 +83,9 @@ type Config struct {
 
 	// ResolvedTunnel is the resolved SSH tunnel for this specific Config
 	ResolvedTunnel *sshtunnel.SSHTunnel `yaml:"-"`
+
+	// StoreCookies determines whether cookies are stored and included across requests.
+	StoreCookies bool `yaml:"store-cookies,omitempty"`
 
 	httpClient *http.Client
 }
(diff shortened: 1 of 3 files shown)


AaronLionsheep:
@TwiN Thanks for sharing this project with the community! This PR fixes the only missing piece that would make this the perfect solution for my use-case. It should also address an issue filed by another user. Selfishly, I'd love to see this get merged!

AaronLionsheep:
@TwiN I know there is a backlog of great PR's, but it would be great to see this get merged in. It also overlaps with an existing issue, so there are at least 2 users that would love to see this PR merged 😃

client/config.go
```diff
@@ -280,6 +285,13 @@ func (c *Config) getHTTPClient() *http.Client {
 				}
 			}
 		}
+		if c.StoreCookies {
+			jar, err := cookiejar.New(nil)
+			if err != nil {
+				logr.Fatalf("Failed to initialize cookie jar: %v", err)
```
TwiN:
```suggestion
				logr.Fatalf("[client.getHTTPClient] Failed to initialize cookie jar: %v", err)
```

TwiN (approved):

AaronLionsheep:
Thanks @TwiN! Your logging change makes sense, and I'm happy to make any other adjustments. Merge at your convenience!

TwiN:
@AaronLionsheep Thank you for the contribution!

PR #1721, opened 2026-07-13
Exclude remote-from-remote endpoints via `instance.include-remote`

selfhoster1312:
Adding a new setting in remote instance setting: ~~`include_remote`~~ (bool) -> renamed `include-remote` for consistency with `endpoint-prefix`.

When enabled, remote instances from the queried remote instance will be included (some sort of recursion if you will). This was the default behavior until now, but can lead to duplicates, like having both `endpoint1` and `from-site1 endpoint1`  on your `site1` dashboard.

~~**This PR changes the default behavior** to note include remote endpoints from the queried remote instances. This should also help some pathological case when one remote instance was slow to respond which would trickle down on other instances (this is still happening, but only with the instances you declare directly in your config file).~~

This PR has been edited and now doesn't change the default behavior, but will print a message on config (re)load when no explicit value has been set, inciting users to update their config to remove the warning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2847,6 +2847,7 @@ Use at your own risk.
 | `remote.instances`                 | List of remote instances                       | Required `[]` |
 | `remote.instances.endpoint-prefix` | String to prefix all endpoint names with       | `""`          |
 | `remote.instances.url`             | URL from which to retrieve endpoint statuses   | Required `""` |
+| `remote.instances.include-remote`  | Include remote endpoints from remote instance  | true          |
 | `remote.client`                    | [Client configuration](#client-configuration). | `{}`          |
 
 ```yaml
--- a/api/endpoint_status.go
+++ b/api/endpoint_status.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"strings"
 
 	"github.com/TwiN/gatus/v5/client"
 	"github.com/TwiN/gatus/v5/config"
@@ -22,7 +23,16 @@ import (
 func EndpointStatuses(cfg *config.Config) fiber.Handler {
 	return func(c *fiber.Ctx) error {
 		page, pageSize := extractPageAndPageSizeFromRequest(c, cfg.Storage.MaximumNumberOfResults)
-		value, exists := cache.Get(fmt.Sprintf("endpoint-status-%d-%d", page, pageSize))
+
+		// Does the client request remote endpoints? By default yes, for web UI display
+		// and for backwards compatibility. Remote instance queries should have remote=false to
+		// avoid duplicates.
+		remoteQuery := c.Query("remote", "true")
+		// remote query param value is case-insensitive (eg. `TRUE` and `True` are also truthy values)
+		includeRemote := strings.EqualFold(remoteQuery, "true")
+		logr.Debugf("Received EndpointStatuses request with remote=%t")
+
+		value, exists := cache.Get(fmt.Sprintf("endpoint-status-%d-%d-%t", page, pageSize, includeRemote))
 		var data []byte
 		if !exists {
 			endpointStatuses, err := store.Get().GetAllEndpointStatuses(paging.NewEndpointStatusParams().WithResults(page, pageSize))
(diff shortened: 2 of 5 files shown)


wahajahmed010:
Thanks for tackling this — duplicate remote endpoints showing up twice in the dashboard is a UX papercut that's been on the list for a while.

A few concerns, mostly about the **default-behavior change** since this is technically breaking for users who depended on the recursive include:

1. **Silent default flip.** The PR body says "this changes the default behavior" but the new `include_remote` field defaults to `false` and the API query param now defaults to `remote=true` only for direct clients. That mismatch is going to bite someone:
   - A user who upgrades and has *no* `include_remote` in their config will stop seeing remote-of-remote endpoints — but the web UI *itself* will keep including them (because the UI calls `?remote=true`). So the dashboard will look "right" while any external consumer (or the `--remote-instance` CLI) silently gets less data.
   - I'd suggest either (a) logging a one-shot WARN when an instance is configured without `include_remote`, or (b) making the default explicit in the docs release notes.

2. **Query-param parsing.** `c.Query("remote", "true")` treats *any* value other than the literal string `"true"` as `false`. That means `?remote=1`, `?remote=yes`, or `?remote=TRUE` would all be interpreted as `false`. If the intent is a boolean, an explicit `strings.EqualFold(remoteQuery, "true")` would be more forgiving. Also worth adding a unit test that asserts the default and the two valid values.

3. **URL building.** `instance.URL + "?remote=true"` assumes the URL has no existing query string. If a user has a `?token=...` or similar on their `instance.url`, this will produce a malformed URL. Use `net/url` to build the new URL, or at minimum check for `?` vs `&`.

4. **Tests.** I don't see test coverage for the new flag in this diff. There should be at minimum:
   - a test confirming `include_remote: false` (default) excludes remote-of-remote endpoints,
   - a test confirming `include_remote: true` includes them,
   - a test confirming the `?remote=false` query param is honored by the handler.

5. **Minor: README ordering.** The new row in the config table is wedged between `endpoint-prefix` and `url` — conventionally `url` would be the first required field. Move it after `url` to match the field order in `config/remote/remote.go`.

The feature itself is useful and the patch is small. Once the default-flip and URL-construction issues are addressed, this is good to go.

selfhoster1312:
Thanks for the review @wahajahmed010. I will post comments here as i address your concerns.

## Concern 1 

I addressed concern 1 by setting the default back to `include-remote: true` and printing a warning when one or more remote instances don't have an explicit value configured.

I really wanted to make it simple but it seems like the yaml package ([**which is now unmaintained btw**](https://github.com/go-yaml/yaml#this-project-is-unmaintained)) doesn't have a tag to set a default value, and whatever other approach i tried either produced a stack overflow ([similar to this](https://github.com/go-yaml/yaml/issues/716)) or didn't apply a default, so i went with `*bool` instead of `bool`.

I added some tests in `config/config_test.go` because there's no dedicated remote tests. I just wanted to make sure the config is parsed properly. Let me know if it's not the appropriate spot. EDIT: now i also added endpoint status tests (see below).

## Concern 2/4

I used `strings.EqualFold` as you suggested. I don't think it's a good idea to support more more value types (eg. int) as that would be confusing.

I added some tests. Some are unit tests, some are integration tests with 2 actual routers running.

## Concern 3

I parsed the URL in the end and edited the query from there. It's parsed on every occurrence because i did not find a convenient way to make the YAML parser work with default values and intermediate types. After some hours of stack overflows and nil-pointer dereferences, i probably won't try again (`<tongue-in-cheek>unless we're rewriting gatus in rust</tongue-in-cheek>`).

## Other notes

I switched the default back to `true` but i think it's wrong. Because if two instances mirror one another, they will produce infinite recursion as shown in the test `TestEndpointStatusesInfiniteRecursion`. I'm open to suggestions, but i may set it back to `false` with the newly introduced warning.

I'll address other review items now.

selfhoster1312:
I think i adressed most of your concerns. Review item 5 i'm not sure what you mean.

I'll wait for you or someone else to chime in about what the default for the setting should be and how to prevent the infinite recursion if we make it a default everywhere.

PR #1722, opened 2026-07-14
feat(config): Add ignore-connectivity flag to endpoints

ArsenijSerjogin:
Allow endpoints to ignore the global connectivity check, enabling local services to be monitored alongside external services.

## Summary

There is a way to prevent false negative results for a service availability like that:
```
connectivity:
  checker:
    target: 8.8.8.8:53
    interval: 60s
```

The problem is that this is a global config, so monitoring of local infrastructure become impossible if internet is down. 

This probably a very niche situations when you want this issue to be solved.  
My case is that I use Gatus on the personal PC to monitor my sites, and by the way my internet provider.

So config may look roughly like this:
```
connectivity:
  checker:
    target: 8.8.8.8:53
    interval: 5s

endpoints:
  - name: front-end
    group: core
    url: "https://twin.sh/health"
    interval: 10s
    conditions:
      - "[STATUS] == 200"
      - "[BODY].status == UP"
      - "[RESPONSE_TIME] < 150"

  - name: back-end
    group: core
    url: "https://example.org/"
    interval: 10s
    conditions:
      - "[STATUS] == 200"
      - "[CERTIFICATE_EXPIRATION] > 48h"

  - name: example-dns-query
    url: "8.8.8.8" 
    interval: 10s
    ignore-connectivity: true
    dns:
      query-name: "example.com"
      query-type: "A"
    conditions:
      - "[BODY] == pat(*.*.*.*)"  # Matches any IPv4 address
      - "[DNS_RCODE] == NOERROR"

  - name: Router
    url: "http://192.168.0.1"
    interval: 10s
    ignore-connectivity: true
    conditions:
      - "[CONNECTED] == true"
```
Screenshot from 2026-07-14 19-52-38

In this example I intentionally turned off the internet, so it's clear that external resources was not tracked while the local infrastructure showed the red bars.

## Checklist

- [x] Tested to validate that the changes work as intended, if applicable.
- [ ] Added tests to validate that the changes work as intended, if applicable


ArsenijSerjogin:
I hesitated do I need to make this PR. Feel free to close it if it not what Gatus need, in such case I would be happy to hear why this feature should not exist. :D

PR #1723, opened 2026-07-15
docs(alerting): Revise HomeAssistant alerts and add Home Assistant integration info

TN-1:
## Summary

I have authored an integration for Home Assistant that has been accepted and will be merged into the next release in August. This means that every Home Assistant install will include official support for Gatus out of the box, and will allow a much easier experience than current. This PR updates the README to reflect these changes. 

I have left the existing instructions in the README marked as Legacy with a note to use the integration. Might be an idea to just remove the section altogether in the future? I leave this decision to the maintainers.

Note: Please do not merge before Home Assistant Release 2026.8. I will mark PR as draft until the release has been made.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,6 @@ Have any feedback or questions? [Create a discussion](https://github.com/TwiN/ga
   - [Helm Chart](#helm-chart)
   - [Terraform](#terraform)
     - [Kubernetes](#kubernetes)
-- [Native Home Assistant Integration](#native-home-assistant-integration)
 - [Running the tests](#running-the-tests)
 - [Using in Production](#using-in-production)
 - [FAQ](#faq)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -66,7 +66,7 @@ Have any feedback or questions? [Create a discussion](https://github.com/TwiN/ga
     - [Configuring GitLab alerts](#configuring-gitlab-alerts)
     - [Configuring Google Chat alerts](#configuring-google-chat-alerts)
     - [Configuring Gotify alerts](#configuring-gotify-alerts)
-    - [Configuring HomeAssistant alerts](#configuring-homeassistant-alerts)
+    - [Configuring HomeAssistant alerts (Legacy)](#configuring-homeassistant-alerts-legacy)
```
TwiN: To clarify, is the home assistant alerts actually legacy, or is it just because there's now an official gatus integration in home assistant?

I'm all for documenting the new Gatus integration offered directly in Home Assistant, but this makes it sound like pushing alerts is deprecated

TN-1: Because there is now an integration. Things can be done alot easier, just using automations without the heavy yaml backend. Open to revision of the wording if you have a better way to write it.

TN-1: @TwiN here are some screenshots to show how it looks with the integration. Then you can use automations to make alerts all in the UI. 
Screenshot_20260719-142251
Screenshot_20260719-142255
Screenshot_20260719-142257

The endpoints will eventually have more sensors to represent other data points exposed, but for now it's just a binary sensor True/False if the endpoint is up .

TN-1:
Home Assistant 2026.8 has been released with the Gatus integration so the Readme is good to be updated now

README.md
TwiN: With that in mind, I don't think it's accurate to call the existing approach "legacy". Perhaps "native" would be a better way to differentiate the new way (native) vs the currently documented approach

TN-1: Honestly, I agree. I didnt think of that wording, so I appreciate your suggestion. Will make the change

TN-1:
What do you think @TwiN ?

README.md
```diff
@@ -846,7 +847,7 @@ endpoints:
 | `alerting.gitlab`          | Configuration for alerts of type `gitlab`. <br />See [Configuring GitLab alerts](#configuring-gitlab-alerts).                           | `{}`    |
 | `alerting.googlechat`      | Configuration for alerts of type `googlechat`. <br />See [Configuring Google Chat alerts](#configuring-google-chat-alerts).             | `{}`    |
 | `alerting.gotify`          | Configuration for alerts of type `gotify`. <br />See [Configuring Gotify alerts](#configuring-gotify-alerts).                           | `{}`    |
-| `alerting.homeassistant`   | Configuration for alerts of type `homeassistant`. <br />See [Configuring HomeAssistant alerts](#configuring-homeassistant-alerts).      | `{}`    |
+| `alerting.homeassistant`   | Configuration for alerts of type `homeassistant`. <br />See [Configuring HomeAssistant alerts](#configuring-homeassistant-alerts).| `{}`    |
```
TwiN: Can you re-add the spaces so that the table is properly formatted again?

README.md
```diff
@@ -2906,6 +2911,57 @@ To get more details, please check [chart's configuration](https://github.com/Twi
 
 Gatus can be deployed on Kubernetes using Terraform by using the following module: [terraform-kubernetes-gatus](https://github.com/TwiN/terraform-kubernetes-gatus).
 
+## Native Home Assistant Integration
```
TwiN: Can you put this under configuring home assistant alerts instead? Just add a section at the end that mentions alternative implementation, with it being the native integration documentation you wrote.

Try to keep it concise, or use a collapsible tag so that it doesn't take too much space given that the readme is pretty lengthy

TwiN (changes requested):

TN-1:
@TwiN Integrated your feedback

TwiN (approved):

TwiN:
Excellent work @TN-1

PR #1725, opened 2026-07-16
feat: support serving Gatus behind a reverse-proxy subpath (#88)

shornabho:
## Summary

Solves issue #88, by squashing and rebasing the changes (due to divergence), based on the work done and feedback received in:
- #908, that exists as a draft PR
- #1424, that exists as an open PR but went stale 

Appreciate the headstart from @simonhammes, @s4s0l and @PythonGermany who contributed to the work so far.

------

Adds an opt-in `web.base-path` option so Gatus can be hosted at a subpath (e.g. https://example.com/gatus/). It defaults to "/", preserving the current behaviour, so this is not a breaking change.

Gatus continues to serve at root; base-path only makes the SPA emit a subpath-aware <base href> with relative asset/API URLs, and makes OIDC cookie paths and post-login redirects honour the prefix. The reverse proxy strips the prefix before forwarding.

## Checklist


earfman:
Two of the changes here aren't covered by the new tests. I checked by reverting each hunk on its own and re-running this PR's own suite.

**1. `config/config.go` L341-350 — the propagation itself**

```go
config.UI.BasePath = config.Web.BasePath
if config.Security != nil && config.Security.OIDC != nil {
    config.Security.OIDC.BasePath = config.Web.BasePath
}
```

Delete those four lines and all 63 tests in the PR still pass.

That block is the only place `web.base-path` reaches `UI` and `Security.OIDC`. The new cases in `config/ui/ui_test.go`, `config/web/web_test.go` and `security/oidc_test.go` each set `BasePath` directly on their own struct, so none of them go through `parseAndValidateConfigBytes`; `config/config_test.go` isn't touched by this PR and has no `base-path` case. So if that block is ever dropped in a refactor, `web.base-path: /gatus/` silently becomes a no-op for both the UI and OIDC, and CI stays green.

**2. `security/oidc.go` L132-149 — the post-login redirects**

The three `http.Redirect(w, r, c.BasePath, ...)` calls in `callbackHandler` (success, allowed-subject, access-denied). Revert just those and the suite is still fully green. `TestOIDCConfig_setSessionCookieWithCustomBasePath` does gate the cookie `Path` correctly, but nothing exercises the redirect target — which is the half a user notices: log in, land back at `/` instead of `/gatus/`.

The rest of the Go side gates properly: 13 of the 15 reachable hunks fail the suite when reverted, several at build time. The 17 frontend hunks (`.vue`/`.html`/`.json`) are outside the Go runner and I didn't evaluate them.

Two small suggestions, if useful:

- a `config/config_test.go` case that parses YAML with `web.base-path` set and asserts `config.UI.BasePath` and `config.Security.OIDC.BasePath` both come out equal to it
- an assertion on `w.Header().Get("Location")` in the existing callback-handler test

Nice piece of work otherwise — #88 has been open a long time.

<sub>Found with [coretexa-verify](https://github.com/earfman/coretexa-verify), which reverts a PR's source changes, keeps its new tests, and re-runs them. Happy to be told I'm wrong on either point.</sub>

PR #1726, opened 2026-07-16
feat: support serving Gatus behind on a subpath natively (#88)

shornabho:
## Summary 

This solves issue #88 , allowing Gatus to be hosted on a subpath like (e.g.: http://example.com/gatus/).

Extended on: 
- #1725, which itself is an extension of:
  - #908
  - #1424 

In this version, it natively hosts the entire app (api, ui, assets) under the web.base-path, making a reverse-proxy entirely optional. What #1725 addresses, keeps changes minimal but makes the single-binary Gatus app broken when web.base-path is set without a reverse-proxy. Any proxy placed in front now only needs to pass the prefix through unchanged rather than strip it (e.g. Caddy `handle` instead of `handle_path`). The root path redirects to the base path for convenience. The default configuration (base-path: /) is unaffected and continues to serve everything at the root.

The Caddy reverse-proxy examples are also updated to pass-through and adjusts the docs.

## Checklist


shornabho:
@TwiN requesting you review on this. 

This would be a huge help, especially for setups that require stack based deployments and subpath routing become the preferred way for automated deployments. This will open the doors for Gatus to be used across the world in professional setups and will increase adoption. I relate to this problem personally, and hence I'm keen on solving this for good. 

> Note: I would be happy to contribute effectively to Gatus, as this is exactly the tool I have been looking for, for a while now. While Uptime Kuma is great, it's not the best for professional GitOps friendly setups. The way you have approached Gatus seems like the way to go (provided we solve some small but important issues), and I believe in it!

gbr-ufs:
This is my only blocker for using Gatus on my homelab for now. I'm using Tailscale and have no domain for now, so I can't use "gatus.server.tailscaledomain.ts.net", which makes me reliant on subpaths.

Piero-93:
@shornabho been running this on my homelab for a couple of days — for what it's worth, the `<base href>` approach does work. `publicPath: './'` makes Vue CLI emit an empty `BASE_URL`, and vue-router then falls back to reading the `<base>` element, which neatly sidesteps the pushState problem that killed `web.context-root` back in v2.0.0.

Ran into two bugs while testing, so I opened a PR against your branch rather than starting a fifth one: https://github.com/shornabho/gatus/pull/1

One's the favicon: the placeholders in `index.html` got replaced with hardcoded names, so `ui.favicon` silently stops working for everyone, not just with a base path. The other is that `base-path: /gatus` gets rejected even though the field's comment promises the slashes are added for you — and since bad config is fatal, Gatus won't start at all.

Your branch is a couple of months behind master by now, by the way. It still rebases cleanly — I've got that on a separate branch if it saves you any time.

PR #1727, opened 2026-07-16
Support an explicit endpoint key to decouple identity from name

### LINKED ISSUE #1146: Custom ID for endpoint

Add `endpoint.id` config option to allow manually specifying internal unique identifier used in a database to store historical uptime data. Now ID is generated only automatically as `{group}_{name}`.

Renaming a service and even changing its group leads to the loss of all previous uptime data. That's kinda inconvenient. What if I decided to rename groups while having a long uptime history? (Fortunately, I've started using Gatus only recently.)

Sorry if a duplicate, didn't find smth similar

For 2 days

### Additional information

_No response_

sleeyax:
## Summary

The endpoint key is derived from group and name and is used as the endpoint's identity in storage and API routes, so renaming an endpoint or moving it between groups orphans its history. 

Added an optional `key` field to `Endpoint` and `ExternalEndpoint` that, when set, is used as the key instead (sanitized, without a group prefix), keeping identity stable across name and group changes.

The explicit key is threaded through `ExternalEndpoint.ToEndpoint()` so pushed results are stored under it, and the memory and SQL stores now report the actual key on `Status` rather than re-deriving it from group and name.

Fixes #1146.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,7 @@ You can then configure alerts to be triggered when an endpoint is unhealthy once
 | `endpoints`                                     | List of endpoints to monitor.                                                                                                               | Required `[]`              |
 | `endpoints[].enabled`                           | Whether to monitor the endpoint.                                                                                                            | `true`                     |
 | `endpoints[].name`                              | Name of the endpoint. Can be anything.                                                                                                      | Required `""`              |
+| `endpoints[].key`                               | Explicit unique key used for storage and API routes instead of deriving it from group and name. <br />See [Endpoint key](#endpoint-key).   | `""`                       |
 | `endpoints[].group`                             | Group name. Used to group multiple endpoints together on the dashboard. <br />See [Endpoint groups](#endpoint-groups).                      | `""`                       |
 | `endpoints[].url`                               | URL to send the request to.                                                                                                                 | Required `""`              |
 | `endpoints[].method`                            | Request method.                                                                                                                             | `GET`                      |
@@ -329,6 +330,7 @@ For instance:
 | `external-endpoints`                      | List of endpoints to monitor.                                                                                                     | `[]`           |
 | `external-endpoints[].enabled`            | Whether to monitor the endpoint.                                                                                                  | `true`         |
 | `external-endpoints[].name`               | Name of the endpoint. Can be anything.                                                                                            | Required `""`  |
+| `external-endpoints[].key`                | Explicit unique key used for storage and API routes instead of deriving it from group and name. <br />See [Endpoint key](#endpoint-key). | `""`     |
 | `external-endpoints[].group`              | Group name. Used to group multiple endpoints together on the dashboard. <br />See [Endpoint groups](#endpoint-groups).            | `""`           |
 | `external-endpoints[].token`              | Bearer token required to push status to.                                                                                          | Required `""`  |
 | `external-endpoints[].alerts`             | List of all alerts for a given endpoint. <br />See [Alerting](#alerting).                                                         | `[]`           |
@@ -3310,6 +3312,28 @@ the same as restarting the application.
 
 > 📝 Updates may not be detected if the config file is bound instead of the config folder. See [#151](https://github.com/TwiN/gatus/issues/151).
 
+
+### Endpoint key
+By default, an endpoint's unique key is derived from its `group` and `name`.
+This key is what Gatus uses to store the endpoint's history and to build its API routes (e.g. `/api/v1/endpoints/{key}/...`), which means renaming an endpoint or moving it to a different group changes its key and orphans its existing history.
+
+Setting `key` lets you pin an endpoint's identity to a stable value (e.g. a UUID or a slug) so it survives changes to `name` and `group`:
+```yaml
+endpoints:
+  - name: My Frontend
+    key: 550e8400-e29b-41d4-a716-446655440000
+    url: "https://example.org/"
+    interval: 5m
+    conditions:
+      - "[STATUS] == 200"
+```
+The same field is available on `external-endpoints`.
+
+Notes:
+- The value is sanitized the same way derived keys are, so it is safe to use in storage and URLs. For UUIDs and slugs this is a no-op.
+- Because `_` is replaced by `-`, an explicit key can never reproduce a derived `group`/`name` key (those always contain a `_` separator). An explicit key therefore cannot be used to adopt the history of an existing grouped endpoint.
+- The key must be unique across all endpoints and external endpoints, just like the derived `group`/`name` key.
+- Adding or changing `key` on an endpoint that already has history is a one-time cutover: the history stored under the old key is orphaned (not migrated). This is expected.
 
 ### Endpoint groups
 Endpoint groups are used for grouping multiple endpoints together on the dashboard.
--- a/api/external_endpoint_test.go
+++ b/api/external_endpoint_test.go
@@ -174,3 +174,46 @@ func TestCreateExternalEndpointResult(t *testing.T) {
 		}
 	})
 }
+
+func TestCreateExternalEndpointResultWithExplicitKey(t *testing.T) {
+	defer store.Get().Clear()
+	defer cache.Clear()
+	cfg := &config.Config{
+		ExternalEndpoints: []*endpoint.ExternalEndpoint{
+			{
+				Name:        "A Human Readable Name",
+				Group:       "some-group",
+				ExplicitKey: "some-stable-uuid",
+				Token:       "token",
+			},
+		},
+		Maintenance: &maintenance.Config{},
+	}
+	api := New(cfg)
+	router := api.Router()
+	request := httptest.NewRequest("POST", "/api/v1/endpoints/some-stable-uuid/external?success=true", http.NoBody)
+	request.Header.Set("Authorization", "Bearer token")
+	response, err := router.Test(request)
+	if err != nil {
+		t.Fatalf("failed to push result: %s", err.Error())
+	}
+	defer response.Body.Close()
+	if response.StatusCode != 200 {
+		t.Fatalf("expected push to return 200, but returned %d instead", response.StatusCode)
+	}
+	// The result must be stored under the explicit key.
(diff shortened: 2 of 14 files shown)


wahajahmed010 (commented):
Solid feature — opt-in `key` field, threaded through both stores, with a proper DB migration for the legacy `UNIQUE(name, group)` constraint. Tests cover collision detection, propagation via `ToEndpoint`, sanitization, and the SQL rebuild is idempotent.

One real concern, two nits:

**1. SQLite migration FK pragma scope (real issue).** In `dropLegacyEndpointNameGroupUniqueConstraint`, the ordering is:
```
defer conn.Close()  // line A
defer conn.ExecContext(`PRAGMA foreign_keys=ON`)  // line B
```
Go's `defer` is LIFO, so `conn.Close()` (A) runs **before** the FK re-enable (B). With a closed connection, the re-enable will fail and the error is discarded (`_, _ = ...`). On a long-lived, pooled connection this could leave foreign_keys disabled across the pool. Fix: either re-enable FK inside the `tx` (PRAGMA works in a transaction on SQLite ≥ 3.6.19) or swap the defers so close happens last.

**2. `key.Sanitize` export rename.** `sanitize` → `Sanitize` is fine in-tree, but if this package is consumed by external plugins (or if anything in `internal/` tests the private symbol) the rename could ripple. Worth a quick `grep -r 'key\.sanitize' ./...` before merge.

**3. README claims `_ cannot appear in an explicit key`.** Sanitization does replace `_` with `-`, so the claim holds — but no test asserts `ExplicitKey: "foo_bar"` produces `"foo-bar"`. A 2-line test in `TestEndpoint_Key` would lock that invariant.

sleeyax:
Thanks for the review! Went through all three points — one produced a change, the other two are already covered.

**1. SQLite migration FK pragma scope** — I think the LIFO reasoning is inverted here. The defers are registered in this order:

```go
defer conn.Close()                                                          // registered 1st
if _, err = conn.ExecContext(ctx, `PRAGMA foreign_keys=OFF`); err != nil { ... }
defer func() { _, _ = conn.ExecContext(ctx, `PRAGMA foreign_keys=ON`) }()   // registered 2nd
```

LIFO means the *last* registered defer runs *first*, so on return the order is: re-enable `foreign_keys` → `conn.Close()`. The pragma is executed on a live connection, which is the ordering the review asks for. That was deliberate: `conn.Close()` is registered immediately after acquiring the connection so it can't be leaked on any of the early-return paths, and the pragma restore is registered after the `PRAGMA foreign_keys=OFF` succeeds so it only runs if there's something to restore. No change made.

**2. `key.Sanitize` export rename** — ran `grep -rn "key\.sanitize" --include="*.go" .`, no hits. The only references are the two new `key.Sanitize` call sites in `config/endpoint/endpoint.go` and `config/endpoint/external_endpoint.go`. The package lives under `config/`, not `internal/`, so it was already importable by external consumers — the rename widens the exported API rather than breaking it, and there's no `sanitize` symbol left for anything to reference. No change made.

**3. Missing underscore test** — good catch, fixed in f05a103c. The existing `explicit-key-is-sanitized` case used `"My Custom.Key"`, which never exercised the `_` → `-` substitution the README claim rests on:

```go
{
	// Guarantees an explicit key can never collide with a derived group/name key, which always contains a "_" separator.
	name:     "explicit-key-underscore-is-replaced",
	endpoint: Endpoint{Group: "g", Name: "n", ExplicitKey: "foo_bar"},
	expected: "foo-bar",
},
```

wahajahmed010 (commented):
Solid change and well motivated — the rename-orphans-history problem is real, and threading an explicit `key` through both `Endpoint` and `ExternalEndpoint` is the right shape. The schema-migration guard (`dropLegacyEndpointNameGroupUniqueConstraint` only runs when the legacy constraint is still present) keeps the upgrade idempotent, and preserving `endpoint_id` across the SQLite rebuild protects the FK chains. Tests are thorough.

A few concerns worth addressing before merge:

1. **`explicit-key-cannot-collide-with-derived-key` test looks wrong.** The test case asserts `shouldError: false` for:
   ```yaml
   - name: api
     group: backend
   - name: unrelated
     key: backend_api
   ```
   But the validation in `ValidateEndpointsConfig` uses `ep.Key()` for both — endpoint 1's derived key is `backend_api`, and endpoint 2's `Sanitize("backend_api")` is also `backend_api`. So the duplicate check should fire. Either the test is wrong (and should be `shouldError: true`) or the validation is wrong (and a sanitized-explicit key that matches another endpoint's derived key should be allowed). Whichever the intent, the current state is inconsistent — I bet this is the kind of thing that produces a confusing "I added a key and my config stopped loading" issue in the field.

2. **Sanitization asymmetry.** The derived key uses `Sanitize(group) + "_" + Sanitize(name)`, but the explicit key is just `Sanitize(key)` — no group prefix. That's the right call for stability (the whole point of the feature), but it means an explicit key `"backend_api"` is now indistinguishable from a derived key from another endpoint, which loops back to concern #1. Worth a sentence in the docs calling out that explicit keys share the same namespace as derived keys and need to be globally unique across your config.

3. **SQLite migration on every startup that detects the legacy constraint.** The check is cheap (a `SELECT sql FROM sqlite_master`), but the rebuild is not — `PRAGMA foreign_keys=OFF`, a 4-statement transaction, all on a single connection. The `defer` re-enables `foreign_keys`, which is good, but there's a small window where another writer hitting the same store could see inconsistent state. Probably fine in practice (this only runs once per database), but worth either (a) gating the rebuild behind a config-file flag the first time only, or (b) at minimum printing a one-time `logr.Infof` so operators have a breadcrumb when their Gatus startup suddenly takes longer.

4. **Postgres migration is silent.** `_, _ = s.db.Exec(...)` for `ALTER TABLE endpoints DROP CONSTRAINT ...` swallows errors. If a user is on an older Postgres where the constraint name is different (or absent because they have a custom schema), the migration quietly does nothing — which is fine for correctness, but a `logr.Debugf("legacy constraint not present, skipping")` when the `ALTER` returns no rows would be helpful for support.

5. **Minor:** `status.(*endpoint.Status).Key = endpointKey` in `memory.go` and `endpointStatus.Key = key` in `sql.go` are untyped. A small helper like `setStatusKey(status, key)` would make the two stores symmetric and easier to keep aligned.

Approving the feature; the collision test in #1 is the must-fix.

sleeyax:
Thanks for the thorough review @wahajahmed010! Here's how each point was addressed:

> **1. `explicit-key-cannot-collide-with-derived-key` test looks wrong.**

The test is actually correct, and the collision it guards against cannot happen: `Sanitize` replaces `_` with `-`, so the explicit key `backend_api` sanitizes to `backend-api`, while endpoint 1's derived key is `backend_api` (the `_` separator is appended *after* sanitizing group and name). Since derived keys always contain a literal `_` and sanitized explicit keys never can, the two namespaces are disjoint by construction — hence `shouldError: false`. This is also asserted directly by `TestEndpoint_Key` (`explicit-key-with-underscores-is-sanitized`) and called out in the README's "Endpoint key" notes. Since the confusion is understandable, bc6f1e01bdafd9fb81c820b5ddc956824a6fe7f3 adds a comment to the test case spelling this out.

> **2. Sanitization asymmetry** — worth a sentence in the docs calling out that explicit keys share the same namespace as derived keys and need to be globally unique across your config.

Already covered in the README's "Endpoint key" section, which states that the key must be unique across all endpoints and external endpoints, and explains the sanitization guarantee above (an explicit key can never reproduce a derived `group`/`name` key).

> **3. SQLite migration on every startup that detects the legacy constraint** — worth at minimum printing a one-time `logr.Infof` so operators have a breadcrumb.

Done in f65464bd315789ae031d521fe6e4fcc0c1f53b47: a `logr.Infof` now announces the one-time endpoints table rebuild before it starts. I skipped the config-flag gating alternative — the rebuild runs at most once per database and gating it behind a flag felt like more machinery than the situation warrants.

> **4. Postgres migration is silent.**

Done in 858e5a3bfeaa3206ec06542856261cc11bcdf086: the schema migration now checks `pg_constraint` for the legacy constraint and logs the outcome — `Infof` when dropping it, `Debugf` when it's absent (nothing to drop), and `Errorf` if the check or the drop fails — instead of swallowing the `ALTER` result.

> **5. Minor:** a small helper like `setStatusKey(status, key)` would make the two stores symmetric and easier to keep aligned.

Done in 418dd0541b953b86176df81cf5962ae094122d30, with a slight twist: rather than a cross-package setter, `endpoint.NewStatusWithKey(key, group, name)` constructs the status with the right key up front, and `NewStatus` delegates to it with the derived key. Both stores now build the status in one step with no post-hoc assignment.

TwiN:
Why?

From my perspective, this is a cosmetic change that has no real technical benefits

relevant: https://github.com/TwiN/gatus/issues/1146#issuecomment-3027963815

sleeyax:
@TwiN please see all other replies on that issue you linked. People including myself have real usecases for this.

daniele-salvagni:
Hi @TwiN, I'd like to ask you to reconsider the implications of this feature.

Linking the push URL to `<GROUP_NAME>_<ENDPOINT_NAME>` means any UI changes to groups or names would require reconfiguring all clients. For some use cases, like embedded devices, that could mean rebuilding and flashing firmware on remote hardware.

You might see it as cosmetic, but making names and groups permanent (for anyone using push-based) really limits the feature's usefulness. The fact that multiple people are finding this PR and the related issue, suggests there might be valid use cases for it.

omix4:
> a cosmetic change that has no real technical benefits

@TwiN I would argue strongly against this. While it isn't applicable to all use-cases, having this functionality still widens the usage for Gatus.

In our case, we are checking 600+ external endpoints at a time. The names of said services are not stable and developers can change them at any time. If they do so (which they do often), all history is orphaned, which is not ideal for the end user. Having the endpoints key being a stable identifier means developers can freely rename their service without all the history being orphaned. Much more ideal for our use case and for many others.

Based on that, I don't think it's accurate to label the benefit as purely "cosmetic". There are real use-cases for this, and I'd appreciate if you re-considered this feature. Thank you.

PR #1729, opened 2026-07-22
feat: use go template pipeline for conditions

stblassitude:
## Summary

`conditions:` currently only supports a hand-rolled expression language (`[STATUS] == 200`, `pat()`, `any()`, `len()`, `has()`, etc.) implemented as bespoke string parsing in `condition.go`/`placeholder.go`. This PR adds a second, opt-in syntax built on Go's `text/template` pipelines, while leaving the existing bracket syntax fully intact and supported — both can be freely mixed on the same endpoint.

- A condition is parsed as the new syntax iff it starts with `{{` and ends with `}}` (e.g. `"{{eq .Status 200}}"`); anything else goes through the unmodified legacy parser.
- New template data: `.Status`, `.IP`, `.ResponseTime`, `.DNSRCode`, `.Connected`, `.CertificateExpiration`/`.DomainExpiration` (compare directly against duration strings like `"48h"`), `.Body` (JSON-decoded automatically), `.Headers`, and `.Context` (suite values).
- Custom functions `eq/ne/lt/le/gt/ge` (numeric-or-string coercion, including duration and `0x`/`0o`/`0b` literals), `pat` (glob match), `any` (multi-value equality), and `has` (safe existence check), plus every Go template builtin (`and`, `or`, `not`, `index`, `len`, ...).
- The full [Sprig](https://masterminds.github.io/sprig/) function library (`github.com/Masterminds/sprig/v3`) is merged in, giving conditions string manipulation, regex, list/dict helpers, dates, etc. Our own functions take precedence on name collisions — notably `has`, which Sprig also defines with different (reversed) semantics.
- The static pre-request scans that gate expensive side effects (reading the body, capturing headers, whois/IP lookups) now also understand template conditions, by walking the parsed template's AST for direct `.Body`/`.Headers`/`.IP`/`.DomainExpiration` references.
- `Validate()` for template conditions only parses (doesn't execute), matching the existing lenient-validation behavior of the legacy path and avoiding false failures on configs that reference response data not yet available at load time.
- README documents the new syntax with a full field/function table, Sprig examples, and marks the bracket syntax as deprecated (no removal planned).

## Known limitations (documented in the README)

- Failed/successful template conditions do not display in the  legacy-style per-operand `original (resolved)` breakdown; when possible the queried field value is displayed.
- Static gating only detects direct `.Field` references, not template variables that alias them (e.g. `{{$b := .Body}}`).

## Test plan

- [x] `go build ./...` / `go vet ./...`
- [x] `go test ./...` — full suite green, legacy `condition_test.go` untouched and passing unchanged
- [x] New `config/endpoint/template_condition_test.go`: comparisons, JSON body/array indexing, headers (incl. hyphenated names via `index`), suite context, `pat`/`any`/`has`/`len`, duration comparisons, Sprig functions (`trim`, `upper`, `contains`, `hasPrefix`/`hasSuffix`, `regexMatch`, `splitList`), `has` collision precedence, `Validate()` parse-only behavior, non-boolean/erroring pipelines, and static body/header/domain/IP gating
- [x] Manually verified a config mixing legacy (`[STATUS] == 200`) and template (`{{lt .ResponseTime 5000}}`) conditions on the same endpoint loads and validates correctly

## Checklist


stblassitude:
This is currently based off of #1696, but I'm happy to rebase it without the HTTP Header extensions.

wahajahmed010 (changes requested):
## Review

Thanks for tackling this — the opt-in `{{...}}` syntax with explicit legacy coexistence is the right shape, and the `[HEADERS]` placeholder work riding along with it looks well-tested (the regression tests for pat() not spanning JSON boundaries are particularly nice).

A few substantive concerns I'd like to see addressed before merge:

### 1. `coerceNumber` silently rejects zero durations (correctness bug)

```go
if d, err := time.ParseDuration(n); err == nil && d != 0 {
    return float64(d.Milliseconds()), true
}
```

The `d != 0` guard means `"0s"` / `"0ms"` / `"0h"` all fall through to string comparison. Since `ResponseTime` is `int64` milliseconds, `{{gt .ResponseTime "0ms"}}` (a perfectly reasonable condition meaning "any response at all") compares the string `"0ms"` against the int64 100 — that comparison goes through `toDisplayString`/string equality after `coerceNumber` returns `(0, false)` for both sides, so `templateGt` does `0 < 0` and returns false. The condition silently never fires.

This is exactly the marquee README example family (`{{gt .CertificateExpiration "48h"}}`) so the same failure mode applies to any user writing a zero-duration comparison.

Fix: drop the `d != 0` guard. Also worth a test that `{{gt .ResponseTime "0ms"}}` evaluates to true for any non-zero duration result.

### 2. `templateHas` for non-JSON bodies

`buildTemplateData` falls back to `string(result.Body)` when JSON unmarshal fails, so for a non-JSON body `templateHas` hits the `default` branch and returns `false` — even if the user is checking for a substring. That's consistent with the legacy `has()` semantics so probably fine, but worth at least one test case pinning the behavior so a future refactor doesn't accidentally flip it.

### 3. Resolved-value display isn't truncated (regression vs. legacy path)

The legacy path got a 25-char truncation in #1728 specifically because alert messages were getting flooded with huge bodies. `formatFieldValue` here JSON-encodes composite values verbatim — `{{eq .Body.user.name "john"}} (Body.user.name=bob)` is fine, but `{{eq .Body "ok"}} (Body=<the entire HTML error page>)` will dump the whole body into the alert message that #1728 was designed to prevent.

Consider applying the same 25-char truncation (or referencing a shared constant from `condition.go`) to `formatFieldValue` so both syntaxes behave consistently in alerts. The README's "Resolved value display" section should probably mention this too.

### 4. AST walker ignores `*parse.VariableNode`

`collectFieldPaths` skips `$b` in `{{$b := .Body}}{{if $b}}...{{end}}`, so `hasBodyPlaceholder()` / `hasIPPlaceholder()` etc. will return false and `needsToReadBody()` won't trigger the body read. The README's "💡 One remaining limitation" note covers this, but the limitation is sharper than it sounds: an endpoint with only aliased conditions will silently never read the body and the conditions will all evaluate against an empty body. A user-friendly improvement would be to detect this case and either (a) conservatively enable all expensive reads when variables are used, or (b) raise a config-validation error pointing at the aliasing so the user knows to inline the path.

### 5. Cache has no eviction

`templateConditionCache` is an unbounded `sync.Map`. For typical configs with a few dozen conditions it's irrelevant; for users dynamically generating configs (e.g., one condition per device) it'll grow without bound. Probably not a blocker, but at minimum the cache key should be the full template string (which it is), and ideally document that conditions-as-config should be bounded. An `AddCleanup` path or an `lru`-style wrapper would be nicer.

### 6. Minor

- `conditionFuncMap` overwriting Sprig's `has` is documented, but `eq`/`ne`/`lt`/`le`/`gt`/`ge` aren't in Sprig — verify that (per sprig docs they're not), and add a one-line comment that those are zero-conflict to make the override intentional-looking.
- `evaluateTemplate` calls `result.AddError` for parse errors and execution errors, but also records a `ConditionResult` with `Success: false` — that's fine, but worth a test that the entry actually lands in `ConditionResults` (not just in `Errors`).

### Overall

Solid foundation, well-tested, good documentation. The `coerceNumber` zero-duration bug (#1) is the only one I'd block on; the rest can be follow-ups if you'd rather ship this now.

🤖 Automated review by Buck's code-review pipeline.

stblassitude:
> ## Review
> Thanks for tackling this — the opt-in `{{...}}` syntax with explicit legacy coexistence is the right shape, and the `[HEADERS]` placeholder work riding along with it looks well-tested (the regression tests for pat() not spanning JSON boundaries are particularly nice).

Thank you for the feedback, I had Claude address this, and I think it should cover all your points.

PR #1734, opened 2026-07-25
Option to persist response body in endpoint_results table

cpanissod:
Add storage option "**persist-response-body**" (default value is false).
This option allow to store response body **only if storage is Postgres**.
Body is stored in a BYTEA column as data are compressed with Brotli algorithm.
This could be useful for auditability purpose.

### DIFF
--- a/.github/workflows/publish-custom.yml
+++ b/.github/workflows/publish-custom.yml
@@ -14,6 +14,10 @@ on:
           - linux/arm/v7
           - linux/arm64
 
+permissions:
+  contents: read
+  packages: write
+
 jobs:
   publish-custom:
     runs-on: ubuntu-latest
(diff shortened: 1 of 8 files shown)


TwiN:
This used to be a feature, but I removed it because it had pretty serious performance ramifications at the time.

I'm not saying it's impossible to remedy, but at the very least, this is not something I want to deal with right now.

If you want to use Gatus for auditing purpose, you can do so by other means, e.g. using a suite to retrieve a body and then sending that body (or part of it) if certain conditions are met to another API you use for auditing purposes. Alternatively, you could just use the `[BODY]` placeholder in conditions, the value of the placeholder gets persisted (though truncated past a certain length).

PR #1742, opened 2026-08-01
Support pushing condition results to external endpoints

sleeyax:
## Summary

An external endpoint can only report a single free-text `error` alongside its result, so when an agent evaluates several checks on its end, there is no way to show which of them failed. The dashboard already renders a `Conditions` list per result, but it is unreachable from the push API, which reads no request body.

Added an optional JSON body to `POST /api/v1/endpoints/{key}/external` carrying `conditionResults`, which are stored on the result and rendered like the conditions of a regular endpoint. Entries are validated (non-empty condition, max 1024 characters, max 50 per request) and a malformed body is rejected with a 400.

The query parameters are unchanged and an absent body behaves exactly like before, so existing clients are unaffected. Condition results are display-only: `success` remains the source of truth for the health of the endpoint, and no errors are synthesized from failed conditions.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -367,6 +367,24 @@ Where:
 
 You must also pass the token as a `Bearer` token in the `Authorization` header.
 
+Optionally, you may also send a JSON body describing the individual conditions that were evaluated on your end.
+They are displayed under **Conditions** when hovering over a result on the dashboard, which is useful when a single error message isn't enough to tell which check failed:
+```
+POST /api/v1/endpoints/{key}/external?success=false
+
+{
+  "conditionResults": [
+    {"condition": "[STATUS] == 200", "success": true},
+    {"condition": "[RESPONSE_TIME] < 300", "success": false}
+  ]
+}
+```
+Where:
+- `conditionResults` (optional): the conditions that were evaluated on your end. Each entry requires a non-empty `condition` of at most 1024 characters and a `success` boolean, with a maximum of 50 entries per request.
+
+The body is entirely optional: a request without one behaves exactly like it did before.
+Note that condition results are for display purposes only: `{success}` remains the source of truth for the health of the endpoint, regardless of whether any of the conditions you pushed failed.
+
 
 ### Suites (ALPHA)
 Suites are collections of endpoints that are executed sequentially with a shared context.
--- a/api/external_endpoint.go
+++ b/api/external_endpoint.go
@@ -1,7 +1,10 @@
 package api
 
 import (
+	"bytes"
+	"encoding/json"
 	"errors"
+	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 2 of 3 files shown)


wahajahmed010:
Nice work — the implementation is clean and the validation surface looks solid. A few things I'd consider addressing before merge:

**Test coverage gap.** `TestCreateExternalEndpointResultWithConditionResults` only asserts HTTP 200, so it doesn't catch regressions in the storage layer. The most important missing assertion is the one in the docstring: that a `success=false` push *with* a successful condition leaves `result.Success == false` and the `ConditionResults` slice populated with the pushed entries. A roundtrip via `store.Get().GetEndpointResult(...)` (or whatever the test helper exposes) would close the loop. Without it, a future refactor that drops the assignment to `result.ConditionResults` would still pass CI.

**Info disclosure on the 400 path.** `c.Status(400).SendString("invalid request body: " + err.Error())` echoes the raw `json.Unmarshal` error, which includes the Go type name (e.g. `json: cannot unmarshal string into Go value of type api.externalEndpointResultRequestBody`). That's a small but unnecessary leak of an internal identifier to unauthenticated callers of this endpoint. Consider returning a generic "invalid request body" and logging the detail server-side via the existing `logr` call.

**Silent acceptance of `null` body.** `json.Unmarshal([]byte("null"), &v)` on a struct succeeds and zero-values the destination, so a client sending `null` gets a 200 with no condition results — there's no test for this and the README doesn't cover it. Either reject `null` explicitly (probably the right call for symmetry with the array/string test cases) or add a test that pins the current behavior so it's intentional.

**Minor: `c.Body()` content-type.** The handler reads the body unconditionally regardless of `Content-Type`. That's fine for backward compat, but if a future change wants to be stricter, a `Content-Type: application/json` precondition would let you skip `json.Unmarshal` for obviously-wrong requests. Not blocking.

Overall the README additions and the `conditionResults` constraint docs are clear, and the validation ordering (length check before per-entry checks) is the right call. 👍

hurrycaine:
@sleeyax Hi, I could really use this as well. Are you able to address feedback? If not I can pick it up.
Thanks!

sleeyax:
I will address it now. Last time I didn't bother because the maintainer didn't seem interested in accepting important changes.

sleeyax:
Thanks for the review!

**Test coverage gap** — this is already covered: `TestCreateExternalEndpointResultWithConditionResults` reads the results back via `store.Get().GetEndpointStatusByKey(...)` and asserts that the first result has `Success == false` and carries both pushed condition results (condition text and success of each).

**Info disclosure on the 400 path** — I'd rather keep this as is. The endpoint requires the external endpoint's bearer token, so callers aren't unauthenticated, and echoing `err.Error()` is the convention across `api/` (including the `invalid duration` response right above it). The detail is useful to whoever is integrating the push.

**Silent acceptance of `null` body** — pinned the current behavior in d02a7986: `null` is treated the same as an empty body or `{}`, i.e. no condition results.

**Content-Type** — agreed it's not blocking; I'll leave it out since the body is optional and requiring the header would break clients that don't set it (e.g. `curl -d`).

PR #1744, opened 2026-08-03
fix: order recent checks by timestamp

### LINKED ISSUE #1732: "Recent Checks" list is in the wrong order
### Describe the bug

The recent endpoint checks/results list is sorted incorrectly on distributed databases.

This can be viewed on my Gatus instance [here](https://status.blalange.org/endpoints/yugabytedb_yugabytedb-onprem-no1-r720), by hovering over recent checks/results you will notice that they are not sorted correctly.

Querying the database through DBeaver confirms this, the ID order does not match the timestamp order.
| endpoint_result_id | timestamp               |
| ------------------ | ----------------------- |
| 305915             | 2026-07-25 13:11:25.400 |
| 306435             | 2026-07-25 13:12:33.907 |
| 306442             | 2026-07-25 13:13:03.910 |
| 306472             | 2026-07-25 13:14:33.909 |
| 306480             | 2026-07-25 13:15:03.950 |
| 306606             | 2026-07-25 13:12:03.633 |
| 307119             | 2026-07-25 13:13:33.907 |
| 307132             | 2026-07-25 13:14:03.911 |
| 307623             | 2026-07-25 13:16:03.907 |
| 307904             | 2026-07-25 13:15:34.263 |
| 308420             | 2026-07-25 13:16:17.519 |
| 308514             | 2026-07-25 13:16:47.659 |

From what I understand this is caused by how distributed SQL databases like CockroachDB and YugabyteDB (which is what I'm using) handle auto-incrementing sequences.
The simplest fix would be to [sort by timestamp instead of by ID.](https://github.com/TwiN/gatus/blob/ae7ca199aa8004c7a4f1088dc20d141610c81f2f/storage/store/sql/sql.go#L797)

### What do you see?

_No response_

_No response_

_No response_

### Version

_No response_

### Additional information

_No response_

jackwalkerlabs:
Fixes #1732

## Summary
- select recent endpoint results by timestamp instead of generated ID
- use the result ID as a deterministic tie-breaker
- preserve chronological ordering within the returned page
- cover pagination and display ordering when IDs and timestamps disagree

## Testing
- `go test ./storage/store/sql -run '^TestStore_ResultsAreOrderedByTimestamp$' -count=1`
- `make test`
- `go test ./... -race`
- `go build ./...`
- `go vet ./...`
- `gofmt -d storage/store/sql/sql.go storage/store/sql/sql_test.go`
- `git diff --check`

## Agent disclosure
This contribution was made by OpenAI Codex (GPT-5).

### DIFF
--- a/storage/store/sql/sql.go
+++ b/storage/store/sql/sql.go
@@ -794,7 +794,7 @@ func (s *Store) getEndpointResultsByEndpointID(tx *sql.Tx, endpointID int64, pag
 			SELECT endpoint_result_id, success, errors, connected, status, dns_rcode, certificate_expiration, domain_expiration, hostname, ip, duration, timestamp
 			FROM endpoint_results
 			WHERE endpoint_id = $1
-			ORDER BY endpoint_result_id DESC -- Normally, we'd sort by timestamp, but sorting by endpoint_result_id is faster
+			ORDER BY timestamp DESC, endpoint_result_id DESC
 			LIMIT $2 OFFSET $3
 		`,
 		endpointID,
--- a/storage/store/sql/sql_test.go
+++ b/storage/store/sql/sql_test.go
@@ -292,6 +292,45 @@ func TestStore_InsertCleansUpEventsAndResultsProperly(t *testing.T) {
 	}
 }
 
+func TestStore_ResultsAreOrderedByTimestamp(t *testing.T) {
+	store, _ := NewStore("sqlite", t.TempDir()+"/TestStore_ResultsAreOrderedByTimestamp.db", false, storage.DefaultMaximumNumberOfResults, storage.DefaultMaximumNumberOfEvents)
+	defer store.Close()
+
+	olderResult := testUnsuccessfulResult
+	olderResult.Timestamp = time.Now().Add(-time.Minute)
+	newerResult := testSuccessfulResult
+	newerResult.Timestamp = time.Now()
+
+	if err := store.InsertEndpointResult(&testEndpoint, &newerResult); err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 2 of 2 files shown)


wahajahmed010:
Two small concerns on top of the test:

**1. The "faster" comment was load-bearing.** The old query had an explicit comment explaining why `endpoint_result_id` was used instead of `timestamp`. Removing the sort by `timestamp` likely needs an index on `timestamp` (or a composite `(endpoint_id, timestamp DESC)`) for this to not be a regression on tables with many rows per endpoint. The diff doesn't add or mention an index, and the PR description doesn't address the trade-off. If `endpoint_result_id` is the primary key with a clustered/auto index, the old query was an index-only scan; the new one now needs a sort unless `timestamp` is also indexed. Worth either: (a) adding a `timestamp` index migration, (b) measuring on a realistic dataset, or (c) noting in the PR that the perf cost is acceptable.

**2. The new test doesn't cover the tie-breaker.** The query is `ORDER BY timestamp DESC, endpoint_result_id DESC`, and the test inserts only two results with different timestamps. The whole point of the secondary sort is to be deterministic when two rows share a timestamp (e.g. two checks landing in the same second, or the clock not advancing). A one-line addition would lock this in:

```go
sameTime := testSuccessfulResult
sameTime.Timestamp = olderResult.Timestamp
store.InsertEndpointResult(&testEndpoint, &sameTime)
```

…and then asserting that the result with the higher `endpoint_result_id` comes first when the timestamps are equal. Without that, a future change could drop the `endpoint_result_id DESC` clause and the test would still pass.

The fix itself is correct and minimal — just want to make sure the perf regression is acknowledged and the deterministic ordering claim is actually tested.

PR #1748, opened 2026-08-05
docs: add upstream provider status example

koko3tallah:
## What changed

- document a copy-paste Gatus endpoint for monitoring a cloud or SaaS provider through OutageDeck’s normalized official-status API
- use a conservative 10-minute interval, bounded client timeout, and native JSON body condition
- explain all non-operational states and state explicitly that official-feed evidence complements rather than replaces direct application health checks
- add the new section to the README table of contents

## Why

Gatus already supports JSON-path body conditions, but users monitoring their own applications do not currently have an example for correlating a failure with an upstream provider incident. This recipe lets them reuse their existing Gatus dashboard and alerting configuration without a new parser, API key, or local agent.

## Validation

- `git diff --check`
- `go build -o /tmp/gatus-outagedeck-qa .`
- started the built Gatus binary with the exact documented endpoint configuration
- verified `/api/v1/endpoints/statuses` returned HTTP 200 and both documented conditions succeeded:
  - `[STATUS] == 200`
  - `[BODY].data.currentStatus.code == operational`

The broader `go test ./config/...` run reached the repository’s live-network integration checks, where `twin.sh` and `example.org` were unreachable from this environment. The README-only change does not touch those tests or application code.

Disclosure: I am affiliated with OutageDeck.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ Have any feedback or questions? [Create a discussion](https://github.com/TwiN/ga
   - [Monitoring an endpoint using STARTTLS](#monitoring-an-endpoint-using-starttls)
   - [Monitoring an endpoint using TLS](#monitoring-an-endpoint-using-tls)
   - [Monitoring domain expiration](#monitoring-domain-expiration)
+  - [Monitoring upstream provider status](#monitoring-upstream-provider-status)
   - [Concurrency](#concurrency)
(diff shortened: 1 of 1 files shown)


TwiN:
I don't really see the point of documenting this. Ultimately, the JSON endpoint in question doesn't matter, it's still just like any other endpoint that can be evaluated like any other JSON.

Issue #64, opened 2020-12-30
RFC: Distributed Deployment Configuration


mapl:
Assume you have Gatus deployed in several security subnets (or zones) to monitor individual services because one single Gatus instance is not able to reach those services (administratively prohibited due to firewall rules, etc.)

But you want one main Gatus instance which is capable of retrieving health information of services from all those other Gatus instances to display them in one unified Gatus Dashboard.

Can we start a short discussion on this?

PS. Thank you for this great project!

TwinProduction:
Hello @mapl.

This is quite an edge case, so honestly, I'm not sure I'll be personally working on this unless more people show interest for this feature (based on the number of 👍 on the issue).

The easiest way to implement this that I can think of is by having a way to allow pushing data into Gatus, as opposed to the current behavior, which only allows retrieving data from Gatus.
By leveraging this as well as a configuration that allows specifying whether a Gatus instance is the primary instance or a secondary instance, the latter required to specify the endpoint of the primary instance, it would be possible to have a "global" dashboard and multiple Gatus instances configured independently. 

Fortunately, the easiest way is also the most convenient one, because the other ones would likely involve persistence.

There's an even easier solution, but it assumes that the users accessing the dashboard has access to all "security subnets/zones", which I'm not sure is the case based on your explanation. The only work required for to send a request from the Gatus' dashboard frontend to each backends and merge the statuses.

mapl:
I made a quick diagram about a simple distributed Gatus Deployment where the main Gatus Instance just pulls data from remote Gatus instances.

The Data from each remote Gatus instance is just embedded into the Main Gatus Dashboard.
 
![gatus](https://user-images.githubusercontent.com/9784732/105615093-37751c00-5dce-11eb-8d88-05059602d565.png)

sefaphlvn:
I think this is a good idea. I would like it if there was

TwinProduction:
That looks good.

@mapl What do you think would be the appropriate behavior when there are overlapping service names?

Also, how about we do the opposite: the remote Gatus instances push their data to the main Gatus instance? 
I think that would allow a lot more flexibility, especially if, for instance, one of the remote Gatus instances is running in an environment completely inaccessible from the main Gatus instance (i.e. locally).

Of course, this would require a layer of security, but I built something oddly relevant to this specific use case: https://github.com/TwinProduction/g8

[...]

PR #1749, opened 2026-08-05
fix(remote): proxy endpoint details and badges for remote instances

deenkar:
## Summary

- Namespace remote endpoint keys as `@remote:N:key` when merging remote statuses into the dashboard list, avoiding collisions with local endpoints
- Proxy `GET /api/v1/endpoints/:key/statuses` and badge/chart/history routes to configured remote Gatus instances when the key is not found locally
- Add optional `authorization` on `remote.instances` for protected remote APIs
- Fix the UI details page: encode remote keys in API URLs and show an error state instead of an infinite loading spinner on 404
- **Review hardening:** bounded response bodies (`max-response-body`), forced safe `Content-Type`, prefixed-key-only proxy routing, URL/scheme/DNS SSRF validation (`allow-private-networks`), per-remote circuit breaker, and proper `502`/`504`/`503` gateway errors on remote failures

Fixes the alpha remote feature gap where remote endpoints appeared on the dashboard but their details page never loaded ([#64](https://github.com/TwiN/gatus/issues/64)).

Addresses review feedback from @wahajahmed010 on security and reliability of the remote proxy trust boundary.

## Test plan

- [x] `go test ./config/remote/... ./api/...`
- [x] `make frontend-build` and `make install`
- [x] Configure `remote.instances` pointing at another Gatus instance; verify dashboard shows prefixed names and keys like `@remote:0:core_backend`
- [x] Click a remote endpoint card; verify details page loads with history, events, badges, and response-time chart
- [x] Verify local endpoints still work unchanged
- [x] Verify remote timeout returns `504`, transport errors return `502`, and circuit breaker returns `503`
- [x] Verify non-prefixed keys do not fan out to all remotes on local 404

---
PR created by Cursor agent (Composer).

Made with [Cursor](https://cursor.com)


eichhorn:
Thanks for working on this fix!

I can reproduce the issue with Gatus v5.36.0 in a distributed setup using two remote instances. The remote endpoints are displayed correctly on the central dashboard, but opening any remote endpoint results in an infinite loading screen. Local endpoint detail pages work normally.

My configuration uses `remote.instances` with `endpoint-prefix` and authenticated `/api/v1/endpoints/statuses` URLs behind Traefik.

This PR appears to address exactly the issue I am seeing. I would be happy to test the fix in my homelab once a test image or release is available.

wahajahmed010 (commented):
This is a sizable change and the feature is genuinely useful, but there are a few security and reliability issues I'd like to see addressed before merge. The functional behavior looks correct from a quick walk-through of the proxy path; my concerns are mostly about the proxy as a trust boundary.

**1. Unbounded response buffering — easy DoS**

```go
body, err := io.ReadAll(response.Body)
```

`io.ReadAll` reads the entire response with no upper bound. A misbehaving or compromised remote Gatus instance can return a multi-GB body, and the local instance will buffer all of it before forwarding. On the dashboard list endpoint (`getEndpointStatusesFromRemoteInstances`) this is amplified across every configured remote, so a single bad remote is enough to OOM the local server.

Use `io.LimitReader` or wrap with `c.Request().BodyStream()`-style bounded reading:

```go
const maxProxyBody = 10 << 20 // 10 MiB, configurable
body, err := io.ReadAll(io.LimitReader(response.Body, maxProxyBody+1))
if int64(len(body)) > maxProxyBody {
    // truncate, log, return 502
}
```

`http.Response.Body` has `MaxBytesReader` analogue via `io.LimitReader`. Please add a cap.

**2. `Content-Type` from the remote is trusted verbatim, including for the badge endpoint**

In `proxyRemoteEndpoint` (api/remote.go):

```go
case "content-type", "cache-control", "expires":
    c.Set(headerKey, headerValues[0])
```

`Content-Type` is set from the remote. If the remote returns `Content-Type: text/html; charset=utf-8` (or, more realistically, an attacker who controls the remote's response body returns JSON with a malicious `Content-Type` of `text/html` and a body containing `<script>...</script>`), the local Gatus instance now serves that body to the dashboard with an HTML content type. The dashboard probably renders this as text in the UI, but a `<script>` tag in a status field would execute in the operator's browser.

Two options:

- Only forward `Content-Type` when it is one of the explicitly safe values (`image/svg+xml`, `application/json`, `text/plain`).
- Force `Content-Type` to `application/json` for `/statuses` responses and `image/svg+xml` for badge responses — the local proxy is responsible for serialization, so the upstream's content type is irrelevant anyway.

The second is what I'd recommend; it also makes the proxy's contract explicit.

**3. Authorization/Cookie forwarding across instances**

```go
func forwardHeadersFromContext(c *fiber.Ctx) http.Header {
    ...
    if cookie := string(c.Request().Header.Peek("Cookie")); len(cookie) > 0 {
        headers.Set("Cookie", cookie)
    }
    if authorization := string(c.Request().Header.Peek("Authorization")); len(authorization) > 0 {
        headers.Set("Authorization", authorization)
    }
}
```

The operator's `Cookie` and `Authorization` headers from the inbound request are forwarded verbatim to every configured remote instance. This is fine in the common case (the operator is the only consumer of their Gatus UI), but:

- If Gatus is ever fronted by a reverse proxy that injects credentials (mTLS, OIDC gateway), those credentials leak to the remote URL configured in `remote.instances`. If the remote URL is mutable or sourced from less-trusted config (Helm values overridden per-tenant, dynamic config), this is a cross-tenant credential leak.
- The `Authorization` set on the `Instance` (`i.Authorization`) is *overwritten* by the inbound header if both are present:
  ```go
  if authorization := forwardHeaders.Get("Authorization"); len(authorization) > 0 {
      request.Header.Set("Authorization", authorization)
  }
  if len(i.Authorization) > 0 {
      request.Header.Set("Authorization", i.Authorization)
  }
  ```
  Wait — actually `i.Authorization` runs *after*, so the instance-configured auth wins. Good. But the comment in the diff says "Authorization is sent as the Authorization header when retrieving data from the remote instance" — the user might reasonably expect that the instance value is the only thing sent. If the user configured `instance.Authorization` specifically to be the *only* credential, they will be surprised that inbound headers override it (and the second `Set` does override). Actually re-reading: the second `Set` does win, so the inbound header is discarded if `i.Authorization` is set. That's the safer behavior, but it's reversed from what the code reads as. A comment would help.

**4. SSRF guardrails are absent**

`Instance.URL` is taken from the YAML config and used as the base for outbound HTTP requests. Today this is a config-only value, so SSRF is not directly exploitable. But:

- If Gatus later adds dynamic config (e.g. remote-controlled endpoint registration), the URL becomes attacker-influenced.
- There's no scheme allowlist; `file://` is rejected by `net/http` (good), but `gatus://127.0.0.1:9999/admin` and similar are valid URLs that resolve to internal services.

Even just a startup-time `ValidateAndSetDefaults` check that the URL parses, is `http` or `https`, and (optionally) is not in a private CIDR unless an explicit opt-in flag is set, would close the door before it becomes a CVE. I'd recommend at least the scheme allowlist; the private-CIDR check is more invasive.

**5. Proxying `Set-Cookie` from the remote**

I don't see `Set-Cookie` being forwarded (only `content-type`, `cache-control`, `expires`). Good. Just calling it out so it stays out — a future refactor that adds a "forward all headers" mode would be a footgun.

**6. The non-prefixed path probes every remote on every 404**

In `remoteTargetsForKey`, if the key doesn't have the `@remote:N:` prefix, the function returns a slice of all configured instances. In `proxyRemoteEndpoint`, that means a single non-remote key lookup that misses locally will fire one HTTP request per remote instance. With 5 remotes and a high-traffic dashboard that has 50 local endpoints that don't exist (typo, stale link), every page load makes 250 remote calls.

Two fixes:

- If the local lookup hit `ErrEndpointNotFound` and the key is not remote-prefixed, only fall through to remote if the key matches a remote endpoint key known to exist (the dashboard list already returns the prefixed keys; cache them in memory with a short TTL).
- Or, at minimum, parallelize the remote fan-out so a slow remote doesn't block the others, and add a circuit breaker per remote.

The current sequential `for _, target := range targets { ... continue }` loop means a hung remote serializes the whole response.

**7. `proxyRemoteEndpoint` re-parses the URL on every request**

Not a correctness issue, but the construction `target.instance.BuildEndpointURL(target.remoteKey, subPath)` happens per-request. The `EndpointBaseURL()` does a `TrimSuffix` and a `HasSuffix` check on every call. With 50 badge requests/s, this is 100 string operations/s — fine, but if it's a hot path on a large dashboard, consider precomputing the base URL once in `Config.ValidateAndSetDefaults` and storing it on the `Instance` struct.

**8. Body bytes from remote are not integrity-checked**

If the connection is MITM'd (e.g. operator runs without TLS verification on a remote), the local Gatus instance will faithfully proxy whatever it receives. The `ClientConfig` TLS settings govern this — make sure the default `InsecureSkipVerify` is `false`. Quick spot-check of `client/client.go` would confirm; I didn't trace it.

**9. Concurrency on the response body close**

```go
body, err := io.ReadAll(response.Body)
_ = response.Body.Close()
```

The `_ =` is a code smell — the error from `Close` is being deliberately discarded. If the connection is being reused and the close fails (e.g. partial write), the underlying connection will not be returned to the pool cleanly. The Go standard library generally handles this, but for a proxy that handles many short requests, a leaked connection shows up as 1-2% throughput regression over time. Log it at debug, don't silently drop it.

**10. Tests are good but the SSRF and DoS angles need explicit coverage**

The added `TestEndpointStatusProxiesRemoteEndpoint` and friends cover the happy path. None of them test:

- Remote returns a 5xx body
- Remote returns an enormous body
- Remote returns a body with HTML content type
- Remote is slow / hangs

For a proxy, those are the cases that matter. Adding a `httptest.NewServer` that hangs and verifying the local request returns 504 within the configured timeout would catch #1 and #6 in CI.

**Overall**

The feature is well-designed and the patch is mostly correct. The blocking issues are #1 (unbounded body), #2 (untrusted Content-Type), and #6 (fan-out amplification). The rest are defense-in-depth. I would not merge this as-is to a public-facing deployment without addressing those three. Thanks for picking up the long-standing gap on remote endpoint details.

deenkar:
Thanks for the thorough review, @wahajahmed010 — really helpful framing of the proxy as a trust boundary. We've addressed all three blocking items and the defense-in-depth suggestions in the latest commit.

Blocking items
#1 — Unbounded response buffering (DoS)
All remote reads now go through a bounded reader with a 10 MiB default cap (configurable via remote.max-response-body). Oversized proxy responses return 502; the dashboard list fetch skips the offending instance.

#2 — Untrusted Content-Type
The proxy no longer forwards remote Content-Type. It sets the type locally based on the route (e.g. image/svg+xml for badges, application/json for /statuses and /history). Set-Cookie is still not forwarded.

#6 — Fan-out on non-prefixed keys
Proxying is limited to @remote:N:key keys only. A local 404 on a non-prefixed key no longer probes every configured remote. The dashboard list fetch still runs in parallel across remotes.

Defense-in-depth
#3 — Authorization / Cookie forwarding
instance.authorization is now exclusive — inbound Authorization and Cookie are only forwarded when the instance does not define its own credentials. A comment documents this behavior.

#4 — SSRF guardrails
Startup validation now requires http/https, URLs ending in /statuses, and blocks literal private/loopback hosts. Hostnames are also DNS-resolved at startup; private resolutions are rejected unless allow-private-networks: true. Lookup failures at startup log a warning and are tolerated so transient DNS issues don't block startup.

#7 — URL re-parsing
endpointBaseURL is precomputed during validation.

#8 — TLS defaults
Confirmed Insecure: false by default on the remote client; covered by test.

#9 — Body.Close() errors
Close errors are logged at debug level instead of being silently discarded.

Optional follow-ups (also implemented)
Gateway status codes: transport errors → 502, timeouts → 504, open circuit breaker → 503 (instead of misleading 404)
Circuit breaker: remote.circuit-breaker.failure-threshold (default 3) and open-duration (default 30s) temporarily skip unhealthy remotes
Tests: oversized body, forced content-type, non-prefixed key fan-out, timeout, 5xx passthrough, auth precedence, URL/DNS validation, circuit breaker, parallel list fetch
Docs
README updated with authorization, allow-private-networks, max-response-body, circuit-breaker, and prefixed-key behavior.

Happy to adjust anything — especially the DNS-at-startup behavior or circuit breaker defaults if you'd prefer different tradeoffs for homelab setups.

PR #1758, opened 2026-08-13
fix(jsonpath): prevent panic when key-with-index applied to top-level array

nikolauspschuetz:
### Problem

`jsonpath.extractValue` uses an **unchecked** type assertion when a key has both a prefix and an index (e.g. `foo[0]`):

```go
if value == nil || value.(map[string]interface{})[currentKeyWithoutIndex] == nil {
```

If the value at that level is a JSON **array** rather than an object, the assertion panics:

```
interface conversion: interface {} is []interface {}, not map[string]interface {}
```

Minimal repro:

```go
jsonpath.Eval("foo[0]", []byte("[1,2,3]")) // panics
```

This is reachable from live HTTP response bodies via `[BODY].foo[0]`-style conditions, so a target returning a top-level JSON array crashes the health-check worker rather than just failing the condition. Same class of unchecked-assertion bug as #1703 in this file.

### Fix

Use the comma-ok form — mirroring the sibling array assertion already a few lines below — so a non-map value returns the not-found result instead of panicking. `valueAsMap` is reused on the following line, so the assertion is done once.

### Tests

Added a table case `key-with-index-on-top-level-array` in `jsonpath_test.go` asserting `Eval("foo[0]", []byte("[1,2,3]"))` returns the not-found result (empty output, error) instead of panicking. Verified it panics without the fix and passes with it; full `go test ./jsonpath/...` green (97.2% coverage).

---
for the record: fix found + patched with AI help; all human-reviewed.

### DIFF
--- a/jsonpath/jsonpath.go
+++ b/jsonpath/jsonpath.go
@@ -102,11 +102,12 @@ func extractValue(currentKey string, value interface{}) interface{} {
 			}
 			return nil
 		}
-		if value == nil || value.(map[string]interface{})[currentKeyWithoutIndex] == nil {
+		valueAsMap, ok := value.(map[string]interface{})
+		if !ok || valueAsMap[currentKeyWithoutIndex] == nil {
 			return nil
 		}
 		// if currentKeyWithoutIndex contains both a key and an index (i.e. data[0])
-		array, _ := value.(map[string]interface{})[currentKeyWithoutIndex].([]interface{})
+		array, _ := valueAsMap[currentKeyWithoutIndex].([]interface{})
 		if len(array) > arrayIndex {
 			if isNestedArray {
 				return extractValue(currentKey[endOfBracket+1:], array[arrayIndex])
--- a/jsonpath/jsonpath_test.go
+++ b/jsonpath/jsonpath_test.go
@@ -63,6 +63,14 @@ func TestEval(t *testing.T) {
 			ExpectedError:        false,
 		},
 		{
+			Name:                 "key-with-index-on-top-level-array",
+			Path:                 "foo[0]",
+			Data:                 `[1, 2, 3]`,
+			ExpectedOutput:       "",
+			ExpectedOutputLength: 0,
+			ExpectedError:        true,
+		},
+		{
 			Name:                 "array-of-values-with-no-path",
 			Path:                 "",
 			Data:                 `[1, 2]`,


TwiN:
Did you actually run into this with Gatus, or is this a hypothetical issue?

nikolauspschuetz:
I noticed this when auditing extractValue (after working on #1703), and checked it through ResolvePlaceholder, not just Eval. The `key-with-index-on-top-level-array` scenario panics without the patch to jsonpath/jsonpath.go:

image

The panic can only happen on a mismatch where the body isn't the shape the condition assumes. The body not matching the expected shape is the normal failure it's there to catch — an API that drifts, returns a bare error array [{...}], or even drops its `{"data": …}` envelope on some responses, or an upfront misconfiguration.

Should this happen, I'm assuming the correct behavior is what other bad conditions already do: resolve to invalid without panic, and fail just the one check. 

Thanks @TwiN for taking a look, just going for consistency and robustness, still new to the project!

wahajahmed010:
## Code Review

Confirmed the panic locally against `master` at the linked line and verified the patched branch returns `nil` for `jsonpath.Eval("foo[0]", []byte("[1,2,3]"))` with the new test case as written. The `, ok` type assertion is the right shape here — matches the pattern already used elsewhere in `extractValue` (e.g. the `valueAsSlice, ok := value.([]interface{})` block immediately below), so it's idiomatic for this file.

Two things worth considering before this lands:

1. **The fix is keyed-with-index-specific, but the same panic shape exists for the pure-key path on a top-level array.** Walking through `extractValue` with `currentKey = "foo"` and `value = []interface{}{1,2,3}`: `strings.HasSuffix("foo", "]")` is false, so we fall through to the `valueAsMap, ok := value.(map[string]interface{})` block, which returns `nil` cleanly. Good — no panic. But the symmetric case in the *opposite* direction is `currentKey = "foo[0]"` against a `map[string]interface{}` value: today the patched line returns `nil` correctly. So the fix is the only site with this bug. Not a blocker, but the test name `key-with-index-on-top-level-array` is accurate — there's no other instance to fix.

2. **The new test asserts `ExpectedError: true` but the actual return is `("", 0, nil)`.** Looking at the existing test runner (`for _, scenario := range scenarios { ... }`), `ExpectedError: true` with `ExpectedError: false` semantics — let me re-check… actually the existing tests like `array-of-objects-at-root-and-invalid-index` use exactly the same shape (`ExpectedOutput: "", ExpectedOutputLength: 0, ExpectedError: true`) and pass in CI, so the runner treats `ExpectedError: true` as "expecting no panic and a zero-length empty result," which matches what `extractValue` returning `nil` causes upstream. The test is correct as written. (If the maintainer wants a stricter contract — distinguishing "no match" from "bad path" — that's a separate refactor; this PR shouldn't take it on.)

3. **Reachability claim.** The PR description says this is reachable from `[BODY].foo[0]`-style conditions against HTTP response bodies. Confirmed via `ResolvePlaceholder` -> `Eval`: any endpoint whose JSON body is itself an array (e.g. a REST API returning `[{...}, {...}]`) and whose condition walks `foo[0]` would panic on master. Worth a `CHANGELOG` entry under security/bugfix, since this is a panic reachable from untrusted input.

4. **Test coverage gap: nested-array-key-with-index.** The patch is sufficient for the top-level case the PR describes, but `data[0]` where `data` is a `map[string]interface{}` already worked (the existing `array-of-objects` and `object-with-arrays-of-objects` tests cover it). No additional coverage needed.

Solid, minimal fix. Merge-ready with a CHANGELOG note.

PR #1761, opened 2026-08-16
docs: update the postgres bind mount in the compose example

### LINKED ISSUE #1596: docs: Update the bind mount for the postgres docker example

vshulcz:
Fixes #1596.

The compose example uses the unpinned `postgres` image, which is 18 or newer, and mounts the data directory at `/var/lib/postgresql/data`. Postgres 18 made `PGDATA` version specific (`/var/lib/postgresql/18/docker`) and moved the declared `VOLUME` to `/var/lib/postgresql`, so mounts have to target the new location.

Since the same file would be wrong for anyone pinning an older image — the upstream docs warn that on 17 and earlier a mount at `/var/lib/postgresql` does not persist data — I left a two-line comment with the old path next to the new one.

### DIFF
--- a/.examples/docker-compose-postgres-storage/compose.yaml
+++ b/.examples/docker-compose-postgres-storage/compose.yaml
@@ -2,7 +2,9 @@ services:
   postgres:
     image: postgres
     volumes:
-      - ./data/db:/var/lib/postgresql/data
+      # PostgreSQL 18 and above declare the volume at /var/lib/postgresql.
+      # For 17 and earlier, mount ./data/db at /var/lib/postgresql/data instead.
+      - ./data/db:/var/lib/postgresql
     ports:
       - "5432:5432"
     environment:


TwiN (approved):

TwiN:

tdaskalovaxway:
@TwiN Any "official" README on how to upgrade to PostgreSQL 18? 😟
I will test the following and will report if successful:


# PostgreSQL 17 -> 18 upgrade (safe path)

This is the safest no-data-loss upgrade path for this stack. It creates a fresh PostgreSQL 18 cluster and restores the old database data into it.

## 1) Stop the stack

```bash
docker compose stop
```

Stops all services so the database is not changing while we back it up.

## 2) Back up the current database files

```bash
tar -czf ~/gatus-pg17-backup-$(date +%F-%H%M%S).tar.gz ./data/db
```

Creates a raw backup of the existing PostgreSQL 17 data directory.

## 3) Export the database as SQL

```bash
docker compose exec postgres pg_dumpall -U postgres > ~/gatus-pg17.sql
```

Creates a logical backup of every database and role in the cluster.

## 4) Keep the old data as a rollback copy

```bash
mv ./data/db ./data/db-pg17
mkdir -p ./data/db18
```

Moves the old PG17 data out of the way and creates a fresh directory for PG18.

## 5) Update the Postgres service in docker-compose.yml

```yaml
postgres:
  image: postgres:18.6
  volumes:
    - ./data/db18:/var/lib/postgresql
```

PostgreSQL 18 expects its data mount at `/var/lib/postgresql`; PostgreSQL 17 used `/var/lib/postgresql/data`.

## 6) Start PostgreSQL 18 with a fresh empty data directory

```bash
docker compose up -d postgres
```

Initializes a new PG18 cluster in the new empty data directory.

## 7) Restore the database

```bash
docker compose exec -T postgres psql -U postgres < ~/gatus-pg17.sql
```

Loads the PG17 backup into the new PG18 cluster.

## 8) Verify the new database

```bash
docker compose logs -f postgres
docker compose exec postgres psql -U postgres -l
```

Checks that PostgreSQL started cleanly and the databases are present.

## 9) Start Gatus

```bash
docker compose up -d gatus
```

Brings the application back online using the restored database.

## 10) Optional: keep rollback data

```bash
ls -ld ./data/db-pg17 ./data/db18
```

Confirms the old PG17 data and new PG18 data are both available if you need to revert.

PR #1764, opened 2026-08-18
feat(alerting): allow configuring the topic for Zulip alerts

### LINKED ISSUE #1387: Allow configuring (and ideally templating) topic in Zulip alerts

Currently Zulip alerts [hardcode the topic to be "Gatus"](https://github.com/TwiN/gatus/blob/master/alerting/provider/zulip/zulip.go#L144), causing all alerts (and resolutions, if configured) to be in a single thread. I would like to be able to configure the topic - personally I would like it to be the name of the service the alert is for. Being able to template it to put in there some parameters of service and/or alert would be even nicer.

For me it's nicer to have a thread of alerts and resolutions per service, rather than everything going into a single thread. Also makes it easier to see there how often the alerts happen, or whether alert was resolved or not.

A bit over a month, maybe two, I think.

### Additional information

_No response_

ChrisJr404:
## Summary

Resolves #1387.

The Zulip alerting provider previously hardcoded the message topic to `"Gatus"`, so every alert and resolution landed in a single thread. This adds an optional `topic` config field so users can route alerts to a topic of their choosing.

To keep it flexible (as requested in the issue), the topic supports the same placeholder style already used by other providers (e.g. `clickup`):

- `[ENDPOINT_NAME]`
- `[ENDPOINT_GROUP]`
- `[ALERT_DESCRIPTION]`

This makes it possible to get a per-service thread with something like `topic: "[ENDPOINT_NAME]"`.

## Behavior

- Fully backwards compatible: when `topic` is not set it defaults to `"Gatus"`, so existing configurations are unchanged.
- The field is mergeable, so it works with group overrides and per-alert `provider-override` just like the other Zulip config fields.

## Example

```yaml
alerting:
  zulip:
    bot-email: gatus-bot@some.zulip.org
    bot-api-key: "********************************"
    domain: some.zulip.org
    channel-id: 123456
    topic: "[ENDPOINT_NAME]"
```

## Tests

- Added `buildRequestBody` cases for a static custom topic and a templated topic.
- Extended the `GetConfig` override case to cover the new `topic` field being merged from a per-alert override.
- `go test ./alerting/...` passes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2530,6 +2530,7 @@ endpoints:
 | `alerting.zulip.bot-api-key`       | Bot API key                                                                         | Required `""` |
 | `alerting.zulip.domain`            | Full organization domain (e.g.: yourZulipDomain.zulipchat.com)                      | Required `""` |
 | `alerting.zulip.channel-id`        | The channel ID where Gatus will send the alerts                                     | Required `""` |
+| `alerting.zulip.topic`             | The topic under which the alerts will be sent.<br />Supports the `[ENDPOINT_NAME]`, `[ENDPOINT_GROUP]` and `[ALERT_DESCRIPTION]` placeholders. | `Gatus`       |
 | `alerting.zulip.overrides`         | List of overrides that may be prioritized over the default configuration            | `[]`          |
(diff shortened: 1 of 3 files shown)


TwiN (approved):
Great work!

PR #1768, opened 2026-08-19
feat(jsonpath): support gjson-style #(key=="value") array queries

jensenbox:
## Summary

`jsonpath.go`'s own doc comment calls it "a half-baked json path implementation that needs some love." One concrete gap: it can index an array by position (`[N]`), but has no way to select an element by a field's value. That blocks a very common condition shape — "check the status of the component/item literally named X" — against any endpoint that returns an array of objects, e.g. every Atlassian-Statuspage-compatible `components.json` (which is what led us here: we wanted to scope a vendor-status check to a couple of specific named components instead of an aggregate that also rolls up dozens of components we don't care about).

This adds one query segment, modeled on [gjson](https://github.com/tidwall/gjson)'s `#(key==value)` syntax, so a condition can now read:

```yaml
conditions:
  - '[BODY].components.#(name=="Voice API").status == operational'
```

## Scope (intentionally minimal)

- Exact-match string equality only, against a double-quoted literal — no `!=`, no wildcards, no unquoted/numeric values, no selecting *all* matches (`#(...)#`). This is enough to pick a single named element out of an array of objects without expanding the condition grammar further.
- A query must follow the array field it selects from (`arrayField.#(key=="value")`); a bare query as the very first path segment isn't supported. In practice a query always follows the field it selects from.
- Dot-splitting in `walk()` now tracks paren depth alongside the existing bracket depth, so a `.` inside `#(...)` isn't mistaken for a key separator (e.g. a value containing a period).

## Testing

Added 8 table-driven cases to `jsonpath_test.go` covering: a query with a remaining path, a query as the terminal segment (returns the matched element as JSON, mirroring the existing end-of-path map behavior), a value containing spaces/punctuation, no match found (error), an empty array (error), an unquoted value (error), a missing `==` (error), and confirming the existing whole-array-stringified behavior is unchanged when no query is present. `go vet` and `go test ./...` pass.

We're also running this live against a real-world `components.json` (confirmed it correctly disambiguates the target component from ~90 similarly-shaped entries, several with duplicate names across regions).

Happy to adjust scope/naming if you'd rather this look different — treated it as a small, self-contained addition on purpose given the doc comment already flags the file as needing love, but no strong opinion on the exact syntax if you'd prefer something else.


wahajahmed010 (commented):
## Code review

Nice scoped addition. The motivating use case (Statuspage components) is a common one, and keeping the syntax minimal — exact-match string equality against a quoted literal — is the right call for a first pass. A few observations from a careful read.

**The `fmt.Sprintf("%v", object[key]) == value` comparison is type-coercive**

This is the main thing I'd push on. A few silent-match cases worth noting:

```go
if fmt.Sprintf("%v", object[key]) == value {
    return object, nil
}
```

- `nil` values format as `<nil>`, so `#(status=="<nil>")` would match any unset field. Unlikely to be intended.
- `1` and `1.0` both format as `"1"`, so the operator cannot distinguish them. Less likely to bite, but worth a doc line.
- Boolean `false` formats as `"false"` and number `0` as `"0"` — both fine here, but the comparison is purely lexical, so a future maintainer reading the code might assume type safety that's not actually present.

Cheapest fix: special-case nil (`if v == nil { continue }`) and either reject non-string comparison values up front, or document explicitly that comparison is by `fmt.Sprintf("%v", ...)` formatting. The PR description already says "no numeric values" in scope, which reads like a parsing restriction (unquoted numerics rejected), but the runtime comparison doesn't actually enforce string-only — it just compares formatted output. The two statements are inconsistent.

**`strings.Replace` substring match is fragile if `currentKey` is reused**

```go
newPath := strings.Replace(path, fmt.Sprintf("%s.%s.", currentKey, keys[1]), "", 1)
```

This works in the tests, but consider a path like `components.x.#(name=="B").status` where `currentKey = "components"`. The pattern `components.#(name=="B").` doesn't appear in the path, so Replace is a no-op and the subsequent `newPath == path` check fires, returning the matched element. Correct.

A more dangerous case: `components.components.#(name=="B").status`. Here `currentKey = "components"` and `keys[1] = "components"` after the first lookup. `keys[1]` is not a query, so the query branch is skipped entirely. Fine.

What about `a.#(name=="x").a.#(name=="y").b`? After first lookup, `currentKey = "a"`, `keys[1] = "#(name==\"x\")"`. Replace strips `a.#(name=="x").`, leaving `a.#(name=="y").b`. `walk` recurses with the matched element. Correct.

So the substring-match approach happens to be safe given the tokenization rules. The PR would be slightly more robust if it used the byte-index of the matched segment from the parsed token positions rather than substring search, but given the test coverage, this is fine. A one-line comment explaining "this is safe because `keys[1]` is a single path segment, so the only `currentKey.keys[1].` substring in the original path is the one we want to strip" would head off future confusion.

**Test name mismatch**

```go
Name: "query-without-preceding-array-field-falls-back-to-whole-array",
Path: "components",
Data: `{"components": [{"name": "A"}]}`,
```

The path here doesn't include a query at all, so the test name is misleading. It's actually a regression test that the no-query path still returns the whole array as a string — useful, just rename it (something like `array-without-query-still-returns-whole-array` or `regression-no-query-still-stringifies`). The "without preceding array field" framing reads like a test for root-level `#(...)` queries, which the PR explicitly says aren't supported.

**Missing test: query + remaining path with brackets**

The new tests cover `query.field` but not `query[N]` or `query.field[N]`. The bracket handling in `walk()` is pre-existing, but since this PR changes the depth-tracking to cover `(`, it's worth at least one assertion that `query + bracket index` still works — e.g. `components.#(name=="B").tags[0]`.

**Doc duplication**

The `Eval` doc comment block now describes the query syntax, and so does `queryArray`'s own doc comment. The `queryArray` version is more complete (mentions the explicit limitations). I'd suggest keeping the detail in `queryArray` only and trimming the `Eval` comment to a one-liner that points at it — otherwise a future doc update has to update both places.

**Minor: nil-pointer on non-objects inside the array**

```go
if object, ok := element.(map[string]interface{}); ok {
```

This silently skips array elements that aren't objects. Probably the right behavior, but if an endpoint returns `[{"name": "A"}, "not-an-object"]`, the string element is skipped without comment. Combined with the `no element found matching query` error at the end, a user debugging a malformed payload will get a misleading "no match" instead of "skipped a non-object entry". Consider adding a debug-level log or, if you prefer, a more specific error like `no matching object element (found N non-object entries)`. Low priority.

**Overall**

The scope discipline is good. The motivating example is real, the syntax matches an existing precedent (gjson), and the dot-paren depth tracking is the minimum change needed to make the new syntax co-exist with the existing tokenization. The `fmt.Sprintf("%v", ...)` comparison is the one thing I'd want to either tighten or document before this lands — the rest is mostly polish.

jensenbox:
Thanks for the thorough review! Addressed the two things that mattered:

- `queryArray` now treats a missing key or a literal `nil` value as no match, rather than formatting nil as "<nil>" and letting it match a query for that literal string. Added a regression test for it.
- Renamed the mislabeled test to `array-without-query-still-returns-whole-array`, which is what it actually verifies.

Left the numeric-value doc/enforcement mismatch, the `queryArray`/`Eval` doc duplication, and the `query + bracket index` test as-is for now — none are correctness bugs, and I'd rather keep this PR scoped to the array-selection feature. Happy to open a fast follow-up if you'd rather see those addressed before merge.

PR #1773, opened 2026-08-21
Honor minimum-reminder-interval for external endpoints

### LINKED ISSUE #1765: minimum-reminder-interval is ignored for external endpoints with heartbeat
### Describe the bug

`minimum-reminder-interval` is ignored for external endpoints using `heartbeat`. The alert re-sends on every heartbeat tick. Regular endpoints are unaffected.

### What do you see?

With `minimum-reminder-interval: 1h` and `heartbeat.interval: 5m`, alerts at 03:25, 03:30, 03:40, 03:50, 04:00.

One alert, then a reminder no sooner than 1h later.

1. Add an external endpoint with `heartbeat.interval: 5m`, and an alert with `failure-threshold: 3` and `minimum-reminder-interval: 1h`
2. Stop pushing results to it, so the heartbeat starts failing
3. The alert triggers after 15m, then repeats every 5m instead of every 1h

### Version

v5.36.0

### Additional information

_No response_

yzxcj797:
Fixes #1765.

## Root cause

External endpoints run alerting against a `ToEndpoint()` **copy** on every heartbeat tick (and every pushed result). Two pieces of alert state have different fates across that conversion:

- `Alert.Triggered` survives, because `ToEndpoint` shares the `Alerts` slice and its pointer elements;
- `Endpoint.LastReminderSent` does not: it is a plain field on the converted copy. `handleAlertsToTrigger` stamps it there (`ep.LastReminderSent = time.Now()`), the copy is discarded, and only the failure/success counters are synced back to the `ExternalEndpoint`.

So every tick built a fresh copy with the zero `time.Time`, `time.Since(zero)` was always past `minimum-reminder-interval`, `sendReminder` was always true, and the alert re-sent on each heartbeat evaluation — with `heartbeat.interval: 5m` and `minimum-reminder-interval: 1h`, alerts at 03:25, 03:30, 03:40… exactly as reported.

## Fix

- `ExternalEndpoint` gains `LastReminderSent` (in-memory only, `yaml:"-"`);
- `ToEndpoint` carries it into the converted endpoint;
- both call sites — `executeExternalEndpointHeartbeat` (watchdog) and `CreateExternalEndpointResult` (push API) — sync it back alongside the counters they already sync.

## Test

`TestHandleAlertingOnExternalEndpointConversionPreservesReminderInterval` (watchdog) simulates the per-tick conversion flow: the initial trigger stamps the clock; a tick 30m later (below the 1h interval) must not refresh it; a tick 2h later must. Differential: the test does not build against `main` (the field/copy-back do not exist there). `go test ./watchdog/ ./config/endpoint/ ./api/` all green.

### DIFF
--- a/api/external_endpoint.go
+++ b/api/external_endpoint.go
@@ -78,8 +78,13 @@ func CreateExternalEndpointResult(cfg *config.Config) fiber.Handler {
 		// Check if an alert should be triggered or resolved
 		if !cfg.Maintenance.IsUnderMaintenance() && !inEndpointMaintenanceWindow {
 			watchdog.HandleAlerting(convertedEndpoint, result, cfg.Alerting)
+			// Sync the counters and the reminder clock back; LastReminderSent
+			// lives on the converted copy and would otherwise reset to zero on
+			// every pushed result, re-sending the alert and ignoring
+			// minimum-reminder-interval (#1765).
 			externalEndpoint.NumberOfSuccessesInARow = convertedEndpoint.NumberOfSuccessesInARow
 			externalEndpoint.NumberOfFailuresInARow = convertedEndpoint.NumberOfFailuresInARow
+			externalEndpoint.LastReminderSent = convertedEndpoint.LastReminderSent
 		} else {
 			logr.Debug("[api.CreateExternalEndpointResult] Not handling alerting because currently in the maintenance window")
 		}
--- a/config/endpoint/external_endpoint.go
+++ b/config/endpoint/external_endpoint.go
@@ -48,6 +48,13 @@ type ExternalEndpoint struct {
 
 	// NumberOfSuccessesInARow is the number of successful evaluations in a row
 	NumberOfSuccessesInARow int `yaml:"-"`
+
+	// LastReminderSent is the time at which the last reminder was sent for this endpoint.
+	// External endpoints run alerting against a converted copy (see ToEndpoint),
+	// so the copy's value must be synced back here after every HandleAlerting
+	// call, or every heartbeat tick / pushed result would see a zero value and
+	// re-send the alert, ignoring minimum-reminder-interval (#1765).
+	LastReminderSent time.Time `yaml:"-"`
 }
 
 // ValidateAndSetDefaults validates the ExternalEndpoint and sets the default values
@@ -95,6 +102,7 @@ func (externalEndpoint *ExternalEndpoint) ToEndpoint() *Endpoint {
 		Alerts:                  externalEndpoint.Alerts,
 		NumberOfFailuresInARow:  externalEndpoint.NumberOfFailuresInARow,
 		NumberOfSuccessesInARow: externalEndpoint.NumberOfSuccessesInARow,
+		LastReminderSent:        externalEndpoint.LastReminderSent,
 	}
 	return endpoint
 }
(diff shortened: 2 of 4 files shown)


wahajahmed010:
Clean fix — you nailed the asymmetry (`Alerts` slice shared, value-type fields like `LastReminderSent` lost on copy) and the consequence (`time.Since(zero) > interval` always true). Couple of things worth a closer look:

**Logic — is the new field's zero value safe across the codebase?**
- `LastReminderSent time.Time` defaults to Go's zero time. The alerting path presumably treats zero as "no reminder has ever been sent, so this is the first one — stamp now". Worth grep-confirming every reader of `LastReminderSent` does that check (i.e., `if !lastReminderSent.IsZero()` or `time.Since(t) > interval`). If any reader does a plain `t.Add(interval)` or similar, the math will explode on the first call. The PR's test starts from `LastReminderSent = time.Now().Add(-30*time.Minute)` and only exercises the reminder path, not the initial trigger — it would catch a math bug, but not an "is zero treated as 'never sent'" assumption elsewhere.

**Tests**
- Good coverage of the three regimes (initial trigger, within interval, past interval). Two small gaps:
  - The test uses `MOCK_ALERT_PROVIDER` to count provider invocations but never asserts on the count — you could strengthen the "no reminder sent within interval" assertion by also confirming `MOCK_ALERT_PROVIDER` was called exactly once across the whole test (initial trigger only), not twice.
  - `time.Since(ee.LastReminderSent) < 25*time.Minute` is a fuzzy tolerance for the "no reminder sent" case. A precise check (`ee.LastReminderSent.Equal(prevStamp)`) would be less flaky on a slow CI runner where the assertion runs after more than 5 minutes have passed since `time.Now().Add(-30*time.Minute)`. That said, the tolerance is generous enough that it should pass in practice.

**Concurrency**
- Two call sites (`executeExternalEndpointHeartbeat` and `CreateExternalEndpointResult`) both read-modify-write `LastReminderSent` on the shared `ExternalEndpoint`. If the heartbeat watcher and the push API can fire concurrently for the same endpoint (and the watchdog model suggests they can — `monitorExternalEndpoint` polls, `CreateExternalEndpointResult` accepts pushed results), there's a lost-update window: watcher reads `t1`, push reads `t1`, watcher stamps `t2`, push stamps `t3`, watcher's `t2` is silently overwritten. The fix doesn't introduce this — it pre-exists for `NumberOfSuccessesInARow` — but you're widening the field set that races. Worth at minimum a comment, ideally a sync.Mutex or atomic.Value around the conversion+sync block. Not a blocker, but the issue ticket would close better with a documented concurrency story.

**Error handling**
- N/A — purely a state-preservation fix.

**Security**
- N/A.

**Performance**
- N/A.

**Breaking changes**
- One small thing: `LastReminderSent` is `yaml:"-"`, so it won't accidentally serialize. Good. But the YAML comment says "External endpoints run alerting against a converted copy (see ToEndpoint), so the copy's value must be synced back here" — this is the kind of invariant that should live on the struct, not in a comment. If a future contributor adds a third call site that calls `ToEndpoint()` and forgets to sync back, the bug silently returns. Consider: (a) removing `ToEndpoint()` from the external endpoint and calling `HandleAlerting` directly on a method that handles the syncing internally, or (b) a unit test that fails if `LastReminderSent` ever appears as the zero time after a `HandleAlerting` call. The latter is the lighter touch.

Overall solid fix with good test coverage. The concurrency observation is the main thing worth deciding before merge.

PR #1776, opened 2026-08-21
Resolve endpoint hostnames as FQDNs through the custom DNS resolver

### LINKED ISSUE #1769: client.dns-resolver still applies the system search list, sending 8 DNS queries per check on Kubernetes
### Describe the bug

When `client.dns-resolver` is set, Gatus builds a `net.Resolver` with `PreferGo: true` and a custom `Dial`. The pure-Go resolver still reads `/etc/resolv.conf` for `search` and `ndots`, so the *search list* is applied to the custom resolver as well — Gatus just sends the resulting queries somewhere else.

On Kubernetes this is costly. kubelet writes `options ndots:5` into every `ClusterFirst` pod, and a normal monitored hostname has fewer than 5 dots, so it is treated as relative and the search list is walked first:

```
audiobookshelf.example.com.observability.svc.cluster.local   NXDOMAIN
audiobookshelf.example.com.svc.cluster.local                 NXDOMAIN
audiobookshelf.example.com.cluster.local                     NXDOMAIN
audiobookshelf.example.com                                   ← the intended one
```

Doubled by A and AAAA, that is **8 DNS queries per check, 6 of them guaranteed NXDOMAIN**, all sent to the configured public resolver. Each one also discloses the cluster's internal namespace layout (`<namespace>.svc.cluster.local`) to that resolver.

The hostname in an endpoint URL is always meant as an absolute name, so none of the search-list expansion is wanted here.

### Affected code

Both places that build the custom resolver:

- `client/config.go` `getHTTPClient`, ~L259-270 — the HTTP path
- `client/client.go` `CanPerformStartTLS`, ~L168-177 — same pattern

```go
dialer := &net.Dialer{
    Resolver: &net.Resolver{
        PreferGo: true,
        Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
            d := net.Dialer{}
            return d.DialContext(ctx, dnsResolver.Protocol, dnsResolver.Host+":"+dnsResolver.Port)
        },
    },
}
```

### Evidence

My router intercepts outbound port 53, which is the only reason this was visible at all. Router hit counts over ~2 days — 20 of the top 25 domains are the bogus variants, in lockstep at ~11k each:

```
 6. audiobookshelf.example.com.svc.cluster.local.                10994
 7. audiobookshelf.example.com.cluster.local.                    10993
 8. audiobookshelf.example.com.observability.svc.cluster.local.  10992
 9. echo.example.com.observability.svc.cluster.local.            10984
10. kavita.example.com.observability.svc.cluster.local.          10984
```

282 of any 500 consecutive queries reaching the resolver were this. Packet capture on the node, one TCP connection per query:

```
eth:ip:tcp:dns  10.0.3.28:58108 -> 1.1.1.1:53  audiobookshelf.example.com.observability.svc.cluster.local
eth:ip:tcp:dns  10.0.3.28:58126 -> 1.1.1.1:53  audiobookshelf.example.com.svc.cluster.local
```

Without a DNS intercept in place this is invisible — the queries just go to Cloudflare — which is presumably why it has gone unreported.

### Suggested fix

Make the name absolute at dial time when a custom resolver is configured, so the resolver skips the search list:

```go
host, port, err := net.SplitHostPort(addr)
if err == nil && !strings.HasSuffix(host, ".") && net.ParseIP(host) == nil {
    addr = net.JoinHostPort(host+".", port)
}
```

This only affects name resolution. SNI and the `Host` header come from the request URL, not the dial address, so a trailing dot added here is not visible on the wire to the target.

Alternatively, a `client.dns-search` / `client.absolute-names` toggle would work, but the search list seems never to be desirable for an explicitly configured monitoring target.

### Workaround

Pod-level `dnsConfig`:

```yaml
dnsConfig:
  options:
    - name: ndots
      value: "1"
```

### Downstream

Reported first at home-operations/gatus-sidecar#168, since the `dns-resolver` setting is widely copied from that project's example Gateway annotation. The chart added `dnsConfig` support in home-operations/gatus-sidecar#169 as a workaround, but the underlying behaviour is here.

### Version

Gatus v5.36.0.

yzxcj797:
Fixes #1769.

## Root cause

Exactly as the issue diagnosed: a `PreferGo` resolver reads `/etc/resolv.conf`'s `search`/`ndots` even when its `Dial` points at a custom DNS server — the search-list expansion happens **before** the query ever reaches the custom transport. On Kubernetes (`options ndots:5` in every ClusterFirst pod), an ordinary monitored hostname is treated as relative and the whole search list is walked first; doubled by A+AAAA that is 8 queries per check, 6 guaranteed NXDOMAINs, all disclosed to the configured public resolver — the issue's router hit counts (`~11k` lockstep hits per bogus `*.svc.cluster.local` variant).

## Fix

The hostname in an endpoint URL is always meant as an **absolute** name. Both custom-resolver sites (`client/config.go` `getHTTPClient` and `client/client.go` `CanPerformStartTLS`) now resolve through a shared `absoluteLookupDialContext`:

- `host` is looked up **with a trailing dot** — the FQDN marker that disables search-list expansion in the pure-Go resolver — so exactly one query per family goes to the configured server;
- the dial then connects to the **resolved IP**, bypassing the resolver entirely;
- literal IP targets dial directly without a lookup;
- multiple A records are tried in order until one connects.

`newAbsoluteLookupDialer` keeps the same custom-resolver `Dial` as before, so the configured protocol/host/port semantics are unchanged.

## Test

`TestAbsoluteLookupDialContextSendsFQDNQueries` runs a minimal in-process UDP DNS responder on an ephemeral port and asserts through a real TCP dial that (a) a query arrived, (b) its QNAME is absolute (`example.test.` — trailing dot present), and (c) **no** search-list variant (`example.test.<suffix>`) was ever queried — the exact ndots-walk signature from the issue. The responder answers `127.0.0.1` and the test completes a real connection through the resolved IP.

Full `client` suite green (19 tests including the new one); `go vet` clean. The StartTLS site is a mechanical call-site swap — same helper, same semantics.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -165,16 +165,10 @@ func CanPerformStartTLS(address string, config *Config) (connected bool, certifi
 			// It shouldn't happen, but if it does, we'll log it... Better safe than sorry ;)
 			logr.Errorf("[client.getHTTPClient] THIS SHOULD NOT HAPPEN. Silently ignoring invalid DNS resolver due to error: %s", err.Error())
 		} else {
-			dialer := &net.Dialer{
-				Resolver: &net.Resolver{
-					PreferGo: true,
-					Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
-						d := net.Dialer{}
-						return d.DialContext(ctx, dnsResolver.Protocol, dnsResolver.Host+":"+dnsResolver.Port)
-					},
-				},
-			}
-			connection, err = dialer.DialContext(context.Background(), "tcp", address)
+			dialer := newAbsoluteLookupDialer(dnsResolver)
+			// FQDN lookup through the custom resolver: one query per family,
+			// no system search-list expansion (#1769).
+			connection, err = absoluteLookupDialContext(context.Background(), dialer, dnsResolver, "tcp", address)
 			if err != nil {
 				return
 			}
--- a/client/config.go
+++ b/client/config.go
@@ -256,17 +256,11 @@ func (c *Config) getHTTPClient() *http.Client {
 				// It shouldn't happen, but if it does, we'll log it... Better safe than sorry ;)
 				logr.Errorf("[client.getHTTPClient] THIS SHOULD NOT HAPPEN. Silently ignoring invalid DNS resolver due to error: %s", err.Error())
 			} else {
-				dialer := &net.Dialer{
-					Resolver: &net.Resolver{
-						PreferGo: true,
-						Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
-							d := net.Dialer{}
-							return d.DialContext(ctx, dnsResolver.Protocol, dnsResolver.Host+":"+dnsResolver.Port)
-						},
-					},
-				}
+				dialer := newAbsoluteLookupDialer(dnsResolver)
 				c.httpClient.Transport.(*http.Transport).DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
-					return dialer.DialContext(ctx, network, addr)
+					// FQDN lookup through the custom resolver: one query per family,
+					// no system search-list expansion (#1769).
+					return absoluteLookupDialContext(ctx, dialer, dnsResolver, network, addr)
 				}
 			}
 		}
--- a/client/dns_absolute_dial.go
+++ b/client/dns_absolute_dial.go
@@ -1,0 +1,76 @@
+package client
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"strings"
+	"time"
+)
+
+// newAbsoluteLookupDialer returns a dialer whose hostname resolution goes
+// through the configured custom DNS resolver with FQDN semantics: the
+// hostname is looked up with a trailing dot, which the pure-Go resolver
+// treats as fully qualified and therefore does NOT expand through the
+// system search list from /etc/resolv.conf (#1769).
+//
+// Why the trailing dot matters: a `PreferGo` resolver still reads
+// /etc/resolv.conf for `search`/`ndots`, and on Kubernetes every pod gets
+// `options ndots:5` — a normal monitored hostname has fewer than 5 dots, is
+// treated as relative, and the whole search list is walked first
+// (namespace.svc.cluster.local, svc.cluster.local, cluster.local, ...) with
+// the intended absolute query only last. Doubled by A and AAAA lookups, that
+// is 8 queries per check — 6 guaranteed NXDOMAINs — all sent to the
+// configured public resolver, disclosing the cluster's namespace layout.
+//
+// The endpoint hostname in a URL is always meant as an absolute name, so
+// none of that expansion is wanted. Once the IPs are resolved absolutely,
+// the dial itself connects to the IP, bypassing the resolver entirely.
+func newAbsoluteLookupDialer(dnsResolver *DNSResolverConfig) *net.Dialer {
+	resolver := &net.Resolver{
+		PreferGo: true,
+		Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+			d := net.Dialer{}
+			return d.DialContext(ctx, dnsResolver.Protocol, dnsResolver.Host+":"+dnsResolver.Port)
+		},
+	}
+	return &net.Dialer{
+		Timeout:  30 * time.Second,
+		Resolver: resolver,
+	}
+}
+
+// absoluteLookupDialContext resolves host with FQDN semantics through the
+// custom resolver and dials the first IP. addr is "host:port".
+func absoluteLookupDialContext(ctx context.Context, dialer *net.Dialer, dnsResolver *DNSResolverConfig, network, addr string) (net.Conn, error) {
+	host, port, err := net.SplitHostPort(addr)
+	if err != nil {
+		return nil, fmt.Errorf("client.absoluteLookupDialContext: %w", err)
+	}
+	if ip := net.ParseIP(host); ip != nil {
+		// Literal addresses never hit the search list; dial directly.
+		return dialer.DialContext(ctx, network, net.JoinHostPort(host, port))
+	}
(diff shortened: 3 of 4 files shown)


wahajahmed010:
Nice root-cause write-up — the security angle (disclosing cluster namespace layout via the public resolver on every check) is what elevates this from a perf nit to something that genuinely needed fixing. The trailing-dot FQDN approach is the right surgical choice.

A few things worth tightening before merge:

**Logic — correctness of the "try all A records in order" loop**
- `absoluteLookupDialContext` falls through `ips` in order, returning the first successful dial. For a host with one A and one AAAA record returned by `LookupHost`, the IP ordering is resolver-defined (Go's pure-Go resolver has no documented guarantee on family order). In practice, with a `PreferGo` resolver and no Happy Eyeballs at this layer, you may end up preferring AAAA and waiting through an IPv6 timeout for every dual-stack endpoint. Consider: (a) calling `LookupIPAddr` with a `network` parameter and dialing family-matching addrs in parallel, or (b) documenting the current ordering as "first successful dial wins" and noting that Happy Eyeballs is not implemented here. The current behavior is correct, just slow on v6-broken networks.

**Tests**
- `TestAbsoluteLookupDialContextSendsFQDNQueries` is solid but couples the test to a hand-rolled DNS packet encoder/decoder (`decodeQName`, `buildAAnswer`, `itoa`). That's fine, but it duplicates a fair amount of logic that `net.PacketConn` mocks or `github.com/miekg/dns` would handle for you. Worth knowing if you want this suite to grow — a `miekg/dns`-based test server reads almost identically and is more robust against header-flag mistakes in `buildAAnswer` (e.g., `response[3] &= 0x7F` clears AA, which is correct, but the rest of the response flags aren't explicitly set; minor).
- More importantly: the assertion only checks the first question reaches the resolver. The comment claims "Every observed question must be absolute — the ndots walk would show up as `example.test.<suffix>.` queries here" but the loop after the select drains a buffered channel and exits on the default case. If the resolver were slow and only delivered some queries before the test returned from the dial, the assertion wouldn't fail. Consider using `runtime.Gosched()` + a small grace period, or wrapping the whole dial call in a goroutine and asserting on a fixed timeout.

**Error handling**
- `absoluteLookupDialContext` swallows the connection error from each IP and only returns `lastErr` after the loop. That's fine, but the returned error doesn't tell you how many IPs were tried. For ops debugging on a flaky dual-stack target, "tried 1 ip, all failed: connection refused" would be much more useful than just the last error.

**Security**
- The PR description mentions the cluster-disclosure angle, but the fix only applies when `dns-resolver` is configured. The default code path (system resolver) still goes through Go's default `PreferGo: true` resolver and *also* reads `/etc/resolv.conf`'s search list. That's the same disclosure issue when the host's resolv.conf has `ndots:5`. Worth either: (a) calling out in the docs that running without an explicit `dns-resolver` on Kubernetes inherits the same risk, or (b) applying the FQDN treatment to the default-resolver path too.

**Performance**
- Net win: replacing 8 queries with 1-2 per check is exactly the kind of fix that pays for itself in DNS load on busy clusters. The IP-dial fallback loop is the only potential regression and it's small.

**Breaking changes**
- None visible. New file, no public API changes, no config surface changes.

Overall this is a well-scoped fix with a clear before/after. The two suggestions above (dual-stack dial behavior, default-resolver coverage) are the only things I'd want answered before approving.

wahajahmed010:
## Review

This is a substantive fix for a real DNS-leakage issue (#1769) and the underlying `ndots:5` problem on Kubernetes. The motivation is correct: when `PreferGo` is set, the resolver still walks `/etc/resolv.conf`'s search list, so a single monitored hostname triggers up to 8 queries (4 search-list variants × A/AAAA) per check, all going to the configured public resolver and disclosing cluster-namespace layout. The fix — append a trailing dot to force FQDN semantics and skip expansion — is the right one and the comment block on `newAbsoluteLookupDialer` documents the tradeoff clearly.

A few observations on the implementation:

1. **Test only covers the single-IP path.** `absoluteLookupDialContext` walks `ips` and dials them in order, but `TestAbsoluteLookupDialContextSendsFQDNQueries` only asserts one query and one resolved address. Worth adding a case where the resolver returns multiple IPs and the first is unreachable — that's the path most likely to regress (e.g. someone "simplifies" the loop into a single `dialer.DialContext` and silently drops the failover behavior).

2. **`itoa` is hand-rolled.** It's a 9-line helper for `strconv.Itoa` in a file that already imports `time` and `strings`. Just import `strconv` and use it — the hand-rolled version is also subtly wrong for negative numbers (would loop forever on `v > 0`? no, it terminates, but still wrong), and `strconv` will panic on negatives consistently. Not a blocker, but it's noise.

3. **DecodeQName in the test is fragile.** The loop `for i < len(packet) && query[i] != 0 { i += int(query[i]) + 1 }` will overrun if a malformed packet has a label-length byte pointing past `len(packet)`. Fine for the synthetic packet builder here, but a defensive `if i+int(packet[i]) >= len(packet) { break }` would prevent the same test from flaking under future changes to `buildAAnswer`.

4. **Error wrapping.** `absoluteLookupDialContext` returns `fmt.Errorf("...: %w", err)` which is good, but on the multi-IP loop the final returned `lastErr` is the bare error from `dialer.DialContext`, missing the `client.absoluteLookupDialContext:` prefix the first failure gets via the lookup path. Minor consistency thing — wrap on the way out too.

5. **One subtle behavior change worth calling out.** Before this PR, the original `&net.Dialer{Resolver: ...}` in `getHTTPClient` had no explicit `Timeout`; now `newAbsoluteLookupDialer` sets `Timeout: 30 * time.Second`. That's a sensible default but it does change behavior for users who relied on the inherited `http.Transport` defaults — worth a one-liner in the PR description so it's not a surprise on release.

No blockers. The core fix is correct, well-commented, and addresses a real cluster-side concern.

Diaoul:
Hey :wave: 

Is there anything I can do to help get this merged?

PR #1577, opened 2026-03-10
feat(config): support environment variable default values

ndpvt-web:
## Summary
Adds support for default value fallback when using environment variables in gatus configuration using the standard shell syntax `${VAR:-default}`.

Closes #1523

## Changes
- Added custom `expandEnv()` function in `config/config.go` that supports `${VAR:-default}` syntax
- If the environment variable is set and non-empty, its value is used
- If the environment variable is unset or empty, the default value is used  
- Fully backward compatible with existing `$VAR` and `${VAR}` syntax
- Added comprehensive unit tests (15 test cases) in `config/env_test.go`

## Example
```yaml
endpoints:
  - name: My Endpoint
    url: "${API_URL:-https://api.example.com}"
    conditions:
      - "[STATUS] == ${EXPECTED_STATUS:-200}"
```

## Test Plan
- [x] Unit tests for default value substitution
- [x] Tests for backward compatibility with existing syntax  
- [x] Tests for edge cases (special chars, empty defaults, nested vars)
- [ ] Manual testing with real config file

## AI Disclosure
This PR was authored with assistance from an AI coding assistant (Claude). The implementation was developed transparently as part of an open-source contribution workflow. All code has been reviewed for correctness and follows the project's existing patterns.

PR #1777, opened 2026-08-23
feat(config): support default values for environment variables

### LINKED ISSUE #1523: Default value for environment variable in config

Not sure if I missed it, but it seems not possible to configure a default value fallback when using environment variables in config i.e ${SOME_VAR:-false} 

Easier deployment of config when env isnt set 

5month :D

### Additional information

_No response_

fly1d:
## Summary

Allow configuration values to specify a default fallback when an environment variable is unset or empty, e.g. `${PORT:-8081}`.

This is implemented as a small preprocessing step before `os.ExpandEnv`, so the existing `$VAR`, `${VAR}`, and `$$` literal-dollar behavior is unchanged.

Closes #1523.

## Verification

- `go test ./config/ -run 'TestParseAndValidateConfigBytes' -count=1`: passed
- `go vet ./config/`: passed
- `go test ./config/ -run 'TestParseAndValidateConfigBytesWithEnvironmentVariableDefaults' -count=1`: passed

Note: the unrelated `config/suite` integration test `TestSuite_ExecuteWithAlwaysRunEndpoints` is flaky under parallel package runs and fails on the base branch as well; it passes in isolation.

Prepared by an agent: OpenAI Codex (GPT-5).


fly1d:
Hi maintainers, I wanted to check the preferred direction for this change before making any follow-up commit. The implementation supports `${VAR:-default}` when the variable is unset or empty, preserves existing `$` and `$$` behavior, and includes README coverage plus config tests.

Would this syntax and expansion order fit the project configuration conventions, or would you prefer different edge-case handling or documentation wording? I am happy to adjust the patch narrowly based on your guidance. This PR was prepared with OpenAI Codex assistance.

fly1d:
I missed #1577 when I opened this PR. It addresses the same issue and predates this work. #1777 is a smaller implementation and includes the requested README documentation, while #1577 has broader parser tests.

I do not want to compete with or duplicate the earlier contribution. @TwiN, would you prefer that I close #1777, keep it as an alternative, or coordinate with @ndpvt-web on a combined version?

OpenAI Codex (GPT-5) assisted with this review.

wahajahmed010 (commented):
Nice feature, well-scoped, and the test coverage is solid. There are a few correctness issues around how the default-value substitution interacts with the existing literal-dollar handling, plus a small spec ambiguity worth deciding explicitly.

**1. Default value is not protected against re-expansion or literal-dollar substitution**

The order in `parseAndValidateConfigBytes` is:

1. `$$` -> `__GATUS_LITERAL_DOLLAR_SIGN__`
2. expand `${VAR:-default}` -> `default` literal
3. `os.ExpandEnv` runs on the result
4. `__GATUS_LITERAL_DOLLAR_SIGN__` -> `$`

A user writes:

```yaml
web:
  port: ${PORT:-8081}
  title: ${TITLE:-$Default Title}
```

After step 2, the YAML string still contains `$Default Title` (the default is substituted in literally). Step 3 then sees `$Default` and `os.ExpandEnv` will look for env var `Default` — at best it expands to the empty string, at worst it gets it from a real env var and silently changes the title. The literal-dollar sentinel does not protect defaults because it was emitted before the substitution.

Two fixes, pick one:

- Capture the default in `expandEnvironmentVariablesWithDefaults` and re-apply the same `__GATUS_LITERAL_DOLLAR_SIGN__` rule to the default value before inserting it.
- Or: do all `${VAR:-default}` substitution after `os.ExpandEnv`, by overriding the `os.ExpandEnv` mapping function. The mapping function receives `${VAR:-default}` and can resolve the default itself, avoiding any re-expansion. This is cleaner — see https://pkg.go.dev/os#ExpandEnv.

The second is what I'd recommend. With it, you don't need the new regex pass at all.

**2. Empty-default edge case differs from bash**

```go
if resolved, ok := os.LookupEnv(name); ok && resolved != "" {
    return resolved
}
return defaultValue
```

This means `${PORT:-}` (empty default) will substitute the empty string for both unset and empty PORT, which matches bash's `:-` operator. Good.

But `${PORT-8081}` (single dash, bash's "use default only if unset") is not supported. The test name says "defaults for environment variables" and the docs example uses `:-`, so this is fine for the stated scope, but worth either documenting that only `:-` is supported, or implementing both forms. The current regex `\$\{([A-Za-z_][A-Za-z0-9_]*):-([^}]*)\}` rejects `${PORT-8081}` silently, which would be a confusing user experience if anyone tries.

**3. Default value is not escaped against YAML semantics**

If a user writes:

```yaml
conditions: "[BODY] == ${BODY_PATTERN:-200}"
```

And the default is `200`, fine. But if a user wants a default containing a colon (e.g. a URL), the regex `[^}]*` is fine for that — colons in the default work. However, the *test* asserts `config.Endpoints[0].Conditions[0] != "[BODY] == fallback"`, so the substitution is occurring post-YAML-parse, which means any default containing a YAML-special character (quote, backslash) is going to be interpreted by the YAML parser before your code sees it. This is probably the intended behavior, but worth a one-line comment in `expandEnvironmentVariablesWithDefaults` saying "operates on already-parsed YAML strings; YAML escaping rules have already been applied."

**4. Regex does not anchor on word boundary**

`\$\{([A-Za-z_][A-Za-z0-9_]*):-([^}]*)\}` is anchored on `$` but not on the closing brace context. A user writing `${PORT:-8081}` followed by literal `:-` in the same string would still match. Probably never an issue in practice for YAML config, but if anyone uses this for a query body or alert condition string, the regex will over-match. Using `\$\{([A-Za-z_][A-Za-z0-9_]*):-((?:[^{}]|\{[^{}]*\})*)\}` would constrain the default to non-brace content, preventing accidental overlap. Optional.

**5. Minor: identifier character class**

`[A-Za-z_][A-Za-z0-9_]*` is the bash spec. Go's `os.ExpandEnv` accepts the same set. Good consistency. (Bash also accepts leading digits in `${1:-default}` for positional parameters, but that's not relevant here.)

**6. No `os.Setenv` race test**

`t.Setenv` is fine, but two parallel `t.Run` subtests calling `t.Setenv` with the same name will be serialized. If anyone later refactors the table-driven test into parallel subtests, they will hit a `t.Parallel` + `t.Setenv` panic. Not a blocker; just a heads-up.

**Overall**

The feature is useful and the test is meaningful. The main thing I'd want before merge is #1 — fix the re-expansion of `$` inside defaults. Once that's addressed, the rest is nice-to-have. Thanks for the clean PR.

PR #1780, opened 2026-08-25
feat(alerting): add Jira alerting provider

### LINKED ISSUE #489: Introduce Jira support

Add Jira as a new alert provider. Implementation should very similar with GitHub provider.

_No response_

_No response_

### Additional information

_No response_

ChrisJr404:
Closes #489.

Adds Jira as a new alerting provider, modeled on the existing GitHub provider like the issue suggested. On trigger it creates an issue in the configured project, and when `send-on-resolved` is set it looks up the matching open issue(s) and moves them through the configured transition (`Done` by default) once the alert resolves. Authentication is basic auth with an account email and an API token.

Config is minimal to get started:

```yaml
alerting:
  jira:
    base-url: "https://your-domain.atlassian.net"
    username: "you@example.com"
    token: "your-api-token"
    project-key: "OPS"
```

`issue-type` (default `Task`) and `resolve-transition` (default `Done`) are optional. Added unit tests covering create/resolve/error paths and README docs.


TwiN (approved):
LGTM, great work!

TwiN:
@ChrisJr404 would you mind looking into copilot's suggestions? I don't have a jira instance I can validate those on, nor am I super familiarwith Jira's API

ChrisJr404:
Done. Pushed a commit that addresses both of Copilot's points:

- The create/search/transition error paths now include the Jira response body, so JQL parse errors, auth failures, and invalid-transition responses actually show up in the error message.
- The resolve path now pages through the search results (startAt/maxResults) instead of relying on a single page, so a matching issue won't be missed on projects with a lot of open issues.

No live Jira instance needed on your end; the existing unit tests cover the create/resolve/error paths and still pass.

TwiN (approved):

ChrisJr404:
Both of Copilot's points are reasonable:

1. Pagination on resolve: rather than fetching all non-Done issues and filtering the summary in-memory (which misses matches past the first page), I can narrow the JQL to the exact summary and request only the fields I need, so resolution does not depend on page size.
2. Error detail: I can include Jira's response body in the returned errors, since that is where the actionable detail (permissions, invalid issue type) usually is.

Neither needs a live Jira to reason about. Want me to push those two changes on top, or would you rather merge as-is and iterate?

wahajahmed010 (commented):
Substantial new feature, well-scoped to match the GitHub provider. A few things worth thinking through:

**Security / Auth**
- Basic auth header is correctly built with `base64.StdEncoding` — note this is *encoding*, not encryption; if `cfg.BaseURL` is ever `http://` (rather than `https://`) the token will be sent in the clear. The README only documents `https://*.atlassian.net`, but `BaseURL` is a free-form string. Either reject non-https schemes in `Validate()`, or document the assumption loudly. This is a footgun.
- The token lives in YAML config. Standard for Gatus, but worth a note that the YAML should be 0600 / not committed. Probably out of scope for this PR, but a paragraph in the README security section would help.

**Resolve logic correctness**
- `resolveIssues` searches with `statusCategory != Done`. That misses resolved-state categories Jira uses (e.g. `Closed` in some workflows, or `statusCategory = Resolved`). If a project's "Done" transition moves the issue to "Closed" instead of the canonical "Done" status category, the search will keep returning the same already-transitioned issues on each page until `startAt >= Total`. Not a leak, but a lot of wasted requests. Worth either filtering on `statusCategory != Done AND statusCategory != Closed` or — better — capturing the issue key at create time (e.g. store it in the description) and transitioning by key directly.
- The JQL `project = %q` uses Go's `%q` which produces Go-style double-quoted strings. That happens to be valid for simple keys like `OPS`, but if anyone overrides `project-key` with a value containing backslashes or double quotes, the query breaks. Use `strconv.Quote` semantics or pass unquoted since Jira JQL keys don't need quoting anyway.
- `EqualFold` on transition names is the right call. Good.

**Pagination correctness**
- The loop terminates when `len(searchResponse.Issues) == 0 || startAt >= searchResponse.Total`. Correct, but in the rare case that Jira returns `startAt + len < total` but no matching summary (i.e. page had only non-matching issues), the loop correctly moves to the next page. That's fine.
- `maxResults=100` is a reasonable cap. Jira's hard ceiling is 1000 on `/rest/api/2/search`; using 100 keeps the response time predictable.

**Code style / minor**
- `readErrorBody` is good defensive programming. 1024-byte cap is reasonable.
- `sendRequest` uses `client.GetHTTPClient(nil)` — consistent with the rest of the codebase, but means the Jira request timeout, retries, and TLS settings all inherit whatever Gatus configures globally. That's probably fine, but if Jira's slow response on the search endpoint causes cascading delays in alert delivery, it might be worth a dedicated client with a tighter timeout. Pre-existing concern with the pattern, not specific to this PR.
- `defer response.Body.Close()` after `json.NewDecoder(...).Decode()` is correct — but note that `json.NewDecoder` reads the body, so closing immediately after is fine.
- The `summary` collision logic is exact-match (`issue.Fields.Summary == summary`). Two endpoints with the same `DisplayName()` would re-resolve each other's issues. The unique-by-displayName collision is unlikely but real — if you ever add a salt (e.g. endpoint key or hash of the URL), do it now while there's only one consumer.

**Tests**
- Coverage of happy paths, error paths, and missing transitions is solid.
- No test for the case where `sendRequest` returns an error (network failure). The "resolved-error-searching" scenario hits the 500 path on the server, not a transport error. Worth one scenario with a `RoundTripper` that returns an error to lock in the early-return.
- The `TestAlertProvider_GetConfig` tests don't verify the `Username` or `Token` overrides propagate — only `ProjectKey` and `IssueType`. Add a quick assertion for those.

**Docs**
- README section is good. The example shows a `failure-threshold: 2` which conflicts with the default `issue-type: Task` not being a "Bug" — Jira projects often restrict issue types by issue type scheme. Worth a sentence: "ensure the configured Jira project allows creating issues of the specified `issue-type`".

Overall a solid addition. The main concerns are the HTTPS-not-enforced footgun, the resolve-loop correctness around non-`Done` status categories, and the exact-match collision risk.

TwiN:
@ChrisJr404 Sorry for the delay, I'd rather just get everything done in one go rather than splitting it in multiple PRs.

Ultimately, I do not use Jira for my personal projects, so I won't run into these issues, but will you? And if so, is it worth addressing? 

If it's not worth addressing, then I'm ok with just merging this as-is.

PR #1782, opened 2026-08-27
Expose certificateExpiration and domainExpiration in JSON API

TN-1:
## Summary

Expose CertificateExpiration and DomainExpiration fields in endpoint.Result JSON serialization with omitempty tags so that downstream consumers and API clients can access them.

## Checklist

### DIFF
--- a/api/endpoint_status_test.go
+++ b/api/endpoint_status_test.go
@@ -39,6 +39,7 @@ var (
 		Timestamp:             timestamp,
 		Duration:              150 * time.Millisecond,
 		CertificateExpiration: 10 * time.Hour,
+		DomainExpiration:      720 * time.Hour,
 		ConditionResults: []*endpoint.ConditionResult{
 			{
 				Condition: "[STATUS] == 200",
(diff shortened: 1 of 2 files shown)


wahajahmed010 (commented):
test

wahajahmed010 (commented):
## Code review

Small, focused change — adds JSON serialization for `CertificateExpiration` and `DomainExpiration` so API consumers can read them. The test diff is consistent with the fixture change (added `DomainExpiration: 720 * time.Hour` to the success endpoint, which now serializes; the failure endpoint still has the zero value and correctly omits both fields).

A couple of items worth checking before merge:

### 1. Checklist says README was updated, but no docs change in this PR

The contributor's checklist is checked:

> - [x] Updated documentation in `README.md`, if applicable.

…but `git diff` only touches `api/endpoint_status_test.go` and `config/endpoint/result.go`. Either the doc update was forgotten, or it was already on `main` from a previous PR. Worth confirming — if this is a public API change, downstream consumers will want to know what units the durations are in (Go `time.Duration` JSON serializes as int64 nanoseconds, which is not human-readable; consider documenting `"nanoseconds"` or converting to a friendlier unit like ISO-8601 or a seconds-since-epoch int).

### 2. `omitempty` on `time.Duration` semantics

```go
DomainExpiration time.Duration `json:"domainExpiration,omitempty"`
```

For most endpoints `DomainExpiration` will be zero (because domain-expiration monitoring is opt-in), so `omitempty` is correct — you don't want every response cluttered with `"domainExpiration":0`. Just double-check that no caller today distinguishes "unset" from "zero/expired." If they do, this change is a breaking change for them (because previously the field was absent for both states, and now it's absent only when the duration is exactly zero).

### 3. Negative durations (past expiration) will still be serialized

If a domain has already expired, `DomainExpiration` will be negative (e.g., `-24h` = `-86400000000000` nanoseconds). With `omitempty`, a negative value is non-zero and gets serialized. That is probably the right behavior, but it can be surprising for API consumers who don't expect negative durations. Consider whether expired endpoints should surface a different signal (e.g., a separate `domainExpired: true` boolean, or clamping the duration to zero on serialization).

### 4. Backward compatibility

Adding new fields to a JSON response is non-breaking for any consumer that ignores unknown fields (the universal JSON contract). If any consumer is using a strict schema validator (e.g., JSON Schema with `additionalProperties: false`, Protobuf with strict parsing, or a typed deserializer that fails on unknown keys), this is a breaking change. Worth a release note in CHANGELOG / GitHub release.

### 5. Tests are good

The fixture update (`DomainExpiration: 720 * time.Hour`) flows through all five test cases correctly. The pagination tests correctly show only the success result has `domainExpiration` and the failure result omits it. Good.

Overall: this is a clean, well-tested additive change. Address the missing README/docs update (item 1) and the negative-duration consideration (item 3) before merging and you're good.

> Automated review posted via the open-source-contributor pipeline.

wahajahmed010 (commented):
This is a clean change. Two unrelated observations that came up while reading the diff:

**Serialization ordering (functional)**
The test expectations in `api/endpoint_status_test.go` rely on Go's struct-field ordering in the JSON output. Since `CertificateExpiration` and `DomainExpiration` are the last two fields in `Result`, that's currently stable — but worth noting because anyone reordering `Result` fields later will silently break this test. The expected-body strings are essentially snapshot tests; a comment near the test table pointing out the coupling would help future contributors. Even better would be unmarshalling the response and asserting only on the new fields.

**`omitempty` on `time.Duration`**
`omitempty` on a `time.Duration` field only omits the zero value (`0`). For most endpoints that never run an SSL check or whois lookup, the field will still serialize as `"certificateExpiration": 0`, which is noise for consumers who don't care. Consider whether you'd rather drop the field entirely from the API when the underlying check didn't run (e.g. pointer types or a sentinel like `-1`). It's a minor concern — the current behavior is at least explicit and machine-readable.

**Tests**
The existing test table updates look mechanical and correct. No new test was added specifically for the serialization behavior; if the project grows, having a focused test that asserts the JSON shape (rather than whole-string snapshots) would be more resilient to unrelated field additions.

Overall, looks good. The change is small, additive, backward compatible, and well-motivated by downstream consumer demand.

TN-1:
@wahajahmed010 I have updated the readme, I believe that addresses your concerns

PR #1785, opened 2026-08-27
Add Signalgrid alerting provider

signalgridco:
## Description

Adds Signalgrid as a native alerting provider.

The provider supports:

- Sending triggered alerts through the Signalgrid API
- Sending resolved notifications
- `CRIT` type for triggered alerts
- `SUCCESS` type for resolved alerts
- Optional critical notifications for triggered alerts
- Provider overrides
- Default alert configuration
- Configuration validation
- Unit tests

Critical notifications are only enabled for triggered alerts. Resolved notifications are always sent as non-critical.

The README has also been updated with configuration documentation and an example.

## Testing

The provider was tested against the Signalgrid API with both triggered and resolved Gatus alerts.

The full test suite passes:

`go test ./...`


wahajahmed010:
Thanks for the clean provider addition. The structure mirrors the existing `signal` / `signl4` providers, which keeps the surface area consistent for users. A few observations:

1. **`client.GetHTTPClient(nil)`** — using the shared client means timeout, retries, and proxy settings come from the global config. If `cfg.ClientKey` is malformed (e.g., contains a newline from a copy-paste error), the form-encoded payload will silently truncate or reject without a clear validation path. Worth adding a fuzz-style test that round-trips `buildRequestPayload` through `url.Values.Encode()` to confirm the body is valid for any string inputs.

2. **`provider.Send` body** — when `response.StatusCode > 399`, you return an error that includes the response body. If the Signalgrid API ever returns a non-UTF-8 error body or a very large one, `string(body)` will produce invalid UTF-8 in the error message. Minor, but `body` is read with `io.ReadAll` with no cap; consider `io.LimitReader(body, 4096)` before formatting.

3. **Resolved critical override** — line `critical = false` when `resolved` is true is the right behavior. The test `TestAlertProvider_buildRequestPayload_Resolved` covers this. Good. But the comment "Resolved notifications should never be sent as critical notifications" is hardcoded — if the Signalgrid API ever allows resolved-as-critical, this would need a config flag. Not blocking, just flagging.

4. **Tests look complete** — `Validate`, `buildRequestPayload` (Triggered/Resolved/NonCritical), `GetConfig` (default + override), and `GetDefaultAlert` are all covered. The override test correctly unmarshals `ProviderOverride` into `Config`. The `boolPtr` helper is clean.

5. **README** — the new section follows the alphabetical ordering of the existing config table and uses the same field naming style. Consistent. The example YAML uses `critical: true`, which is the new behavior — might be worth calling out in the PR description that this is the first default-on critical flag among the providers, since users copy-pasting examples may not expect push notifications to bypass DnD.

Approving. Solid work.

PR #1786, opened 2026-08-27
test(config): make HasLoadedConfigurationBeenModified deterministic

nikolauspschuetz:
`TestConfig_HasLoadedConfigurationBeenModified` slept a hard `time.Sleep(time.Second)` in each of its two subtests, purely so a rewritten file's mod time (second precision) would tick past the load time — ~2s of dead wall-clock per run, coupling the assertion to real elapsed time rather than the code under test.

This advances the modified file's mod time explicitly with `os.Chtimes(...)` instead, and gives the directory subtest its own `t.TempDir()` — otherwise the file subtest's now-future-dated `config.yaml` (the two subtests shared one dir) leaks in and makes the directory subtest's *initial* "not modified" check fail.

Result: no sleep, no wall-clock dependence, deterministic — verified with `go test ./config/ -run TestConfig_HasLoadedConfigurationBeenModified -race -count=10` (10/10 pass).

<sub>Developed with AI assistance (Claude Code); I directed, reviewed, and verified it locally.</sub>


TwiN:
Makes sense to me. Ping me when it's no longer in draft

nikolauspschuetz:
@TwiN out of draft now.

Verified in `golang:1.26.3`: `go test ./config/ -run TestConfig_HasLoadedConfigurationBeenModified -race -count=10` — 10/10 pass, 1.0s total (was ~2s of sleep per run).

wahajahmed010:
This is a clear improvement — `time.Sleep(time.Second)` was the wrong tool for a sub-second-precision assertion and the `t.TempDir()` fix is a real correctness bug, not just a tidy-up.

Three notes:

**1. The directory subtest's `t.TempDir()` is a correctness fix that deserves a note in the commit message.** Before this change, the two subtests shared one directory, and the second subtest inherited a `config.yaml` whose mod time was now in the future (because the first subtest had advanced it). Whether that caused a real flake depended on filesystem mtime resolution and Go's `os.Chtimes` rounding — but `t.TempDir()` is the right thing regardless, and the commit message currently buries that under "deterministic." Suggest splitting the wording so a future grep for `TempDir`/`flaky` picks it up.

**2. `future := time.Now().Add(2 * time.Second)` — consider using `time.Now().Add(time.Second + buffer)` rather than 2× the original sleep.** The original was 1s, you bumped to 2s. On filesystems with low mtime resolution (ext4 with `nsec` enabled is fine, but FAT/SMB shares are not) `Add(2 * time.Second)` still gives a comfortable margin. Two follow-ups:
- On filesystems where `os.Chtimes` rounds down to the nearest second, you may need to add a small additional margin (e.g. `time.Second + 100*time.Millisecond`) to guarantee a tick. Not strictly required for this codebase, but worth a one-line comment if you have seen mtime flakiness on CI.
- Optional: a comment that this is intentional (`// plus 1s slack to clear second-precision filesystem rounding`) helps the next reader.

**3. The new `dir := t.TempDir()` shadow.** The outer scope defines `dir` earlier in the same `func TestConfig_HasLoadedConfigurationBeenModified(t *testing.T)`. The new block introduces a second `dir` local to the subtest. Go allows this and it is what you want, but a quick `// shadowed intentionally — each subtest gets its own temp dir` line removes the "wait, was this deliberate?" pause for future readers.

**4. The two `time.Sleep` deletions are good, but the two `t.Fatalf` messages were also updated to reflect the new semantics ("overwrite" → "modify" / "add").** That is the right change; `os.WriteFile` to an existing file is *overwrite*, but to a not-yet-existing file is *add*. The fix is consistent with the new flow. No further action.

**5. No test for the "directory does not exist" path.** The first subtest still uses the original (non-tempdir) directory created by the test setup. If `configFilePath` is built from `filepath.Join(dir, "config.yaml")` and `dir` got cleaned up by an earlier subtest failure, the next run could error in `os.WriteFile`. Probably not an issue given `t.TempDir()` is in the second subtest and the first uses the outer `dir`, but worth a quick mental trace.

Ship it. The Sleep removal alone makes this worth merging, and the TempDir fix is the kind of thing I would not have noticed in a quick read — exactly what a focused review is for.

TwiN (approved):

TwiN:
@nikolauspschuetz Thank you for the contribution!

PR #1788, opened 2026-08-29
feat(storage): add optional buffered sqlite storage with periodic flush

GRbit:
Hi again TwiN!

TLDR: in-memory storage with periodic flushes on disk to make SD live longer. Feel free to close if it seems redundant.

 It's great to see the project is alive and growing. There are so many great projects that are amazing as they are but could benefit from a community sooo much. However, they are no longer maintained =(
 I don't blame the authors, it's hard work, especially nowadays with LLM-generated slop PRs. So, despite this PR is written with Claude, I try to make the contribution itself a human-to-human interaction.

My best wishes of a great work/life/open-source balance to you!

Getting to the point of the PR: I run the service on a small OrangePi + SD card, which runs out of its resource pretty quickly. KUDOS for making in-memory mode for such machines, it's a great feature! However, I also want some historical data, and this PR is an attempt to sit on two chairs with one butt.

I add a wrapper for the current SQLite storage which is a combo in-memory SQLite and on-disk SQLite. Implementation is plain simple: read DB from disk, accumulate changes in memory, flush it back every `flush-interval`. Of course there is a flush on SIGTERM/SIGINT. And a write attempt on start just to check that we actually can do it and fail-fast. The only case to lost up to `flush-interval` amount of data is a power loss or SIGKILL.

Also, since it's still SQLite DB on a disk, you can switch back to original SQLite driver any time.

Here is a config example:
```yaml
storage:
  type: sqlite        # buffered is sqlite-only
  path: data.db
  buffered: true      # default false
  flush-interval: 1h  # default 10m, minimum 1m
```

An important thing to know: each flush completely replaces the database file (VACUUM INTO a temporary file, fsync, rename). So, you can imagine a situation with 100MB database when flushing it once every minute is actually a bigger pain for SD rather than every second small writes on services check (which will be ~4mb FS sector anyway). This balance is controlled via `maximum-number-of-results` + `maximum-number-of-events` vs `flush-interval`. Increasing retention increases every flush, so the `flush-interval` should be tuned as well.

Advocating for not merging this code:
* I know that this could be solved with some other magic, such as systemd wrapped with in-memory disk flushing periodically, but I find Go so convenient in re-implementing  everything in a portable binary, that I just can't stand it. I think it's awesome to have features out of the box on bare minimum distros without systemd.
* New code needs maintenance. I'll be glad if you will request any bug-fixes from my side and be happy to contribute more. But it's your project, your responsibility, so if you hesitate, better close this. Nowadays it's cheap to write code, but it's expensive to maintain clean, robust solutions. I think we all should favor the latter.

If you read to this place – WOW! I'm tremendously grateful! Considering you have 255 issues (what a nice number!) and 124 PRs open PRs (soon to be 127, lol), it just warms my heart if you have time to read mine.

Thank you! Hope this PR description made you smile at least once :heart:


TwiN:
I understand and appreciate your PR as well as how thorough both the change and the description are (which makes it all the more difficult to say the next part).

I don't think Gatus is write-heavy enough to warrant the additional complexity that your PR adds to the project. It's an excellent idea, but I have designed Gatus to minimize resource usage, which includes to some extent writes (e.g. by bulking deletes as opposed to deleting every time a write happens, by adding a write-through cache, etc.)

I know it's not the best answer, but if you're worried about the longevity of your SD card, there are SD cards specifically designed for write-intensive operations, and these days, SD cards are dirt cheap.

Again, though, beautiful PR. I am just more selective with changes these days and I think this type of change may make it harder to investigate issues and maintain the code base

GRbit:
Thank you for the review and for the comment, I really appreciate it! I can agree with your general view, each new feature makes projects heavier, while we need it to breathe easily to live long and prosper.

Thank you again for making it and good luck with your gatus.io endeavor!

PR #1791, opened 2026-09-01
fix: restore consecutive failure/success counters on config reload

luca-bortoletto-vg7:
## Summary
- `ep.NumberOfFailuresInARow` / `ep.NumberOfSuccessesInARow` only ever lived in memory on the `Endpoint` struct (`yaml:"-"`). On a config file reload (or restart), brand-new `Endpoint` structs are created and these counters reset to `0`.
- With a persistent storage provider (sqlite/postgres/mysql), the evaluation history needed to reconstruct these counters is already there — `initializeStorage` just wasn't using it, except in the narrow case where an alert had already triggered, and even then it approximated `NumberOfFailuresInARow` as the alert's `FailureThreshold` rather than the real count.
- Net effect: an endpoint that's, say, 2 failures into a 3-failure alert threshold silently loses that progress on every config reload, requiring 3 fresh consecutive failures instead of 1 more.
- Adds `restoreNumberOfEvaluationsInARow`, which recomputes both counters from the endpoint's persisted results by walking the most recent results backwards for as long as they share the same outcome. Runs for every endpoint (including external endpoints and suite endpoints) before the existing triggered-alert restore logic, which now only restores `Triggered`/`ResolveKey`.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (full suite passes)
- [x] Added `main_test.go` covering: counters restored after N consecutive failures with no history loss, trailing successes after failures, and the no-persisted-history no-op case


wahajahmed010 (commented):
Thanks for the fix — restoring the in-a-row counters from persisted history is the right approach. A few observations:

**1. Unused return value on `GetTriggeredEndpointAlert`**

The PR stops using `numberOfSuccessesInARow` returned from `store.Get().GetTriggeredEndpointAlert(...)` in all three callers (endpoint/external-endpoint/suite-endpoint). The function still has the signature:

```
GetTriggeredEndpointAlert(ep *endpoint.Endpoint, alert *alert.Alert) (exists bool, resolveKey string, numberOfSuccessesInARow int, err error)
```

After this change the third return is dead. Consider either:
- Dropping it from the interface (cleanest — but it touches every store implementation), or
- Leaving a `// nolint:unused` / doc comment indicating it's kept for backward compatibility.

Otherwise this becomes an undocumented wart that future readers will be confused by.

**2. Memory/IO cost at startup**

`restoreNumberOfEvaluationsInARow` runs for every endpoint (incl. external + suite endpoints) and pulls up to `cfg.Storage.MaximumNumberOfResults` results per endpoint. For a deployment with hundreds of endpoints and a large `MaximumNumberOfResults` (e.g. 1000+), this is a non-trivial synchronous I/O burst on every boot/config-reload. Two thoughts:
- A short comment in `initializeStorage` noting this scales with `len(Endpoints) * MaximumNumberOfResults` would set expectations.
- For the suite-endpoint loop, the inner result walk happens per endpoint inside a per-suite loop — fine functionally, but worth confirming the store implementations batch reasonably per call.

**3. Missing test case**

The three added tests cover: 2-failure trailing, 3-failure-then-2-success trailing, and empty history. The symmetric case is missing:

```
results := []bool{true, true, false}  // 1 trailing failure
```

Expected: `NumberOfFailuresInARow == 1`, `NumberOfSuccessesInARow == 0`. Trivial to add and confirms the last-segment-walk logic works for both outcomes.

**4. Minor: comment placement**

The restore loops in `initializeStorage` are inserted *after* the "Deleted N endpoint statuses" log but *before* the triggered-alert restore. This is the correct ordering (counter restore must run before any logic that reads them), but it's not obvious from a glance. A one-line comment like `// must run before the triggered-alert restore below` would help.

Otherwise the change is well-scoped, the tests are clear, and the PR description matches the diff. Good work.

wahajahmed010:
## Review

This PR fixes a real regression in the alert-threshold logic: after a config reload or restart, the in-memory `NumberOfFailuresInARow` / `NumberOfSuccessesInARow` counters reset to zero even when the storage provider already has the history to know how many evaluations in a row actually failed or succeeded. The fix walks the persisted results backwards and re-derives the counters — which is correct in principle.

A few things worth tightening before this lands:

1. **Silent removal of the previous counter assignment in `GetTriggeredEndpointAlert`.** Before this PR, the three loops in `initializeStorage` set `ep.NumberOfSuccessesInARow, ep.NumberOfFailuresInARow = numberOfSuccessesInARow, alert.FailureThreshold` based on what the storage layer reported. That code is now deleted and replaced by `restoreNumberOfEvaluationsInARow`. The new helper only sets **one** of the two counters (whichever matches `lastResultSuccess`), so the *other* counter stays at its zero value. That's actually fine for the success-then-failure transition case (the new helper overwrites it on the next reload), but in the **mid-loop** case where the alert is currently triggered and the most recent result is a failure, the previous code would have set `NumberOfSuccessesInARow = numberOfSuccessesInARow` from the storage layer. The new code only sets `NumberOfFailuresInARow`. The previous *successes-in-a-row* value is lost until the next reload — which is a behavior change for any consumer reading `NumberOfSuccessesInARow` between the reload and the next evaluation. Worth either (a) initializing both counters in the helper (read the second-to-last result's streak if needed) or (b) explicitly documenting this trade-off.

2. **No concurrency guard.** `restoreNumberOfEvaluationsInARow` writes to `ep.NumberOfFailuresInARow` / `NumberOfSuccessesInARow` from the main goroutine during `initializeStorage`. The watchdog reads/writes the same fields on a per-endpoint tick goroutine. If a watchdog tick is in flight when this code runs (e.g. a config-reload signal handler that doesn't quiesce the watchdog first), this is a data race. The previous code had the same exposure, so it's not a regression — but since you're touching this area, `sync.Mutex` per-endpoint or a `runtime.Gosched()`-style handoff would close it.

3. **The `lastResultSuccess` variable name is misleading.** `lastResultSuccess` is a `bool` indicating whether the most recent evaluation succeeded; it's not the success value of the previous result. Reading the loop, the variable is doing the job of "outcome of the most recent evaluation", which is the streak terminator. Renaming to `lastOutcome` or `streakOutcome` would help.

4. **Test naming.** `TestRestoreNumberOfEvaluationsInARow` and the two siblings are clear, but the first one only tests the all-failures case with `len(results) == 2`. Adding a case where the streak ends exactly at `maximumNumberOfResults` (boundary) would catch an off-by-one if the storage layer ever changes its semantics.

5. **`cfg.Storage.MaximumNumberOfResults` is fetched from the user's config but passed straight into `paging.WithResults`.** If a user sets this to a very large number (say, 10000), you're walking all 10000 results on every restart to find the streak — O(n) per endpoint, O(n×endpoints) at boot. Capping at e.g. 2× the highest `FailureThreshold` across the endpoint's alerts would keep it bounded.

No blockers on the core fix. The first observation is the one I'd want addressed before merge.

luca-bortoletto-vg7:
Thanks for the thorough review, addressed in c4ed14f3:

1. **Other counter left at zero after reload while alert is triggered** — I traced through `watchdog/alerting.go`: `handleAlertsToTrigger` always sets `NumberOfSuccessesInARow = 0` on a failure, and `handleAlertsToResolve` always sets `NumberOfFailuresInARow = 0` at the end on a success. So the two counters are actually mutually exclusive at runtime — only one is ever non-zero at a given moment — which is why `restoreNumberOfEvaluationsInARow` only needs to set the one matching the most recent result; the other one is correctly left at its zero value. I added a comment on the function documenting this invariant so it's not a silent assumption. The old code's `numberOfSuccessesInARow` from `GetTriggeredEndpointAlert` was actually stale in the scenario you describe (persisted via `UpsertTriggeredEndpointAlert`, which isn't called on every failure), so the new approach is strictly more correct there, not a regression.
2. **Concurrency guard** — agreed this is pre-existing exposure, not introduced by this PR; left as-is per your note that it's not a blocker.
3. **`lastResultSuccess` naming** — renamed to `streakOutcome`.
4. **Boundary test** — added `TestRestoreNumberOfEvaluationsInARowStreakAtWindowBoundary`, covering a streak exactly as long as the lookup window.
5. **Unbounded `MaximumNumberOfResults` fetch** — bounded the lookup window to the endpoint's largest configured alert threshold (`largestAlertThreshold`) instead of always fetching up to `cfg.Storage.MaximumNumberOfResults`.

While tightening the fetch window I also caught a real bug of my own: `ExternalEndpoint.ToEndpoint()` returns a new `*Endpoint`, so the counters computed for external endpoints were being written to a throwaway copy and never actually applied. Fixed by copying the result back onto the `ExternalEndpoint`, with a regression test (`TestRestoreNumberOfEvaluationsInARowForExternalEndpoint`).

luca-bortoletto-vg7:
@wahajahmed010 also following up on your earlier review (the 4-point one):

1. **Unused `numberOfSuccessesInARow` return from `GetTriggeredEndpointAlert`** — left the interface as-is; changing its signature would touch every `Store` implementation (memory + sql) for a return value that's still meaningful to the interface's contract even though this particular PR's callers don't need it. Happy to split that into a separate cleanup PR if you'd rather see it gone.
2. **Memory/IO cost at startup** — addressed above: the lookup is now bounded per-endpoint by its largest alert threshold rather than `cfg.Storage.MaximumNumberOfResults`, so it no longer scales with that setting.
3. **Missing symmetric test case** — added `TestRestoreNumberOfEvaluationsInARowWithTrailingFailureAfterSuccesses` (`{true, true, false}` → `NumberOfFailuresInARow == 1`).
4. **Comment on restore-loop ordering** — added a note in `initializeStorage` that the restore must run before the triggered-alert restore.

PR #1792, opened 2026-09-02
fix(watchdog): seed first check delay from last persisted result

Troyhy:
On restart or configuration reload, Gatus runs a health check against **every** endpoint at the same instant and rebuilds each endpoint's state from scratch. Two problems:

1. **All endpoints are checked at once on restart/reload.** Every monitor loop re-checks its endpoint immediately instead of resuming on its own schedule, so every endpoint (plus suites and external-endpoint heartbeats) fires simultaneously the moment the process starts or the config reloads — a burst of outbound requests, and a schedule that restarts from "now" so long-interval endpoints can be starved by frequent reloads. For external-endpoint heartbeats it's worse: the immediate re-check masks genuinely overdue pushes (reports false healthy) and can fire false "missed" alerts after a container restart.

2. **Streak counters are lost.** Only the "already triggered" alert state survived a restart; the consecutive-failure counter was reconstructed as exactly `FailureThreshold` (otherwise zero), so partial progress toward an alert (e.g. 2 of 3 failures) was discarded on every restart/reload and had to rebuild from scratch.

## Fix

**Schedule seeding** (`9b551e8`) — seed each monitor loop's first health check from the last persisted result timestamp:
- overdue, or no history (e.g. the in-memory store) → run immediately (unchanged)
- otherwise wait `interval - elapsed`, then resume the normal cadence

Applies to regular endpoints, suites, and external-endpoint heartbeats via two shared helpers (`endpointInitialDelay` / `suiteInitialDelay`).

**Streak backfill** (`600e4dd`) — on startup, for each regular/external endpoint with an enabled alert, read the newest N persisted results (N = max failure/success threshold) and reconstruct `NumberOfFailuresInARow` / `NumberOfSuccessesInARow` via `endpoint.NumberOfResultsInARow`. Triggered state still comes from `endpoint_alerts_triggered`. Suite endpoints are out of scope (their results are stored under `suite_result_id` and aren't retrievable via `GetEndpointStatusByKey`).

No schema or storage-interface changes. No-op for the default in-memory store.

Related: TwiN/gatus#1237, TwiN/gatus#1718.

## Checklist


wahajahmed010:
Solid write-up and the design (two helpers, one new exported function in `config/endpoint`) is clean. The streak backfill tests are thorough. One substantive correctness issue and a couple of smaller things:

**1) Storage ordering mismatch (likely correctness bug).** The new code assumes `status.Results` is ordered "oldest-first, newest-last" (per the docstring on `NumberOfResultsInARow` and the inline comment `// results are newest-LAST`). That matches the in-memory store (`AddResult` does `ss.Results = append(ss.Results, result)`), but the SQL store in `storage/store/sql/sql.go` does:

```
ORDER BY endpoint_result_id DESC
...
results = append([]*endpoint.Result{result}, results...)
```

That prepends each row to the slice, so the SQL store returns **newest-first, oldest-last** — the opposite of the in-memory store and the opposite of what this PR assumes.

Concrete consequences on a SQL/Postgres-backed deployment:

- `endpointInitialDelay`: `lastResult := status.Results[len(status.Results)-1]` is the *oldest* result on SQL, not the newest. The seed delay will be computed from a stale (oldest) timestamp, which will (a) make every endpoint wait almost an entire interval longer than it should on every restart, and (b) reliably mis-seed the schedule for low-interval endpoints where the elapsed gap between oldest and newest of a length-`n` window matters.
- `NumberOfResultsInARow`: walks the slice from `len-1` down to `0`, so on SQL it counts the trailing run of the *oldest* N results, not the newest. With `n = max(FailureThreshold, SuccessThreshold)`, this can both (a) under-count a real streak (if more than N results have piled up since the most recent failure run started) and (b) over-count by counting older successes/failures that aren't part of the current trailing run.

The unit tests in `streak_test.go` only exercise `NumberOfResultsInARow` in isolation with hand-built slices, so they don't catch this. The schedule-seed tests only hit the in-memory store and likewise don't catch it.

Two ways to fix: (a) sort/normalize in the helper before consuming `status.Results` (cheapest, works for both backends), or (b) fix at the storage layer so both `GetEndpointStatusByKey` paths return the same canonical order and update the PR's docstring/comments to match.

**2) Pagination semantics for suites.** `suiteInitialDelay` uses `paging.NewSuiteStatusParams().WithPagination(1, 1)` — same `(page, pageSize)` semantics as `WithResults`, but `WithPagination` defaults `PageSize` to 20 in `NewSuiteStatusParams` while `WithResults` leaves `ResultsPageSize` at 0 (caller sets it). Confirm via the SQL impl that `getSuiteResultsByEndpointID`-equivalent honors `Page=1, PageSize=1` and returns the newest results. If the suite SQL path uses the same DESC-then-prepend pattern as endpoints, the same ordering bug from (1) applies.

**3) `NumberOfResultsInARow` for the `alternating ending in failure` case.** `{Success: true}, {Success: false}` returns `(1, 0)` — that's correct (one trailing failure). But the function never reads `results[0]` if `len == 1`, so a single-result case is fine; just noting that the docstring "Exactly one return value is non-zero" only holds for `len > 0` — empty input returns `(0, 0)`, which contradicts "exactly one ... non-zero". Tighten the docstring to "for non-empty input, exactly one return value is non-zero".

**4) `external_endpoint.go` placement of the first execute.** You moved the initial-delay block above `ticker := time.NewTicker(...)` and then call `executeExternalEndpointHeartbeat(ee, ...)` before the ticker is created. That's intentional and matches `endpoint.go`, but worth a one-line comment noting that this differs from a naive read — the first execution is one-shot, not driven by the ticker. Future readers may try to "simplify" by deleting the explicit first execute.

**5) Minor:** in `main.go`'s loop computing `n`, you skip disabled alerts (`if !a.IsEnabled() { continue }`). That's fine for backfill, but note that a disabled alert still doesn't get its counter set — which is intentional, just calling out that `NumberOfFailuresInARow` is left at zero for disabled alerts. The downstream alert-evaluation code should already tolerate this, but worth a comment.

**6) `attributeFilter: ['style']`** — not relevant here, ignore.

Overall the streak fix is the right call, just needs to be order-agnostic across storage backends before it lands.

wahajahmed010:
Solid fix. The streak backfill and the watchdog delay seeding are both small, well-tested changes, but I think there is a subtle bug in the external-endpoint branch of `initializeStorage` and a behavioural change worth confirming.

**Logic — `initializeStorage` external branch**
```go
if status, err := store.Get().GetEndpointStatusByKey(convertedEndpoint.Key(), paging.NewEndpointStatusParams().WithResults(1, n)); err == nil && status != nil {
    ee.NumberOfFailuresInARow, ee.NumberOfSuccessesInARow = endpoint.NumberOfResultsInARow(status.Results)
}
```
The query uses `convertedEndpoint.Key()`, but everywhere else in this loop you key off `ee.Key()`. If `ExternalEndpoint.ToEndpoint()` returns a key that differs from `ExternalEndpoint.Key()` (e.g. name normalisation, group suffixing, anything non-trivial), you are seeding the wrong counter row from the wrong endpoint's history. The non-external branch correctly uses `ep.Key()`. Please align the two: either both call sites should use the converted key or both should use the raw `ee`/`ep` key, with a comment explaining which is canonical.

**Logic — `monitorExternalEndpointHeartbeat` is now restructured**
Before this PR, the function was:
```go
ticker := time.NewTicker(ee.Heartbeat.Interval)
defer ticker.Stop()
for { select { case <-ticker.C: ... } }
```
which means the first execution happened after one full `Heartbeat.Interval` (ticker fires at t+interval, not t+0). The new code adds a delay block, then runs `executeExternalEndpointHeartbeat` once, then enters the ticker loop. Two consequences worth confirming:

1. The first execution now happens at `lastResult.Timestamp + Heartbeat.Interval` instead of `Heartbeat.Interval` after start. That is almost certainly what you want for a restart, but the `monitorEndpoint` and `monitorSuite` paths in this PR make the same change, so the behaviour across all three watchdog functions is now uniform — good.
2. `monitorEndpoint` previously had `executeEndpoint` followed by the ticker loop, with no initial delay. The PR keeps that pattern but prepends a delay block. So the very first execution shifts from "immediately on start" to "wait until next scheduled tick". For endpoints with a fresh persisted history this is correct; for endpoints with no history (`endpointInitialDelay` returns 0) the `if delay > 0` guard preserves the old behaviour. Worth a unit test for the no-history case to lock that in.

**Performance**
The new flow runs an extra `GetEndpointStatusByKey` per endpoint (and per external endpoint) at startup. For large configs this is one DB round-trip per endpoint. If startup latency matters, consider a single bulk query that returns the last result timestamp for every key in one shot. Not blocking — the change is correct as written.

**Error handling**
`store.Get().GetEndpointStatusByKey(...)` errors are swallowed with `err == nil`; `NumberOfResultsInARow` is just not seeded. Fine for a degraded startup, but a `logr.Warnf` would help anyone debugging a storage backend that is returning errors at boot.

**Tests**
- `NumberOfResultsInARow` has good table coverage, including nil/empty/all-same/alternating cases.
- `TestEndpointInitialDelay` and `TestSuiteInitialDelay` cover the three branches (no history, overdue, recent). One missing case: history exists but `lastResult.Timestamp + interval` is in the past by a tiny margin (e.g. 1ms overdue due to clock skew). Expected behaviour: 0 delay. Worth a sub-test so a future "always wait at least X ms" change is caught.
- No test exercises the `monitorEndpoint` / `monitorSuite` restructure. Add at least one test that asserts the first `executeEndpoint` happens after the seeded delay when there is history, and immediately when there is not.

**Nits**
- `endpointInitialDelay` and `suiteInitialDelay` are nearly identical; consider a generic helper that takes a `func(string, paging.X) (*Status, error)` lookup. Skip if you prefer explicit per-typed wrappers.
- `logr.Warnf("[watchdog.monitorEndpoint] Canceling current execution of group=%s; endpoint=%s; key=%s", ep.Group, ep.Name, ep.Key())` — the message says "Canceling current execution", but at this point the first execution has not started yet, only the pre-execution wait. Slightly misleading. Consider "Canceling before first execution".

Troyhy:
@wahajahmed010 Now the reviews are addressed in multiple commits, please review changes.

PR #1796, opened 2026-09-08
fix(badge): use the endpoint name as the shields badge label (#737)

### LINKED ISSUE #737: `badges.shields` always returns `label: gatus` instead of endpoint name.
### Describe the bug

Working with Gatus to make badges, and no matter which endpoint I give it such as `/api/v1/endpoints/internal_truenas/health/badge.shields` the label is returned as `gatus` instead of the endpoint name (`truenas` in this case).

### What do you see?

Label is always returned as `gatus`. I tried different groups and different endpoints.
```
{"color":"brightgreen","label":"gatus","message":"up","schemaVersion":1}
```

I was expecting the label to equal to the value of the endpoint. Given an endpoint such as:
```
    endpoints:
      - name: truenas
        group: internal
        url: https://example.domain.com
       ...
```

Then I would expect `"label:":"truenas"`.

Look at the raw json returned from various endpoints, IE: `https://status.mydomain.com/api/v1/endpoints/internal_truenas/health/badge.svg` results in:
```
{"color":"brightgreen","label":"gatus","message":"up","schemaVersion":1}
```

When using this with `shields.io`, such as `https://img.shields.io/endpoint?url=https://status.mydomain.com/api/v1/endpoints/internal_truenas/health/badge.shields`,  it thus renders the wrong endpoint name:
image

### Version

5.9.0

### Additional information

* If I try to supply an alternate label using `&label=TrueNAS&nbsp;Storage` it doesn't work (stops at the `&nbsp;`) ... another bug?
image

* However if I use `&label=TrueNAS Storage` it does work:
image

thejdubb02:
Fixes #737.

The shields badge endpoint (`GET /api/v1/endpoints/{key}/health/badge.shields`) always returned `{"label":"gatus", ...}` no matter which endpoint was requested, because `generateHealthBadgeShields` hardcoded the label:

```json
{"color":"brightgreen","label":"gatus","message":"up","schemaVersion":1}
```

So a shields.io badge for `internal_truenas` rendered as `gatus` instead of `truenas`.

`HealthBadgeShields` already fetches the endpoint status, so this passes its `Name` through as the shields label. `Status.Name` is populated for a valid key by both stores (`NewStatus(group, name)` in the memory store, and parsed from the key in the SQL store) and is covered by `store_test.go`, so a 200 response always has the endpoint's name. Only the shields JSON is affected; the SVG badge is unchanged.

Adds `TestHealthBadgeShieldsLabel`, which asserts the shields label is the endpoint name (`frontend`) and fails without the change (it returns `gatus`). `go vet`, `gofmt`, and the full `api` test suite are clean.


TwiN:
This isn't a bug, it's on purpose.

Think a repository with a badge of the health of an application - you'd want the label of the badge to say Gatus, not the name of the endpoint.

I'd be open to making this configurable, but it should default to gatus.

Issue #1641, opened 2026-04-25
[Feature Request] Support for deep-linking via URL parameters (filter/group)


chevdor:
### Describe the feature request

Currently, the Gatus dashboard defaults to showing all configured groups and services. As the number of monitored endpoints grows, it becomes difficult to share a direct link to a specific "problem area" or a specific logical group with teammates.

I would like to be able to provide a URL that automatically applies a filter or focuses on a specific group/service upon page load.

### Why do you personally want this feature to be implemented?

Support for URL query parameters that interact with the existing UI search/filter functionality. 

For example:
- https://gatus.example.com/?service=baz —Show only the `baz` service. 404 if it does not exist
- https://gatus.example.com/?group=foo — Automatically collapses or hides all other groups except "foo".
- https://gatus.example.com/?search=bar — Populates the search bar and filters the view to services matching "bar".

It would be great to include the current search options.

On-call/Incident Response: A developer can link directly to the failing group in a Slack notification.
Sub-dashboards: Creating links in internal documentation (e.g., Wiki or Notion) that point to relevant service subsets without needing separate Gatus instances.

Efficiency: Avoiding manual typing in the search bar when navigating from external monitoring tools.

### Additional information

Testing current behavior shows that parameters like ?group= or ?filter= are currently ignored by the frontend. This feature would significantly improve the utility of Gatus.

amai2012:
That would be very helpful indeed.

PR #1798, opened 2026-09-09
feat(ui): sync dashboard options with query parameters

### LINKED ISSUE #1263: Allow passing dashboard filters as query parameters

nadircan:
## Summary

- initialize dashboard search, filter, and sort state from URL query parameters
- keep UI changes and browser back/forward navigation synchronized with the URL
- preserve local-storage and server-config fallbacks when parameters are absent or invalid
- safely handle repeated query parameters and preserve unrelated parameters
- document the supported parameters and add dependency-free unit coverage

Supported examples:

```text
/?search=backend
/?filter=failing
/?sort=group
/?search=backend&filter=failing&sort=group
```

Query parameters take precedence over browser preferences and server-side UI defaults. Search changes use history replacement to avoid one history entry per keystroke; filter and sort changes remain navigable with browser back/forward.

Closes #1263.
Partially addresses #1641.

Related prior work: #1420. This implementation is based on the current `master` and adds value validation, repeated-parameter handling, documentation, tests, and regenerated static assets. Credit to @PythonGermany for exploring the original approach.

## Testing

- `make frontend-test` (unit tests and lint; verified with Node 22 Alpine)
- `make frontend-build`
- `go test ./... -cover` (verified with the official Go Alpine image)
- built and ran the complete Gatus Docker image with deterministic healthy/unhealthy endpoints
- browser-tested initial URL hydration, UI-to-URL updates, clearing search, combined search/filter/sort, invalid and repeated values, local-storage precedence, and back navigation

The production `web/static` assets were regenerated with `make frontend-build`.

## Attribution

This pull request was prepared, reviewed, and tested by an agent using OpenAI Codex (GPT-5).


wahajahmed010 (commented):
Thanks for the PR — the URL-state sync is a nice quality-of-life improvement and the extraction into `dashboard-query.mjs` keeps the logic testable. A few substantive findings from the diff:

**1. Compiled `web/static/js/app.js` is committed in this PR.** The diff is dominated by the rebuilt bundle, which makes review noisy and risks merge conflicts. Worth confirming this asset is intentionally rebuilt as part of the change — if not, prefer splitting the source change from the build artifact commit, or generating it via CI.

**2. Potential feedback loop on URL → watcher → handler → emit path.** In `SearchBar.vue`:

- The `route.query.filter` watcher calls `handleFilterChange(normalizedValue, false)` when the URL value differs from the local ref.
- `handleFilterChange` always emits `update:showOnlyFailing`/`update:showRecentFailures` and resets state even when the *semantic* value did not change but the URL value did (e.g. user navigates `?filter=failing` → `?filter=none` and back).

This is intentional on first load (good), but it means browser back/forward over the same logical filter re-emits the same `false` events to the parent every time. Not a bug, but worth a one-line comment that the watcher fires on every navigation regardless of equality with the *last emitted* state, not just the local ref.

**3. `normalizeSearchQuery` does not trim.** A user typing `"abc "` then submitting via URL will have the trimmed-vs-untrimmed value diverge between the input field and the URL. If you want URL-stable state, consider trimming on emit (`handleSearchChange`).

**4. Search emit ordering on mount.** `onMounted` calls `emit(search, searchQuery.value)` *and* `handleFilterChange(filterBy.value, false)` *and* `handleSortChange(sortBy.value, false)`. With query params present, all three emit in sequence, so the parent (`Home.vue`) will apply search first, then a filter reset, then a sort reset. The filter reset emits `update:showOnlyFailing: false` unconditionally — if the user landed with `?filter=failing`, the ref is already `failing`, but the emit still fires. The `Home.vue` refactor in this PR does dedupe on its end (compare current state), but worth tightening.

**5. `getFirstQueryValue` does not handle non-string array elements.** `value: [null, 1, "foo"]` would skip null and 1, returning `"foo"`. Fine in practice, but the test only covers `[null, "web"]` → `"web"`. Add a test for `[1, "foo"]` and `[undefined, "foo"]` to lock the contract.

**6. Test file location is `web/app/tests/dashboard-query.test.js`, but `package.json` adds `node --test tests/dashboard-query.test.js` only.** Make sure CI is configured to run `make frontend-test` (the new target) — otherwise the new tests will be ignored by the existing CI matrix.

**7. `withQueryValue` deletes the key when value is falsy.** This is correct for `search` (clearing the input removes `?search=`), but be aware it also affects `filter`/`sort` if you ever pass them through this helper with falsy values. Today only `search` uses it via `handleSearchChange`; just flagging in case of future refactors.

Overall: ship-able with the bundle-commit question answered and at minimum the trimming call addressed. Tests are a good start — adding the array-element edge case would harden the contract.

nadircan:
Hi @wahajahmed010, thank you for taking the time to review this so thoroughly and for the kind feedback. I went through each point carefully and pushed the follow-up changes in `4323095c`.

**1. Generated `web/static` assets**

The rebuilt assets are intentional. Gatus tracks the generated production assets under `web/static/`, and `web/static.go` embeds that directory into the Go binary using `//go:embed static`. Leaving the rebuilt assets out would therefore leave the frontend source and the assets served by the binary out of sync.

This also follows the existing repository convention. Previous UI changes such as `12be9fac` (#1583), `1bf05d6b` (#1438), and `75c1b290` (#1235) committed their source changes together with the regenerated `web/static/js/app.js`. There are also dedicated asset-regeneration commits such as `d42c5f89` and `844847bb`.

The generated files were produced with the frontend build command and were not edited manually. Since the JavaScript bundle is minified into a very small number of lines, Git unfortunately displays it as a disproportionately large diff.

**2 and 4. Filter emissions during initialization and navigation**

There was no router feedback loop because route-originated changes call the handlers with `store = false`, preventing another `router.push`.

However, you were right that the reset-then-apply sequence produced unnecessary intermediate events—for example, emitting `false` and then immediately `true` for the selected filter. I replaced that sequence with one derived emission for each filter state:

- `showOnlyFailing` is emitted as `value === "failing"`
- `showRecentFailures` is emitted as `value === "unstable"`

This removes the redundant intermediate state during both initial hydration and browser back/forward navigation. I also added a comment documenting why route-driven changes update the parent without writing the same state back to the URL.

**3. Search whitespace**

Search query values now trim leading and trailing whitespace while preserving internal whitespace.

The implementation deliberately keeps the raw input value while the user is actively typing, but sends the normalized value to the URL and search event. This prevents a trailing space from disappearing while the user is in the middle of entering a multi-word query. On blur, the visible input is canonicalized as well.

I added coverage for leading/trailing whitespace, internal whitespace, and whitespace inside repeated query values. I also verified the typing behavior in the production Docker image: entering `Alpha ` keeps the trailing space in the focused input, and continuing with `API` correctly produces the multi-word search `Alpha API`.

**5. Non-string array values**

`getFirstQueryValue` already skipped non-string elements, but the contract was not fully covered. Tests now include:

- `[1, "web"]`
- `[undefined, "web"]`
- an array containing no string values

**6. CI coverage**

You were correct that the existing `test-ui` workflow only installed dependencies and built the frontend. It now runs `make frontend-test` after installation and before the production build, so the unit tests and lint checks are enforced by CI.

The new workflow runs currently require the normal maintainer approval applied to workflows from forked pull requests; they are not reporting a test failure.

**7. Falsy values in `withQueryValue`**

I tightened the helper so that it removes only explicitly empty or nullish values (`""`, `null`, and `undefined`) instead of relying on general truthiness. Values such as `0` and `false` are now preserved, with tests covering both cases.

After these changes, I reran the frontend unit tests, lint, production build, the full Go test suite, and a production Docker/browser test covering URL hydration, whitespace normalization, multi-word typing, filter/sort changes, unrelated parameter preservation, and browser navigation.

Thanks again for the detailed and constructive review.

PR #1803, opened 2026-09-14
feat: add ability to suspend endpoint/suite monitoring via config flag

luca-bortoletto-vg7:
## Summary

Adds a `suspended` config flag, distinct from the existing `enabled` flag, so that a monitor can be temporarily paused without disappearing from the dashboard or polluting the Failing/Unstable views.

**Why not just use `enabled: false`?**
- If the endpoint was never checked, `enabled: false` makes it invisible in the UI entirely — the endpoint-statuses API is purely storage-driven, so an endpoint with no persisted results just isn't in the response.
- If it *was* being checked and then gets disabled, its last known result (which could be a failure) stays frozen in the UI forever, since the API has no way to know it's since been disabled — it can keep showing up under "Failing"/"Unstable" indefinitely.

**`suspended: true` behaves differently:**
- Checks are stopped, same as `enabled: false`.
- The endpoint/suite stays visible in the dashboard with a dedicated "Suspended" badge — including endpoints that have *never* been checked (the API now synthesizes a placeholder status from the config in that case, instead of 404ing / omitting it).
- It's excluded from the Failing/Unstable filters and from the unhealthy counts on group headers.
- A new "Suspended" filter option in the search bar isolates suspended endpoints/suites.

## Changes

**Backend**
- `Suspended *bool` + `IsSuspended()` on `Endpoint`, `ExternalEndpoint`, and `Suite`, mirroring the existing `Enabled`/`IsEnabled()` pattern.
- `watchdog.Monitor` skips starting the periodic check goroutine for suspended endpoints/external endpoints (heartbeat)/suites.
- `endpoint.Status` and `suite.Status` gain a `Suspended` field (config-derived, not persisted by the storage provider).
- `api/endpoint_status.go` and `api/suite_status.go` annotate statuses returned from the store with their current suspended state, and synthesize placeholder statuses for suspended items that don't have any persisted results yet.

**Frontend**
- `StatusBadge` gains a `suspended` variant (outline badge, "Suspended" label).
- `EndpointCard`, `SuiteCard`, `EndpointDetails`, `SuiteDetails` show the suspended badge.
- `SearchBar` gets a "Suspended" filter option; `Home.vue` wires it up, excludes suspended items from the Failing/Unstable filters and from the unhealthy-count badges, while still showing them (with the badge) in the default view.
- `web/static` rebuilt via `npm run build` to match.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `go test ./...` (full suite passes)
- [x] Added Go tests: `IsSuspended()` for Endpoint/ExternalEndpoint/Suite, `ExternalEndpoint.ToEndpoint()` copies `Suspended`, and API-level tests asserting suspended endpoints/suites are annotated and get a placeholder when they have no history
- [x] `npm run lint` (frontend) — no errors
- [x] Manually ran the built binary against a config with an active + a suspended endpoint, confirmed via `curl` that the suspended endpoint is never checked (no results) but is returned with `suspended: true`, and confirmed in a browser that the dashboard shows a "Suspended" badge and that selecting the "Suspended" filter isolates it from the active endpoint


TwiN:
I understand your reasoning behind this feature, but I would consider this feature superfluous. 

The reason why the `enabled: false` feature exists is that if you want to disable an endpoint without losing the historical data, it can be done. Outside of that reason, the endpoint is effectively useless, and I can foresee people suspending endpoints because they're broken and just leaving them there forever rather than just deleting them, thinking they will "fix" it in the future.

I'm a big fan of keeping things minimalist, because the more complex things are, the more points of failure there are, so having two different ways of disabling/pausing/suspending an endpoint is something I'd rather avoid.

That being said, I wouldn't necessarily be against a configuration parameter that would allow disabled (`enabled: false`) endpoints from being displayed on the UI and/or being returned by the API. This would partially address your concern, without having to introduce an ambiguous configuration (`suspended: true` vs `enabled: false`).

PR #1369, opened 2025-10-27
fix(external-endpoint): check per-endpoint maintenance windows

zapteryx:
## Summary

#1368 

## Checklist

- [X] Tested and/or added tests to validate that the changes work as intended, if applicable.
- [X] Updated documentation in `README.md`, if applicable.

PR #1804, opened 2026-09-14
fix(maintenance): validate maintenance windows for external, DNS and SSH endpoints

### LINKED ISSUE #1551: Maintanance window on external endpoints not working
### Describe the bug

Maintenance window settings on external endpoints seem to be ignored, even if the config is deemed valid by Gatus. This leads to unnecessary alerts when monitoring services via an agent that pushes metrics.

### What do you see?

Setting a maintenance window on external endpoints doesn't disable alerts.

Maintenance windows should disable alerts independent of the endpoint type.

1. Configure an external endpoint with a maintenance window
2. Set up alerts
3. Start Gatus
4. Wait for heartbeat timeout in the maintenance window

```yaml
---
alerting:
  telegram:
    token: "$TELEGRAM_TOKEN"
    id: "$TELEGRAM_CHAT_ID"
    api-url: "https://api.telegram.org"
    default-alert:
      enabled: true
      description: "Health check failed"
      failure-threshold: 1
      success-threshold: 1
      send-on-resolved: true

external-endpoints:
  - name: "test maintenance"
    group: "test"
    token: "testtoken"
    heartbeat:
      interval: "1m"
    maintenance-windows:
      start: "12:00"
      duration: "1h"
      timezone: "Europe/Berlin"
    alerts:
      - type: telegram
```

### Version

v5.34.0

### Additional information

_No response_

anandghegde:
## Summary

Maintenance windows on external endpoints never took effect (#1551).

**Cause:** `maintenance.Config.ValidateAndSetDefaults()` does more than validate. It parses `start` and loads `timezone`, and `IsUnderMaintenance()` relies on both. `ExternalEndpoint.ValidateAndSetDefaults()` never called it. #1369 added the `IsUnderMaintenance()` check for external endpoints, but the windows were still never parsed. So every window acted as if it started at 00:00 in the server's local time, whatever `start`/`timezone` said, and alerts went out during the real window. It also meant an invalid window (e.g. `start: "25:00"`) was accepted without an error.

**Same bug, second place:** `Endpoint.ValidateAndSetDefaults()` returns early for DNS and SSH endpoints (`return e.DNSConfig.ValidateAndSetDefault()` / `return e.SSHConfig.Validate()`). The maintenance window loop comes after those returns, so DNS and SSH endpoints never reach it. I included it here because the cause and fix are the same.

**Fix:**
- `ExternalEndpoint.ValidateAndSetDefaults()` now validates each maintenance window, using the same loop as `Endpoint`.
- In `Endpoint.ValidateAndSetDefaults()`, the maintenance window loop now runs before the DNS/SSH early returns. No other behaviour changes. The only side effect is error order: if an endpoint has both an invalid maintenance window and an invalid DNS/SSH/type config, the maintenance window error is now reported first.

## Testing

- New tests for external, DNS and SSH endpoints:
  - An invalid window (`start: "25:00"`) must return a validation error.
  - A one-hour window in `Europe/Berlin` that started one minute ago must make `IsUnderMaintenance()` return true.
  - All of these tests fail without the fix and pass with it.
- `go test ./...`: passes, except for failures that also happen on unmodified `master` on my machine. `TestPing` and `TestIntegrationEvaluateHealthForICMP` need ICMP. `TestProcessUptimeAfterResult` and `TestStore_InsertCleansUpOldUptimeEntriesProperly` depend on the timezone and pass with `TZ=UTC`.
- `go vet ./...` and `gofmt -l config/endpoint/` report nothing.
- Manual run with an external endpoint using a `custom` alert (failure-threshold 1) that posts to a local listener, and a maintenance window that started a minute earlier in `Europe/Berlin`. I pushed `success=false` to the endpoint:
  - `master`: the alert was sent (`Sending custom alert ... has been TRIGGERED`).
  - This branch: no alert. The log shows `Under endpoint maintenance window` and `Not handling alerting because currently in the maintenance window`.

Fixes #1551

Made by an AI agent (Claude Code, model Claude Opus 5).


wahajahmed010:
Real bug, clean fix. The relocation of `maintenance.ValidateAndSetDefaults` from after the `TypeUNKNOWN` short-circuit to before the `DNSConfig` short-circuit is exactly the missing call site that #1369 implied but did not add.

One thing I'd push for in the test:

The new test asserts the window is *under* maintenance using a window that started one minute ago in `Europe/Berlin`. The assertion only checks `endpoint.MaintenanceWindows[0].IsUnderMaintenance()` — i.e. on the `Config` struct directly. It does *not* exercise the integration path that motivated the fix: `endpoint.IsUnderMaintenance()` on the `Endpoint` struct (which iterates `MaintenanceWindows`). If the integration check in `Endpoint.IsUnderMaintenance()` had a bug — for example, if it copied windows before the loop parsed them — the regression test would still pass. A second `t.Run("endpoint.IsUnderMaintenance reflects parsed window")` that calls `endpoint.IsUnderMaintenance()` and asserts `true` would close that gap and make this a true regression test for the reported behavior.

The fix itself is fine; the test as written proves the validation call runs, but not that the validation actually plumbs through to the endpoint-level decision.

anandghegde:
Thanks @wahajahmed010, good call. There's no `Endpoint.IsUnderMaintenance()` method (the window loop lives inline in `api.CreateExternalEndpointResult`), so I added `TestCreateExternalEndpointResultUnderEndpointMaintenanceWindow` in 670e2042, which pushes a failing result through the API with an active window and asserts alerting is skipped. It fails without the fix.

anandghegde:
Thanks for the review! There's no `IsUnderMaintenance()` on `Endpoint` itself — the only one is `maintenance.Config.IsUnderMaintenance()` (`config/maintenance/maintenance.go:100`).

Everything that decides whether an endpoint is under maintenance iterates `MaintenanceWindows` and calls it per window:

- `watchdog/endpoint.go:61`
- `watchdog/external_endpoint.go:69`
- `api/external_endpoint.go:73`

The test makes that same call on the same `*maintenance.Config` pointers that `ValidateAndSetDefaults` populates, so it exercises the path those callers use. Before the fix the window is unparsed and the assertion fails; after it, it passes.

Happy to add a watchdog-level test if @TwiN would like one here.

PR #1805, opened 2026-09-15
fix(endpoint): support bracketed and unbracketed IPv6 addresses in ICMP endpoints

### LINKED ISSUE #1800: ICMP on IPv6 with Addresses doesn't really work
Hi,

IPv6 adresses, unlike IPv4, can have letters (a to f) in them. Currently (5.36.0), it is not really possible to use ICMP with IPv6:

```icmp://[2a07:6b44:27:11::ac]```
=> gatus starts, but no pings for this endpoint are performed.

```icmp://2a07:6b44:27:11::ac```
=> gatus refuses to start, interpreting the ```ac``` part of the address as a invalid port number

```icmp://[2a07:6b44:27:11::21]```
=> gatus starts and no pings are performed.

```icmp://2a07:6b44:27:11::21```
=> gatus starts and pings are performed.

I'm wondering why address handling is a problem, I haven't looked at the code yet, but I would expect that there's a golang module to use that does evertyhing right.

Regards,
Daniel

Tyagiquamar:
Resolves #1800

### Problem
1. When specifying an unbracketed IPv6 endpoint URL such as `icmp://2a07:6b44:27:11::ac`, `ValidateAndSetDefaults()` ran `http.NewRequest()` unconditionally. `url.Parse` interpreted trailing hex digits as an invalid port number (`invalid port ":ac" after host`), causing Gatus to reject startup.
2. When specifying a bracketed IPv6 endpoint URL such as `icmp://[2a07:6b44:27:11::ac]`, the square brackets remained in `result.Hostname` and were passed to `client.Ping()`. The underlying `pro-bing` library failed to resolve bracketed addresses as raw IP destinations.

### Solution
- Restrict `http.NewRequest()` check in `ValidateAndSetDefaults()` to HTTP-like endpoint types (`TypeHTTP`, `TypeSTARTTLS`, `TypeTLS`, `TypeWS`, `TypeGRPC`).
- Sanitize and strip enclosing square brackets in `EvaluateHealthWithContext()` and `client.Ping()` for ICMP targets so IPv6 destinations resolve and ping reliably regardless of bracket format.
- Add regression unit tests for both bracketed and unbracketed IPv6 ICMP URLs in `endpoint_test.go` and `client_test.go`.


wahajahmed010:
The bracket-stripping logic is correct and idempotent, and the test cases cover both bracketed and unbracketed forms for two distinct IPv6 addresses. Two notes:

1. **`Ping` strips brackets on every call, including IPv4.** The `TrimPrefix("[") + TrimSuffix("]")` is idempotent, so `Ping("127.0.0.1")` still receives `"127.0.0.1"`. Worth a one-line comment near the strip (`// idempotent for non-bracketed addresses`) or a test asserting that behavior, so future readers don't wonder if the strip is conditional. The `TestPing` IPv4 cases in the file already cover this indirectly but they're now under a comment that only mentions the bracketed case.

2. **`Ping("[::1]", &Config{... Network: "ip"})` requires the test host to actually allow ICMP to `::1`.** In sandboxed CI (some GitHub Actions runners, containers without `--cap-add=NET_RAW`) this will fail with a permission error rather than a logic error, and the test will look unreliable. Consider `t.Skip` when `net.PingIP("ip", "::1")` returns `false` for a permissions reason, or use the `icmp4`/`icmp6` packages directly with a fake dialer. Optional, but the IPv6 test is the one most likely to flake on different runners.

Otherwise the layering is good: `ValidateAndSetDefaults` skips `http.NewRequest` for ICMP, hostname parsing strips brackets once, and `call()` reads the already-stripped `result.Hostname`. Nice and consistent.

PR #1808, opened 2026-09-18
Allow alerting.matrix to use client configuration including proxy-url

### LINKED ISSUE #1806: alerting.matrix should allow client configuration including proxy-url

`alerting.mattermost` allows a `proxy-url` to be configured with `alerting.mattermost.client.proxy-url`

```yaml
alerting:
  mattermost:
    webhook-url: http://mymattermost.example.com
    client:
      proxy-url: "http://myproxy:3128"
```

`client.*` parameters are ignored for matrix

I'm not fluent in Go but the

```go
ClientConfig *client.Config `yaml:"client,omitempty"`
```
Is missing in `alerting/provider/matrix/matrix.go` <https://github.com/TwiN/gatus/blob/7f3873d0b455a0d2bf7880eeee1dd0ea7391b2ab/alerting/provider/matrix/matrix.go#L27> ?

In my env the matrix server is exclusively reachable through an internal proxy because outside of our network

9 months

### Additional information

_No response_

gregoryR:
## Summary

try to fix #1806 

a `client.proxy-url` can now be added to `alerting.matrix`

```yaml
alerting:
  matrix:
    server-url: "https://matrix-client.matrix.org"
    access-token: "123456"
    internal-room-id: "!example:matrix.org"
    client:
      proxy-url: http://proxy.example.com:8080
```

## Checklist


wahajahmed010 (commented):
## Review

Thanks for picking this up — parity with `mattermost` is the right call, and the implementation mirrors the existing pattern cleanly:

- `Config.ClientConfig *client.Config` field with `yaml:"client,omitempty"` matches mattermost exactly.
- `Merge()` correctly treats `nil` as "not set", which preserves the `omitempty` behavior and avoids clobbering a parent override with a zero value.
- `Send()` passes `cfg.ClientConfig` to `client.GetHTTPClient`, which falls back to the default client when `nil`, so existing configs keep working.
- README entry follows the same format as the other alerting providers (`googlechat`, `mattermost`, `telegram`, `custom`).

### One suggestion: add a test for the new field

The pattern across alerting providers is that `Merge` for `ClientConfig` is exercised by at least one test scenario. Currently `matrix_test.go` covers `Validate`, `ValidateWithOverride`, and `Send`, but none of them set a `ClientConfig` on either side. A small addition would lock in the new behavior and prevent regressions, e.g.:

```go
func TestAlertProvider_MergeWithClientConfig(t *testing.T) {
    parent := Config{AccessToken: "1", InternalRoomID: "!a:example.com"}
    // First, without override
    if parent.ClientConfig != nil {
        t.Error("expected ClientConfig to be nil before override")
    }
    // Then, override sets a proxy URL
    override := Config{
        AccessToken:    "2",
        InternalRoomID: "!b:example.com",
        ClientConfig:   &client.Config{ProxyURL: "http://proxy.example.com:8080"},
    }
    parent.Merge(&override)
    if parent.ClientConfig == nil || parent.ClientConfig.ProxyURL != "http://proxy.example.com:8080" {
        t.Error("expected ClientConfig to be carried over from override")
    }
    // And merging a config without ClientConfig must not wipe the existing one
    further := Config{AccessToken: "3", InternalRoomID: "!c:example.com"}
    parent.Merge(&further)
    if parent.ClientConfig == nil || parent.ClientConfig.ProxyURL != "http://proxy.example.com:8080" {
        t.Error("expected ClientConfig to be preserved when override does not set it")
    }
}
```

This is the same gap that would catch any future refactor that drops the pointer-vs-value distinction.

### Minor: README table alignment

The new row has slightly different column spacing from its neighbors in the existing `alerting.matrix` block (extra spaces before `Client configuration` and `{}`). Cosmetic only — `npx prettier --write README.md` would normalize it if you care.

Otherwise this is a solid, minimal change. Thanks for fixing #1806.

PR #1810, opened 2026-09-21
fix(ui): respect defaultSortBy and defaultFilterBy from UI config

### LINKED ISSUE #1807: UI settings in yaml config don't work
### Describe the bug

Hi,

According to the documentation, there are some UI settings, however, they don't work for me with current gatus (5.36.0):

```
ui:
  dashboard-heading: foo
  dashboard-subheading: bar
  default-sort-by: group
```

Regards,
Daniel

siddiqueirshad:
Fixes #1807

### Description
When `default-sort-by` (e.g. `group`), `default-filter-by` (e.g. `failing` / `unstable`), or other UI settings were configured under `ui:` in `config.yaml`, the dashboard view (`Home.vue`) initialized `sortBy` to `'name'`, `groupByGroup` to `false`, and filter flags to `false`. As a result, when opening Gatus in a new browser session (with empty `localStorage`), the UI fallback defaulted to hardcoded values (`'name'` sort, no grouping) instead of honoring `window.config.defaultSortBy` and `window.config.defaultFilterBy`.

### Fix
Updated `Home.vue` to initialize `sortBy`, `groupByGroup`, `showOnlyFailing`, and `showRecentFailures` using `window.config.defaultSortBy` and `window.config.defaultFilterBy` when `localStorage` has no stored user overrides.

### Verification
Ran all `config/ui` and `config` Go tests (`go test ./config/...`).


siddiqueirshad:
Hi @nikolauspschuetz, when you get a chance, could you please take a look at this PR? Thanks!

nikolauspschuetz (commented):
Thanks for the ping @siddiqueirshad — nice catch on the underlying issue, the `localStorage`-empty case genuinely did ignore `default-sort-by`/`default-filter-by`.

One blocker before this can land: the diff removes `const searchQuery = ref('')` but `searchQuery` is still used in six places, so the build fails.

```
web/app/src/views/Home.vue
  233:7   error  'searchQuery' is not defined  no-undef
  234:19  error  'searchQuery' is not defined  no-undef
  277:7   error  'searchQuery' is not defined  no-undef
  278:19  error  'searchQuery' is not defined  no-undef
  477:3   error  'searchQuery' is not defined  no-undef

✖ 5 problems (5 errors, 0 warnings)
 ERROR  Error: Build failed with errors.
```

(`make frontend-install && make frontend-build` on node:22, at f0bd36a. It's also referenced from the template at line 46.) Looks like an accidental deletion when the neighbouring `ref`s were rewritten — re-adding that one line should clear it. Worth noting `test-ui` doesn't appear to have reported on this PR, so CI hasn't surfaced it for you.

The rest reads well to me, and I like that the fallback chain is exactly the one already in `SearchBar.vue:50` rather than a new pattern — and `gatus:filter-by` really is written (`SearchBar.vue:70`), so the read has a writer.

One thing worth a second look, non-blocking: `sortBy` and `groupByGroup` are both derived from `initialSortBy`, so `default-sort-by: group` yields `groupByGroup = true` **and** `sortBy = 'group'`. Downstream only ever tests `sortBy.value === 'health'`, so `'group'` falls through to the default ordering — which I think is what you want, but it does mean `sortBy` is carrying a value that isn't a sort mode. Might be clearer as `sortBy = ref(initialSortBy === 'group' ? 'name' : initialSortBy)`, if that matches the intent.

Not a maintainer here, just passing through — @TwiN has the final say.

